PR #42, opened 2015-01-06
Fixes #8850 - import gutterball cert after katello nssdb exists

dustints:


dustints:
abandon because of new gutterball plugin in puppet-katello.
plugin will be added from katello-installer project

ehelms:
ACK

PR #44, opened 2015-01-21
Refs #8756: Deploy the server_ca to the Capsule directories for RHSM.

ehelms:
This fixes an issue where for stand alone Capsules that didn't have
access to the server_ca (since it was not deployed) the bootstrap
RPM would fail to be created. Further, this removes the unused
candlepin-local.pem which in reality was just the root CA cert.

### DIFF
--- a/manifests/katello.pp
+++ b/manifests/katello.pp
@@ -16,6 +16,7 @@ class certs::katello (
   $candlepin_consumer_name        = "katello-ca-consumer-${::fqdn}"
   $candlepin_consumer_summary     = "Subscription-manager consumer certificate for Katello instance ${::fqdn}"
   $candlepin_consumer_description = 'Consumer certificate and post installation script that configures rhsm.'
+
   file { $katello_www_pub_dir:
     ensure => directory,
     owner  => 'apache',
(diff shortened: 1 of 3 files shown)


stbenjam:
ACK

ehelms:
@stbenjam I need to push some updates to this change I encountered last night.

ehelms:
Updated.

stbenjam:
ACK

manifests/katello.pp
```diff
@@ -36,10 +37,9 @@
     dir              => $katello_www_pub_dir,
     summary          => $candlepin_consumer_summary,
     description      => $candlepin_consumer_description,
-    files            => ["${rhsm_ca_dir}/candlepin-local.pem:644=${certs::ssl_build_dir}/${certs::default_ca_name}.crt",
-      "${rhsm_ca_dir}/katello-server-ca.pem:644 =${certs::ssl_build_dir}/${certs::server_ca_name}.crt"],
+    files            => ["${rhsm_ca_dir}/katello-server-ca.pem:644 =${certs::pki_dir}/certs/${certs::server_ca_name}.crt"],
```
iNecas: Is this change compatible with katello-agent https://github.com/Katello/katello-agent/blob/master/etc/gofer/plugins/katelloplugin.conf. Btw. default_ca != server_ca, it's only coincidence they are the same, when using self-signed generated certs, but are different when using custom certs for ssl.

stbenjam: Discussed w/ @iNecas on IRC and it appears this breaks the agent with custom certs.  RHSM would be using the server_ca but qpid-dispatch-router would be using default_ca.

iNecas: The BZ describing the issue is here https://bugzilla.redhat.com/show_bug.cgi?id=1222912

ehelms: Ok... so lets get a Redmine issue and fix it :)

PR #45, opened 2015-01-21
Refs #7745: Deploy client cert bundle specifically for use by the Capsul...

ehelms:
...e.

Note this is a bundle since that is required by the reverse proxy
feature being added to the Capsule.

### DIFF
--- a/manifests/apache.pp
+++ b/manifests/apache.pp
@@ -57,7 +57,7 @@ class certs::apache (
     } ->
     file { $apache_key:
       owner => $::apache::user,
-      group => $::foreman::group,
+      group => $::certs::group,
       mode  => '0440',
     } ->
     Service['httpd']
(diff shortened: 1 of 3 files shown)


manifests/capsule.pp
```diff
@@ -10,31 +10,67 @@
 # $certs_tar::               path to tar file with certs to generate. REQUIRED
 #
```
dustints: comment for `$generate` argument needed

manifests/capsule.pp
```diff
@@ -10,31 +10,67 @@
...
+  }
 
-  # until we support again pushing the cert rpms to the Katello,
-  # make sure the certs_tar path is present
-  validate_present($certs_tar)
-  validate_present($capsule_fqdn)
+  if $generate {
+    # until we support again pushing the cert rpms to the Katello,
+    # make sure the certs_tar path is present
+    validate_present($certs_tar)
```
dustints: when installing katello bin/katello-installer exec in katello installer, validation fails.

ehelms:
@dustints @stbenjam updated - this is still required for the reverse proxy (which I will be updating shortly) capsule PR but this can in theory be a stand alone change.

stbenjam:
Tested & works, ACK from me!

PR #48, opened 2015-02-03
Fixes #9204: Resolve conflict with similar cert names.

ehelms:
If two certs have roughly the same name, one can over ride the over
during lookup for existence. For example, katello-foreman-proxy and
katello-foreman-proxy-client would have been resolved to the latter name
and then when the certs are deployed to the certs directory from
the RPMs they would not exist.


jlsherrill:
ACK

PR #49, opened 2015-02-11
Fixes #8636 - Katello CA cert now trusted system wide

parthaa:
This commit adds the katello ca trust pem to the system wide trust store
so that services like docker and subscription manager can now use it.

### DIFF
--- a/templates/rhsm-katello-reconfigure.erb
+++ b/templates/rhsm-katello-reconfigure.erb
@@ -23,7 +23,7 @@ BASEURL=https://$KATELLO_SERVER/pulp/repos
 PREFIX=<%= @deployment_url %>
 CFG=/etc/rhsm/rhsm.conf
 CFG_BACKUP=$CFG.kat-backup
-
+CA_TRUST_ANCHORS=/etc/pki/ca-trust/source/anchors
 # Get version of RHSM
 RHSM_V="`rpm -q --queryformat='%{VERSION}' subscription-manager 2> /dev/null | tr . ' '`"
 if test $? != 0 ; then
@@ -61,6 +61,17 @@ if grep --quiet full_refresh_on_yum $CFG; then
 else
   full_refresh_config="#config for on-premise management\nfull_refresh_on_yum = 1"
   sed -i "s/baseurl.*/&\n\n$full_refresh_config/g" $CFG
+fi
+
+# also add the katello ca cert to the system wide ca cert store
+if [-d $CA_TRUST_ANCHORS]; then
+  update-ca-trust enable
+  cp %%(ca_cert_dir)skatello-server-ca.pem $CA_TRUST_ANCHORS
+  update-ca-trust
+
+  #restart if docker service is installed
+  service docker status >/dev/null && \
+    service docker restart >/dev/null 2&>1
 fi
 
 # restart goferd if it is installed and running


templates/rhsm-katello-reconfigure.erb
```diff
@@ -63,6 +63,16 @@ else
   sed -i "s/baseurl.*/&\n\n$full_refresh_config/g" $CFG
 fi
 
+# also add the katello ca cert to the system wide ca cert store
+if [-d $CA_TRUST_ANCHORS]; then
+  cp %%(ca_cert_dir)skatello-server-ca.pem $CA_TRUST_ANCHORS
+  update-ca-trust
```
stbenjam: On some distros you need to do `update-ca-trust enable` first.  I think that may include EL6.  Not sure, just something I always run

templates/rhsm-katello-reconfigure.erb
```diff
@@ -63,6 +63,16 @@ else
...
+
+  #restart if docker service is installed
+  service docker status >/dev/null && \
+    service docker restart >/dev/null 2&>1
+fi
+
 # restart goferd if it is installed and running
```
mccun934: do you need to account for systemd on el7+ here?

parthaa: I tried service docker restart on my el7 instance. Seems to work ok.

parthaa: # service docker restart
Redirecting to /bin/systemctl restart  docker.service

stbenjam:
ACK

templates/rhsm-katello-reconfigure.erb
```diff
@@ -63,6 +63,17 @@ else
   sed -i "s/baseurl.*/&\n\n$full_refresh_config/g" $CFG
 fi
 
+# also add the katello ca cert to the system wide ca cert store
+if [-d $CA_TRUST_ANCHORS]; then
+  update-ca-trust enable
+  cp %%(ca_cert_dir)skatello-server-ca.pem $CA_TRUST_ANCHORS
```
ehelms: The name of the cert needs to be set based on the parameter lest it not work when someone configures it -- https://github.com/Katello/puppet-certs/blob/master/manifests/init.pp#L111

ehelms:
Please see my comment before merging.

parthaa:
I got the name from https://github.com/parthaa/puppet-certs/blob/cert-add/templates/rhsm-katello-reconfigure.erb#L49 .. want me to change there also ....

stbenjam:
Also L55.

It's hard coded a lot of places, I seriously doubt you could change the name today without problems, but might as well fix it while in the file...

templates/rhsm-katello-reconfigure.erb
elyezer: I think service redirects to systemctl as a deprecated action, need to check if it will be removed in the future.

elyezer:
Any update on this? We have been doing work around in order to test pulling from katello docker repository, but would be good to test this sooner than later before beta.

jlsherrill:
updated here:  https://github.com/Katello/puppet-certs/pull/53 based on a request from @parthaa

parthaa:
thx!

PR #50, opened 2015-02-16
Fixes #9392 - Substitute foreman-config (deprecated) for foreman-rake config

eLobato:


stbenjam:
Thanks @eLobato, you didn't need to close the `katello-installer` PR, you'll still need one (needs `librarian-update katello-certs` to get this update in the installer).

PR #51, opened 2015-02-19
refs #8175 - certificates for dispatch router

stbenjam:
Using the existing broker ones just isn't feasible, it needs its own unfortunately...

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -93,6 +93,13 @@ class certs::params {
 
   $messaging_client_cert = "${pki_dir}/qpid_client_striped.crt"
 
+  $qpid_router_server_cert = "${pki_dir}/qpid_router_server.crt"
+  $qpid_router_client_cert = "${pki_dir}/qpid_router_client.crt"
+  $qpid_router_server_key  = "${pki_dir}/qpid_router_server.key"
+  $qpid_router_client_key  = "${pki_dir}/qpid_router_client.key"
+  $qpid_router_owner       = 'qpidd'
+  $qpid_router_group       = 'root'
+
   $pulp_server_ca_cert   = '/etc/pki/pulp/server_ca.crt'
   # Pulp expects the node certificate to be located on this very location
   $nodes_cert_dir        = '/etc/pki/pulp/nodes'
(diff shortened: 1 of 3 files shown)


manifests/params.pp
```diff
@@ -93,6 +93,13 @@
 
   $messaging_client_cert = "${pki_dir}/qpid_client_striped.crt"
 
+  $qpid_dispatch_server_cert = "/etc/qpid-dispatch/server.crt"
+  $qpid_dispatch_client_cert = "/etc/qpid-dispatch/client.crt"
+  $qpid_dispatch_server_key  = "/etc/qpid-dispatch/server.key"
+  $qpid_dispatch_client_key  = "/etc/qpid-dispatch/client.key"
```
ehelms: Are these the dispatch router defaults or expected locations? i.e. are we able to configure the dispatch router where to look for certs ?

manifests/qpid_router.pp
```diff
@@ -0,0 +1,89 @@
...
+  $deploy                 = $::certs::deploy,
+  $server_cert            = $::certs::qpid_dispatch_server_cert,
+  $client_cert            = $::certs::qpid_dispatch_client_cert,
+  $server_key             = $::certs::qpid_dispatch_server_key,
+  $client_key             = $::certs::qpid_dispatch_client_key,
+  $owner                  = $::certs::qpid_dispatch_owner,
+  $group                  = $::certs::qpid_dispatch_group,
+) inherits certs::params {
+
+  $server_keypair = "${hostname}-qdrouterd-server"
```
ehelms: What is qdrouterd?

manifests/params.pp
stbenjam: No, where would you put them?

manifests/qpid_router.pp
stbenjam: **Q**pid **D**ispatch **Router** **D**aemon, but I renamed it to qpid_router to be consistent with the class name.

stbenjam:
@ehelms Updated

ehelms:
ACK

PR #53, opened 2015-03-03
fixes #8636 - Katello CA cert now trusted system wide

jlsherrill:
This commit adds the katello ca trust pem to the system wide trust store
so that services like docker can now use it

### DIFF
--- a/manifests/katello.pp
+++ b/manifests/katello.pp
@@ -3,6 +3,7 @@ class certs::katello (
   $hostname                      = $fqdn,
   $deployment_url                = undef,
   $rhsm_port                     = 443,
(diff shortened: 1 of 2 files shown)


ehelms:
From a 'did this properly deploy the cert to the right place standpoint' - ACK for me. We will need to file an issue and fix the case for previous setups to allow regenerating the bootstrap RPM via the installer for upgrade process. Users can just delete the bootstrap RPM to force regeneration, but since this would need to occur for all cases we just automate it.

parthaa:
Ack manually patched and tested on a beaker machine with Sat6. Works great with docker stuff.

jlsherrill:
filed:  http://projects.theforeman.org/issues/9665

PR #54, opened 2015-03-10
Fixes #9699: Check for nssdb creation before running certutil.

ehelms:
Switches to checking for the nssb execute that creates the actual
database instead of just the directory that contains them. This
prevents the following error on EL6:

  certutil: function failed: SEC_ERROR_LEGACY_DATABASE: The
  certificate/key database is in an unsupported format.


jlsherrill:
ACK

PR #56, opened 2015-03-23
Fixes #9875: Better docker service restart

elyezer:
Improve the check for the docker service in order to restart it if
needed. Also check for systemd based system first and then fall back to
the service based utilities.

### DIFF
--- a/templates/rhsm-katello-reconfigure.erb
+++ b/templates/rhsm-katello-reconfigure.erb
@@ -71,9 +71,14 @@ if [ -d $CA_TRUST_ANCHORS ]; then
   cp $KATELLO_CERT_DIR/$KATELLO_CERT $CA_TRUST_ANCHORS
   update-ca-trust
 
-  #restart if docker service is installed
-  service docker status >/dev/null && \
+  # restart docker if it is installed and running
+  if [ -f /usr/lib/systemd/system/docker.service ]; then
+    systemctl status docker >/dev/null && \
+    systemctl restart docker >/dev/null 2&>1
+  elif [ -f /etc/init.d/docker ]; then
+    service docker status >/dev/null && \
     service docker restart >/dev/null 2&>1
+  fi
 fi
 
 # restart goferd if it is installed and running


mccun934:
Can you file an issue here: 

http://projects.theforeman.org/projects/foreman/issues/new

and update the commit message to indicate that it fixes that issue, for example:

https://github.com/Katello/puppet-certs/commit/ffd3db1b0097f91de76a5b63384381b753d02471

elyezer:
@mccun934 update, thanks for pointing the guidelines.

templates/rhsm-katello-reconfigure.erb
```diff
@@ -71,9 +71,14 @@ if [ -d $CA_TRUST_ANCHORS ]; then
   cp $KATELLO_CERT_DIR/$KATELLO_CERT $CA_TRUST_ANCHORS
   update-ca-trust
 
-  #restart if docker service is installed
-  service docker status >/dev/null && \
+  # restart docker if it is installed and running
+  if [ -f /usr/lib/systemd/system/docker.service ]; then
```
Ichimonji10: Unit files installed by packages are placed in `/usr/lib/systemd/system/`, but system administrators may also install unit files in `/etc/systemd/system/`. An administrator might even do something crazy like nuke the default service file and install a custom one to `/etc/systemd/system/`. As a result, this sort of test may be better:

    if which systemctl >&/dev/null; then echo true; else echo false; fi

elyezer: @Ichimonji10 even though I use the check you proposed I will have to check for those unit files to know if docker is installed or not. If the sysadmin changes places so the message showing that the services was restarted will not be printed.

Also other stuff I left according to what I found in other places in this codebase. But I can definitely update as per your suggestions, let's see what others tell about them.

Ichimonji10: @elyezer, I totally forgot about checking for docker.service in particular, so my example code is incomplete. Whoops! Still, I think there's a case to be made for using utilities to check for the presence of unit files or service files instead of directly querying the filesystem for the presence of files. I'd love to hear what others have to say.

Ichimonji10: This example deals with the case where the `systemctl` executable and `docker.service` unit file may or may not exist in standard locations on the system:

```sh
if systemctl list-unit-files docker.service | grep '^docker.service' >&/dev/null; then
    echo 'systemctl docker.service may be used'
elif [ -f /etc/init.d/docker ]; then
    echo 'service docker may be used'
fi
```

I am not terribly knowledgeable about `service` and where service files may exist on a system.

elyezer: @Ichimonji10 I tried your suggestion and get the following. Also I have to remove the docker.service option from `list-unit-files`. I will need to deal with the output of systemd not being available and that will make the code bigger. I will stick with 'Simple is better than complex.' here.

```
[root@rhel66 ~]# if systemctl list-unit-files | grep '^docker.service' >&/dev/null; then echo 'have docker'; else echo 'does not have docker'; fi
-bash: systemctl: command not found
does not have docker
```

Ichimonji10: @elyezer OK. That sounds good.

ehelms:
@parthaa can you take a look at this?

ehelms:
Thanks @elyezer @Ichimonji10 !

Ichimonji10:
:heart:

elyezer:
Thank you!

PR #57, opened 2015-03-24
Fixes #9888 - use random_password over generate_password

dustints:
generate_password has a dependency on openssl.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -120,7 +120,7 @@ class certs (
   $ca_key = "${certs::pki_dir}/private/${default_ca_name}.key"
   $ca_cert = "${certs::pki_dir}/certs/${default_ca_name}.crt"
   $ca_cert_stripped = "${certs::pki_dir}/certs/${default_ca_name}-stripped.crt"
-  $ca_key_password = cache_data('ca_key_password', generate_password())
+  $ca_key_password = cache_data('ca_key_password', random_password(24))
   $ca_key_password_file = "${certs::pki_dir}/private/${default_ca_name}.pwd"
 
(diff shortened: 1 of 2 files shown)


manifests/init.pp
```diff
@@ -120,7 +120,7 @@
   $ca_key = "${certs::pki_dir}/private/${default_ca_name}.key"
   $ca_cert = "${certs::pki_dir}/certs/${default_ca_name}.crt"
   $ca_cert_stripped = "${certs::pki_dir}/certs/${default_ca_name}-stripped.crt"
-  $ca_key_password = cache_data('ca_key_password', generate_password())
+  $ca_key_password = cache_data('ca_key_password', random_password())
```
stbenjam: random_password takes a length argument, and you should bump the foreman requirement in metadata.json to >= 2.0.0, as that is when it was introduced

dustints: updated

manifests/init.pp
```diff
@@ -120,7 +120,7 @@
   $ca_key = "${certs::pki_dir}/private/${default_ca_name}.key"
   $ca_cert = "${certs::pki_dir}/certs/${default_ca_name}.crt"
   $ca_cert_stripped = "${certs::pki_dir}/certs/${default_ca_name}-stripped.crt"
-  $ca_key_password = cache_data('ca_key_password', generate_password())
+  $ca_key_password = cache_data('ca_key_password', random_password(8))
```
stbenjam: Nitpicky, but 8 is very short.  I'd say 24 at least.  No one's going to be typing this in

dustints:
@stbenjam sounds good, updated!

stbenjam:
ACK

PR #58, opened 2015-04-15
Fixes #11755: Validate absolute path for custom certificates

ehelms:


ehelms:
@iNecas would you mind reviewing this change?

manifests/init.pp
```diff
@@ -112,7 +112,7 @@
   ) inherits certs::params {
 
   if $server_cert {
-    validate_file_exists($server_cert, $server_cert_req, $server_key, $server_ca_cert)
+    validate_absolute_path([$server_cert, $server_cert_req, $server_key, $server_ca_cert])
```
iNecas: This is not the same: we need to check that the file really exists. On the other hand, even more checks happen now in https://github.com/Katello/katello-installer/blob/master/hooks/pre/20-certs_update.rb#L62, so I don't have that much issues then

ehelms:
@iNecas I re-purposed this PR to instead add absolute path validation for custom certificates since this is a common pitfall.

iNecas:
Can there be a case, where the `validate_absolute_path` would fail and `validate_file_exists` would pass and the input would be not suitable for the installation?

ehelms:
I don't follow the question. If the absolute path check fails, then it
would fail all together?

iNecas:
I mean if it would not pass the `validate_absolute_path` criteria, but the check woudln't be there. Basically my question is, what are the now-invalid data that cause the installer to fail, when this patch it not applied?

ehelms:
One situation is where you have specified these parameters without an absolute path, and you re-run it from a different location. This is probably less a problem with normal puppet workflows, but when used with something like kafo, if you re-run an installer it will complain about not finding these certs since they are not absolute.

iNecas:
Ok, if this is the case, I agree this will fix it, given it will properly report the error message so that the user knows to use absolute path instead

ehelms:
@iNecas is that an ACK or are you asking for something else in addition?

iNecas:
Sry for not being explicit, ACK :)

ehelms:
Thanks @iNecas

PR #59, opened 2015-05-01
fixes #10350 - switch to qdrouterd user for certs + keys

mccun934:
The latest version of the Dispatch Router runs under the qdrouterd
user instead of root as it previously did. This means the certs+keys
all need to be owned by the qdrouterd user instead of qpidd.


mccun934:
@stbenjam if you don't mind a review, would appreciate it.

stbenjam:
Looks good to me, do we have the new packages to test? Do we need to change the config ownership?
  
  https://github.com/Katello/puppet-qpid/blob/master/manifests/router/config.pp#L28

stbenjam:
But ACK for this one

PR #61, opened 2015-05-27
Fixes #10097 - Fixed references to city

adamruzicka:
In many places it was referenced as 'sity'


stbenjam:
Looks like we need to update katello-devel module as well:
 `puppet-katello_devel/manifests/foreman_certs.pp: :    city          => $::certs::sity`

adamruzicka:
I'll check all the katello's repositories for more appearances of it tomorrow

stbenjam:
That's the only reference I saw in katello-installer, but anyway, I'll ack whatever PR's come out of this together so we don't do break anything.

stbenjam:
ACK

ehelms:
@stbenjam please go ahead and merge this

PR #62, opened 2015-06-01
Fixes #10670 - deploy the katello-default-ca as part of the bootstrap RPM

iNecas:


manifests/katello.pp
```diff
@@ -39,7 +39,10 @@
     dir              => $katello_www_pub_dir,
     summary          => $candlepin_consumer_summary,
     description      => $candlepin_consumer_description,
-    files            => ["${rhsm_ca_dir}/katello-server-ca.pem:644 =${certs::pki_dir}/certs/${certs::server_ca_name}.crt"],
+    # katello-default-ca is needed for the katello-agent to work properly
+    # (especially in the custom certs scenario)
+    files            => ["${rhsm_ca_dir}/katello-default-ca.pem:644=${certs::ssl_build_dir}/${certs::default_ca_name}.crt",
```
ehelms: Will the default_ca be present in the ssl_build_dir on a capsule? If we re installing the RPMs, they would deploy it to the pki_dir I believe similar to the server_ca_name.

iNecas: Good catch, updated

ehelms:
APT

PR #65, opened 2015-09-08
refs #11737 - support cnames and add localhost cname to qpid certs

stbenjam:
This will require the user to do a `--certs-update-all` for the next upgrade, so qpid gets the new cert with the additional CN.

### DIFF
--- a/lib/puppet/provider/cert/katello_ssl_tool.rb
+++ b/lib/puppet/provider/cert/katello_ssl_tool.rb
@@ -10,6 +10,7 @@ Puppet::Type.type(:cert).provide(:katello_ssl_tool, :parent => Puppet::Provider:
               '--server-cert-req', File.basename(req_file),
               '--server-key', File.basename(privkey),
               '--server-rpm', rpmfile_base_name ]
+
     if resource[:custom_pubkey]
       FileUtils.mkdir_p(build_path)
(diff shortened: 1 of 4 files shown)


lib/puppet/provider/cert/katello_ssl_tool.rb
```diff
@@ -25,6 +26,15 @@ def generate!
                    '--ca-key', ca_details[:privkey]])
       args.concat(common_args)
     end
+
+    if resource[:cname]
+      if resource[:cname].is_a?(String)
+        args << ['--set-cname', resource[:cname]]
+      else
+        resource[:cname].map { |cname| ['--set-cname', cname] }.flatten
```
ehelms: Does this need to be added to the args array?

stbenjam: That would make things actually work, yes :-)

beav:
@stbenjam is there a way to force the `--certs-update-all` to happen on upgrade, so users don't have to remember to type it?

stbenjam:
Maybe in the installer pre hook on the 3.2->3.3 upgrade?

https://github.com/Katello/katello-installer/blob/master/hooks/pre/20-certs_update.rb

ehelms:
Will this also require new certs to be generated for each capsule?

stbenjam:
Yea, unfortunately, everywhere qpid runs

PR #66, opened 2015-09-09
Pulp consumer_ca_cert is now ca_cert

ehelms:
The puppet-pulp 1.0 module removed the consumer_ca_cert parameter
in favor of ca_cert.


stbenjam:
ACK

PR #67, opened 2015-09-11
Fixes #10052 - Code to setup rhsm.conf for atomic hosts

parthaa:
Atomic hosts don't deal with Rpms. This script provides an alternative way to setup the rhsm.conf by including the cert + rpm post script as a separate entity. It also generates a katello-rhsm-atomic-consumer.tar.gz , which has this. So the user is expected download the tar gz , extract, cd into atomic and then run "katello-rhsm-atomic-consumer" script for it to setup rhsm.conf .

### DIFF
--- a/manifests/katello.pp
+++ b/manifests/katello.pp
@@ -3,8 +3,7 @@ class certs::katello (
   $hostname                      = $fqdn,
   $deployment_url                = undef,
   $rhsm_port                     = 443,
-  $server_ca_name                = $::certs::server_ca_name,
-  $candlepin_cert_rpm_alias_filename = undef
+  $candlepin_cert_rpm_alias_filename = undef,
   ){
 
   $candlepin_cert_rpm_alias = $candlepin_cert_rpm_alias_filename ? {
@@ -14,6 +13,9 @@ class certs::katello (
 
   $katello_www_pub_dir            = '/var/www/html/pub'
   $rhsm_ca_dir                    = '/etc/rhsm/ca'
+  $katello_rhsm_setup_script      = 'katello-rhsm-consumer'
+  $katello_rhsm_setup_script_location = "/usr/bin/${katello_rhsm_setup_script}"
+
   $candlepin_consumer_name        = "katello-ca-consumer-${::fqdn}"
   $candlepin_consumer_summary     = "Subscription-manager consumer certificate for Katello instance ${::fqdn}"
   $candlepin_consumer_description = 'Consumer certificate and post installation script that configures rhsm.'
@@ -33,23 +35,28 @@ class certs::katello (
   # Placing the CA in the pub dir for trusting by a user in their browser
   file { "${katello_www_pub_dir}/${certs::server_ca_name}.crt":
     ensure  => file,
-    source  => "${certs::pki_dir}/certs/${certs::server_ca_name}.crt",
+    source  => $certs::katello_server_ca_cert,
     owner   => 'root',
     group   => 'root',
     mode    => '0644',
-    require => File["${certs::pki_dir}/certs/${certs::server_ca_name}.crt"],
+    require => File[$certs::katello_server_ca_cert],
   } ~>
-  # We need to deliver the server_ca for yum and rhsm to trust the server
-  # and the default_ca for goferd to trust the qpid
+  # Generate the the rhsm setup script in the pub dir for rhsm setup
+  file { "${katello_www_pub_dir}/${katello_rhsm_setup_script}":
+    ensure  => file,
+    owner   => 'root',
+    group   => 'root',
+    mode    => '0755',
+    content => template('certs/rhsm-katello-reconfigure.erb'),
+  } ~>
   certs_bootstrap_rpm { $candlepin_consumer_name:
     dir              => $katello_www_pub_dir,
     summary          => $candlepin_consumer_summary,
     description      => $candlepin_consumer_description,
     # katello-default-ca is needed for the katello-agent to work properly
     # (especially in the custom certs scenario)
-    files            => ["${rhsm_ca_dir}/katello-default-ca.pem:644=${certs::pki_dir}/certs/${certs::default_ca_name}.crt",
-                        "${rhsm_ca_dir}/katello-server-ca.pem:644=${certs::pki_dir}/certs/${certs::server_ca_name}.crt"],
-    bootstrap_script => template('certs/rhsm-katello-reconfigure.erb'),
+    files            => ["${katello_rhsm_setup_script_location}:755=${katello_www_pub_dir}/${katello_rhsm_setup_script}"],
+    bootstrap_script => inline_template('/bin/bash <%= @katello_rhsm_setup_script_location %>'),
     alias            => $candlepin_cert_rpm_alias,
     subscribe        => $::certs::server_ca,
   }
--- a/templates/rhsm-katello-reconfigure.erb
+++ b/templates/rhsm-katello-reconfigure.erb
@@ -1,6 +1,6 @@
 #!/bin/bash
 #
-# Copyright 2013 Red Hat, Inc.
+# Copyright 2016 Red Hat, Inc.
 #
 # This software is licensed to you under the GNU General Public License,
 # version 2 (GPLv2). There is NO WARRANTY for this software, express or
@@ -16,22 +16,26 @@
 # Configures rhsm on client. Called from the certificate RPM.
 #
 
+#copy the certificate to this variable KATELLO_SERVER_CA_DATA
+read -r -d '' KATELLO_SERVER_CA_DATA << EOM
+<%= File.exist?(scope['certs::katello_server_ca_cert']) ? File.open(scope['certs::katello_server_ca_cert']).read : "" %>
+EOM
+
+#copy the default ca certificate to this variable KATELLO_DEFAULT_CA_DATA
+read -r -d '' KATELLO_DEFAULT_CA_DATA << EOM
+<%= File.exist?(scope['certs::katello_default_ca_cert']) ? File.open(scope['certs::katello_default_ca_cert']).read : "" %>
+EOM
+
 KATELLO_SERVER=<%= @hostname %>
-KATELLO_CERT=<%= @server_ca_name %>.pem
+KATELLO_SERVER_CA_CERT=<%= @server_ca_name %>.pem
+KATELLO_DEFAULT_CA_CERT=<%= @default_ca_name %>.pem
 KATELLO_CERT_DIR=<%= @rhsm_ca_dir %>
 PORT=<%= @rhsm_port %>
-BASEURL=https://$KATELLO_SERVER/pulp/repos
 
 PREFIX=<%= @deployment_url %>
 CFG=/etc/rhsm/rhsm.conf
 CFG_BACKUP=$CFG.kat-backup
 CA_TRUST_ANCHORS=/etc/pki/ca-trust/source/anchors
-# Get version of RHSM
-RHSM_V="`rpm -q --queryformat='%{VERSION}' subscription-manager 2> /dev/null | tr . ' '`"
-if test $? != 0 ; then
-  RHSM_V="0 0 0"
-fi
-declare -a RHSM_VERSION=($RHSM_V)
 
 # exit on non-RHEL systems or when rhsm.conf is not found
 test -f $CFG || exit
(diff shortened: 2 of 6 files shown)


ehelms:
What is different about the atomic reconfigure from the existing reconfigure script? 

Thoughts on deploying a script that machines download and run that fetches the cert and reconfigure script and then runs them? I imagine this process will be scripted anyway so we could just handle that directly.

parthaa:
> What is different about the atomic reconfigure from the existing reconfigure script?
> 
Some differences.
1) We know that the newer version of subscription manager is available on atomic so we dont need to do a "rpm -q .." and figure out version of subs manager. Moreover rpm command is not available on atomic so the script will fail right away if we try to do "rpm -q".
2)  The source directory for the  katello ca cert is different in the case of rpm. In the case of the atomic the ca cert is going to be available is in a different place.

But I understand your concern that it will add 2 places to maintain this logic.. Any suggestions on combining? Concat would be hard, would be better if I could do includes or variable magic.

> Thoughts on deploying a script that machines download and run that fetches the cert and reconfigure script and then runs them? I imagine this process will be scripted anyway so we could just handle that directly.

So I noticed that I would have to add a "--insecure" every time I did a wget. I felt it would be better user experience to say, download this tar gz and have the script use the files in that extracted directory. In other word the tar gz would be the equivalent of the rpm and the script inside it would be the equivalent of the post script.

ehelms:
On Mon, Sep 14, 2015 at 12:03 PM, Partha Aji <notifications@github.com>
wrote:

> What is different about the atomic reconfigure from the existing
> reconfigure script?
>
> Some differences.
> 1) We know that the newer version of subscription manager is available on
> atomic so we dont need to do a "rpm -q .." and figure out version of subs
> manager. Moreover rpm command is not available on atomic so the script will
> fail right away if we try to do "rpm -q".
> 2) The source directory for the katello ca cert is different in the case
> of rpm. In the case of the atomic the ca cert is going to be available is
> in a different place.
>
> But I understand your concern that it will add 2 places to maintain this
> logic.. Any suggestions on combining? Concat would be hard, would be better
> if I could do includes or variable magic.
>
I think checks within the script would be better than maintaining two whole
versions of it. Could you detect the presence of 'rpm' with which?

> Thoughts on deploying a script that machines download and run that fetches
> the cert and reconfigure script and then runs them? I imagine this process
> will be scripted anyway so we could just handle that directly.
>
> So I noticed that I would have to add a "--insecure" every time I did a
> wget. I felt it would be better user experience to say, download this tar
> gz and have the script use the files in that extracted directory. In other
> word the tar gz would be the equivalent of the rpm and the script inside it
> would be the equivalent of the post script.
>
Do you not have to add '--insecure' when retrieving the tar.gz? Difference
with the RPM is that the script gets run automatically -- install RPM with
one command and you are done. This requires a few steps that I imagine
would get scripted anyway so I question whether we ought to just automate
it down to one script to begin with?

> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Katello/puppet-certs/pull/67#issuecomment-140125673>.
>



-- 
Eric D. Helms
Red Hat Engineering
Ph.D. Student - North Carolina State University

parthaa:
> Do you not have to add '--insecure' when retrieving the tar.gz? Difference
> with the RPM is that the script gets run automatically -- install RPM with
> one command and you are done. This requires a few steps that I imagine
> would get scripted anyway so I question whether we ought to just automate
> it down to one script to begin with?

I am not sure I get this. Here is the proposed workflow
```
$ wget --no-check-certificate http://<sat fqdn>/pub/katello-rhsm-atomic-consumer.tar.gz  
$ tar zxf katello-rhsm-atomic-consumer.tar.gz 
$ cd atomic
$ ./katello-rhsm-atomic-consumer
```
That tar gz has all the certs and other potential future rpm artifacts that may be needed so that I don't need  the script do -> wget  --no-check-certificate http://<hard coded sat fqdn>/pub/katello-ca.crt

ehelms:
On Mon, Sep 14, 2015 at 2:03 PM, Partha Aji <notifications@github.com>
wrote:

> Do you not have to add '--insecure' when retrieving the tar.gz? Difference
> with the RPM is that the script gets run automatically -- install RPM with
> one command and you are done. This requires a few steps that I imagine
> would get scripted anyway so I question whether we ought to just automate
> it down to one script to begin with?
>
> I am not sure I get this. Here is the proposed workflow
>
> $ wget --no-check-certificate http://<sat fqdn>/pub/katello-rhsm-atomic-consumer.tar.gz
> $ tar zxf katello-rhsm-atomic-consumer.tar.gz
> $ cd atomic
> $ ./katello-rhsm-atomic-consumer
>
> That tar gz has all the certs and other potential future rpm artifacts
> that may be needed so that I don't need the script do -> wget
> --no-check-certificate http:///pub/katello-ca.crt
>

What am I asking is, why wouldn't this workflow work:

$ wget --no-check-certificate http://<sat
fqdn>/pub/katello-ca-consumer-latest.sh
$ ./katello-ca-consumer-latest.sh


You had mentioned RPM isn't available on an atomic host, are we guaranteed
that tar and gunzip are as well?

—
> Reply to this email directly or view it on GitHub
> <https://github.com/Katello/puppet-certs/pull/67#issuecomment-140161662>.
>



-- 
Eric D. Helms
Red Hat Engineering
Ph.D. Student - North Carolina State University

parthaa:
> What am I asking is, why wouldn't this workflow work:

It may work. I just feel its more convinient and consistent with the way we do rpm + script to have it packaged as tar gz

jlsherrill:
I think we were hesitant to write a script that downloads the ca file insecurely.  Leaving it up to the user to transfer the tar file means both the script and the ca make it to the box.  Also there is really no need to gzip the file if that is a concern.  Just tar'ing it is sufficient.

mccun934:
I'm in favor of anything that reduces the steps for the user to a 1-2 liner statement if possible.

You could even reduce it to:

curl  http://<katello fqdn>/pub/katello-ca-consumer-latest.sh | sudo bash

keep in mind, wget isn't always installed by default in RHEL/CentOS, curl/tar/gzip are so rely on curl when you can

ehelms:
Tar or not it's still generally insecure unless they go outside of http to
transfer?

parthaa:
> What is different about the atomic reconfigure from the existing reconfigure script?
> 
> Thoughts on deploying a script that machines download and run that fetches the cert and reconfigure script and then runs them? I imagine this process will be scripted anyway so we could just handle that directly.

I tried very hard on this and almost had it working except for this stupid change in the subs manager command
Atomic Version
```
subscription-manager config \
  --server.hostname="$KATELLO_SERVER" \
  --server.prefix="$PREFIX" \
  --server.port="$PORT" \
  --rhsm.repo_ca_cert="%(ca_cert_dir)s$KATELLO_CERT" \
  --rhsm.baseurl="$BASEURL"

```
vs

RPM Version
```
subscription-manager config \
  --server.hostname="$KATELLO_SERVER" \
  --server.prefix="$PREFIX" \
  --server.port="$PORT" \
  --rhsm.repo_ca_cert="%%(ca_cert_dir)s$KATELLO_CERT" \
  --rhsm.baseurl="$BASEURL"
```

Notice the extra % in --rhsm.repo_ca_cert (extra % necessary because its an rpm post script)

ehelms:
Ahh, bummer, maybe two is the proper way due to these subtleties :/  Could you re-work the declarations in katello.pp so that the atomic chain of types is not intermixed throughout the RPM side of things?

parthaa:
> Could you re-work the declarations in katello.pp so that the atomic chain of types is not intermixed throughout the RPM side of things?

Could you please expand on your comments here, I am trying to figure out what exactly you want me to change. I thought katello.pp was the place put everything that goes on in pub.

parthaa:
> Could you re-work the declarations in katello.pp so that the atomic chain of types is not intermixed throughout the RPM side of things?

@ehelms addressed

manifests/katello.pp
```diff
@@ -46,5 +54,35 @@
     bootstrap_script => template('certs/rhsm-katello-reconfigure.erb'),
     alias            => $candlepin_cert_rpm_alias,
     subscribe        => $::certs::server_ca,
+  } ~>
```
ehelms: This still ties the atomic workflow to the certs_bootstrap_rpm, I think it would be better to break this dependency and let the atomic stuff stand alone i.e. change this line to `}` and insert a blank line.

parthaa:
@ehelms  @stbenjam @jlsherrill @bbuckingham  updating this as per our discussion. 

Main intent of this commit is to setup rhsm conf on an atomic host to
setup consumer registration (a role similar to the one played by the
consumer bootstrap rpm).

Artifacts produced

1) /var/www/html/pub/katello-rhsm-consumer on the satellite. This script
contains the code needed to setup rhsm conf on both atomic as well as
non atomic machines. The contents of this script came out of the template
used for post install script in the bootstrap consumer rpm. It has been
tuned to work for both atomic/non-atomic machines

2) Update to the bootstrap rpm to copy the new script to /usr/bin and
run that as a part of post install.

manifests/katello.pp
```diff
@@ -14,10 +15,16 @@
...
   $katello_www_pub_dir            = '/var/www/html/pub'
   $rhsm_ca_dir                    = '/etc/rhsm/ca'
+  $katello_rhsm_setup_script      = 'katello-rhsm-consumer'
+  $katello_rhsm_setup_script_location = "/usr/bin/${katello_rhsm_setup_script}"
+
   $candlepin_consumer_name        = "katello-ca-consumer-${::fqdn}"
   $candlepin_consumer_summary     = "Subscription-manager consumer certificate for Katello instance ${::fqdn}"
   $candlepin_consumer_description = 'Consumer certificate and post installation script that configures rhsm.'
 
+  $katello_default_ca_data = file("${certs::pki_dir}/certs/${certs::default_ca_name}.crt")
```
ehelms: You'd want to reference `certs::katello::default_ca_name`

parthaa:
@ehelms, @bbuckingham  green!

bbuckingham:
@parthaa, I tested this out on my katello-devel environment with an atomic host and it is behaving perfectly.  The full flow that i did was:
- import manifest w/ RH OStree content
- enable repo & sync OStree repo
- create content view - add repo to it, publish it, promote it
- create activation key - associate lifecycle environment, content view and subscriptions
- installed a rhel atomic host
- downloaded the katello-rhsm-consumer from katello  (this step and the ones that follow are the important ones wrt the cert)
- subscription-manager register the host to katello
- performed basic sanity testing of atomic host (e.g. atomic host status, upgrade, ...etc)

Note: i haven't verified a RHEL7 host (non-atomic) yet, but plan to try that as well.

bbuckingham:
@parthaa, I ran a test with a RHEL 7 vagrant host.  I was able to download/install the bootstrap rpm, register the host, subscribe to custom content and download RPMs from katello.

bbuckingham:
@parthaa, ACK from me; however, I'd recommend that also get ACK from an additional dev that is more familiar with the puppet-certs module.

ehelms:
I'll take a quick pass today

ehelms:
Testing this against a fresh installation, I immediately run into the following:

```
[ERROR 2016-02-19 10:29:53 verbose]  Could not find any files from /etc/pki/katello/certs/katello-default-ca.crt at /usr/share/katello-installer/modules/certs/manifests/katello.pp:25 on node nightly.example.com
```

bbuckingham:
@ehelms, Good catch!  My install was a dev install where I re-applied the puppet-certs update and re-ran the katello-installer-devel.  (i.e. testing in a clean, but pre-existing environment)

parthaa:
@ehelms updated. If you can figure out a way to get custom types to work with my template let me know :)

parthaa:
@ehelms is the change I had to make to address your comment -> https://github.com/Katello/puppet-certs/pull/67/files#diff-aad4d37a11e8099dfec9d8ad8c38dec0R21

manifests/katello.pp
```diff
@@ -14,10 +15,16 @@
...
   $rhsm_ca_dir                    = '/etc/rhsm/ca'
+  $katello_rhsm_setup_script      = 'katello-rhsm-consumer'
+  $katello_rhsm_setup_script_location = "/usr/bin/${katello_rhsm_setup_script}"
+
   $candlepin_consumer_name        = "katello-ca-consumer-${::fqdn}"
   $candlepin_consumer_summary     = "Subscription-manager consumer certificate for Katello instance ${::fqdn}"
   $candlepin_consumer_description = 'Consumer certificate and post installation script that configures rhsm.'
 
+  $katello_default_ca_file = "${certs::pki_dir}/certs/${certs::katello::default_ca_name}.crt"
+  $katello_server_ca_file = "${certs::pki_dir}/certs/${certs::katello::server_ca_name}.crt"
```
ehelms: This is essentially redefining https://github.com/Katello/puppet-certs/blob/master/manifests/init.pp#L130 which you could just use to DRY this up.

manifests/katello.pp
```diff
@@ -31,25 +38,30 @@
     mode   => '0755',
   } ->
   # Placing the CA in the pub dir for trusting by a user in their browser
-  file { "${katello_www_pub_dir}/${certs::server_ca_name}.crt":
+  file { "${katello_www_pub_dir}/${certs::katello::server_ca_name}.crt":
     ensure  => file,
-    source  => "${certs::pki_dir}/certs/${certs::server_ca_name}.crt",
+    source  => "${certs::pki_dir}/certs/${certs::katello::server_ca_name}.crt",
```
ehelms: Could use https://github.com/Katello/puppet-certs/blob/master/manifests/init.pp#L130 here

manifests/katello.pp
```diff
@@ -31,25 +38,30 @@
...
     owner   => 'root',
     group   => 'root',
     mode    => '0644',
-    require => File["${certs::pki_dir}/certs/${certs::server_ca_name}.crt"],
+    require => File["${certs::pki_dir}/certs/${certs::katello::server_ca_name}.crt"],
```
ehelms: And here https://github.com/Katello/puppet-certs/blob/master/manifests/init.pp#L130

templates/rhsm-katello-reconfigure.erb
```diff
@@ -16,22 +16,26 @@
...
 #
 
+#copy the certificate to this variable KATELLO_SERVER_CA_DATA
+read -r -d '' KATELLO_SERVER_CA_DATA << EOM
+<%=  File.exist?(scope['katello_server_ca_file'])  ? File.open(scope['katello_server_ca_file']).read : ""%>
+EOM
+
+#copy the default ca certificate to this variable KATELLO_DEFAULT_CA_DATA
+read -r -d '' KATELLO_DEFAULT_CA_DATA << EOM
+<%=  File.exist?(scope['katello_default_ca_file'])  ? File.open(scope['katello_default_ca_file']).read : ""%>
```
ehelms: Nitpick extra spaces mixed in here

templates/rhsm-katello-reconfigure.erb
```diff
@@ -16,22 +16,26 @@
 # Configures rhsm on client. Called from the certificate RPM.
 #
 
+#copy the certificate to this variable KATELLO_SERVER_CA_DATA
+read -r -d '' KATELLO_SERVER_CA_DATA << EOM
+<%=  File.exist?(scope['katello_server_ca_file'])  ? File.open(scope['katello_server_ca_file']).read : ""%>
```
ehelms: Nitpick extra spaces mixed in here

ehelms:
Functionally this worked for me through and install and a regenerate.

ehelms:
ACK - thanks @parthaa

PR #69, opened 2016-01-15
add trustore_pass for candlepin

cristifalcas:
add support for truststore password for candlepin. This is in connections with puppet-katello PR#94 and puppet-candlepin PR#32


ehelms:
I'm not sure this is needed since we use the keystore password for both keystore and truststore. If they do have separate passwords via this change, you will need to update this file to deploy both files and ensure the truststore calls use the correct password.

cristifalcas:
I will close it then and fi the katello PR

PR #70, opened 2016-01-15
Fixes #13188 - Creates certificates for capsule authentication

cfouant:
Needed by https://github.com/Katello/puppet-katello/pull/95/

### DIFF
--- a/manifests/pulp_client.pp
+++ b/manifests/pulp_client.pp
@@ -1,0 +1,47 @@
+# Pulp Client Certs
+class certs::pulp_client (
+  $hostname    = $::certs::node_fqdn,
+  $generate    = $::certs::generate,
+  $regenerate  = $::certs::regenerate,
+  $deploy      = $::certs::deploy,
+  $common_name = 'admin',
+  ){
+
+  $client_cert_name = "${hostname}-pulp-client"
+  $client_cert      = "${::certs::pki_dir}/certs/${client_cert_name}.crt"
+  $client_key       = "${::certs::pki_dir}/private/${client_cert_name}.key"
+  $ssl_ca_cert      = $::certs::ca_cert
+
+  cert { $client_cert_name:
+    hostname      => $hostname,
+    common_name   => $common_name,
+    purpose       => client,
+    country       => $::certs::country,
+    state         => $::certs::state,
+    city          => $::certs::city,
+    org           => 'PULP',
+    org_unit      => 'NODES',
+    expiration    => $::certs::expiration,
+    ca            => $::certs::default_ca,
+    generate      => $generate,
+    regenerate    => $regenerate,
+    deploy        => $deploy,
+    password_file => $::certs::ca_key_password_file,
+  }
+
+  if $deploy {
+    Cert[$client_cert_name] ~>
+    pubkey { $client_cert:
+      key_pair => Cert[$client_cert_name],
+    } ~>
+    privkey { $client_key:
+      key_pair => Cert[$client_cert_name],
+    } ~>
+    file { $client_key:
+      group => $katello::group,
+      owner => 'root',
+      mode  => '0440',
+    }
+  }
+}
+


manifests/katello.pp
```diff
@@ -47,4 +50,40 @@
...
+    city          => $::certs::city,
+    org           => 'KATELLO',
+    org_unit      => 'CAPSULE',
+    expiration    => $::certs::expiration,
+    ca            => $::certs::default_ca,
+    generate      => $generate,
+    regenerate    => $regenerate,
+    deploy        => $deploy,
+    password_file => $certs::ca_key_password_file,
+  }
```
jlsherrill: I think this is missing the common name attribute that should be the default admin user for pulp

manifests/katello.pp
```diff
@@ -47,4 +50,40 @@
...
+    city          => $::certs::city,
+    org           => 'KATELLO',
+    org_unit      => 'CAPSULE',
+    expiration    => $::certs::expiration,
+    ca            => $::certs::default_ca,
+    generate      => $generate,
+    regenerate    => $regenerate,
+    deploy        => $deploy,
+    password_file => $certs::ca_key_password_file,
+    common_name   => $::pulp::default_login,
```
ehelms: We shouldn't tie this in to the pulp module, but rather have this as a parameter that can be passed in

jlsherrill: And to clarify that would require a small change to puppet-capsule to pass in the parameter

ehelms:
I think you will only want to deploy this on the server itself; given certs::katello is called for every Capsule installation, you'll be best off creating a separate manifest for this generation or putting it in the pulp.pp section. And then adjusting puppet-katello and not puppet-capsule.

cfouant:
@ehelms - changes made, let me know what you think.

manifests/pulp_client.pp
```diff
@@ -0,0 +1,48 @@
+# Pulp Client Certs
+class certs::pulp_client (
+  $hostname                          = $::certs::node_fqdn,
+  $generate                          = $::certs::generate,
+  $regenerate                        = $::certs::regenerate,
+  $deploy                            = $::certs::deploy,
+
+  $common_name                       = $certs::common_name
```
ekohl: Nitpick: make it consistent with the others and use ```$::certs```. I'm also a fan of trailing comma's because it keeps the diff smaller if another parameter is added.

manifests/pulp_client.pp
```diff
@@ -0,0 +1,48 @@
...
+    state         => $::certs::state,
+    city          => $::certs::city,
+    org           => 'PULP',
+    org_unit      => 'NODES',
+    expiration    => $::certs::expiration,
+    ca            => $::certs::default_ca,
+    generate      => $generate,
+    regenerate    => $regenerate,
+    deploy        => $deploy,
+    password_file => $certs::ca_key_password_file,
```
ekohl: Also ```$::certs```

cfouant:
@ekohl - fixed, thanks!

manifests/pulp_client.pp
```diff
@@ -0,0 +1,48 @@
+# Pulp Client Certs
+class certs::pulp_client (
+  $hostname                          = $::certs::node_fqdn,
+  $generate                          = $::certs::generate,
+  $regenerate                        = $::certs::regenerate,
+  $deploy                            = $::certs::deploy,
+
```
ehelms: unneeded blank line

manifests/pulp_client.pp
```diff
@@ -0,0 +1,48 @@
...
+  $common_name                       = $::certs::common_name
```
ehelms: Could you reduce the horizontal space between the parameters and their values here?

ehelms: I wonder if we shouldn't default `$common_name` to 'admin' here. Given the common_name very much matters for doing client authentication to Pulp, using the default would end up in a broken certificate as I understand things. Using 'admin' at least would default this to the de-facto pulp admin login name and provide a sane certificate if the user didn't pass in a value.

ehelms:
@cfouant the recent change hardcodes it to 'admin' versus setting the default to admin. The difference being where it is set. This is more what I envisioned:

```class certs::pulp_client (
  $hostname    = $::certs::node_fqdn,
  $generate    = $::certs::generate,
  $regenerate  = $::certs::regenerate,
  $deploy      = $::certs::deploy,
  $common_name = 'admin',
  ){

  $client_cert_name = "${hostname}-pulp-client"
  $client_cert      = "${::certs::pki_dir}/certs/${client_cert_name}.crt"
  $client_key       = "${::certs::pki_dir}/private/${client_cert_name}.key"
  $ssl_ca_cert      = $::certs::ca_cert

  cert { $client_cert_name:
    hostname      => $hostname,
    common_name   => $common_name,
    purpose       => client,
    country       => $::certs::country,
    state         => $::certs::state,
    city          => $::certs::city,
    org           => 'PULP',
    org_unit      => 'NODES',
    expiration    => $::certs::expiration,
    ca            => $::certs::default_ca,
    generate      => $generate,
    regenerate    => $regenerate,
    deploy        => $deploy,
    password_file => $::certs::ca_key_password_file,
  }
```
This way it is still configurable, but also has a sane default.

cfouant:
@ehelms - can't change the labels, but changes have been made.

ehelms:
ACK - thanks @cfouant !

ehelms:
Blast I overlooked this -- https://github.com/Katello/puppet-certs/pull/70/files#diff-939f2bb1027179e9fc9e8a0e22f1d8aeR41

COuld you open a small PR to change that to `certs::group` ?

cfouant:
@ehelms - https://github.com/Katello/puppet-certs/pull/71/

PR #72, opened 2016-02-09
Fixes #13634 - Adding Katello cert to ca-trust

parthaa:
Adding Katello Server cert to the machines local ca-trust

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -1,8 +1,9 @@
 fixtures:
   repositories:
-    stdlib:        "git://github.com/puppetlabs/puppetlabs-stdlib.git"
-    extlib:        "git://github.com/puppet-community/puppet-extlib.git"
-    foreman:       "git://github.com/theforeman/puppet-foreman.git"
-    common:        "git://github.com/katello/puppet-common.git"
+    stdlib:            "git://github.com/puppetlabs/puppetlabs-stdlib.git"
+    extlib:            "git://github.com/puppet-community/puppet-extlib.git"
+    foreman:           "git://github.com/theforeman/puppet-foreman.git"
+    common:            "git://github.com/katello/puppet-common.git"
+    trusted_ca:        "git://github.com/evenup/evenup-trusted_ca.git"
   symlinks:
     certs: "#{source_dir}"
--- a/manifests/katello.pp
+++ b/manifests/katello.pp
@@ -18,6 +18,12 @@ class certs::katello (
   $candlepin_consumer_summary     = "Subscription-manager consumer certificate for Katello instance ${::fqdn}"
   $candlepin_consumer_description = 'Consumer certificate and post installation script that configures rhsm.'
 
+  include ::trusted_ca
+  trusted_ca::ca { 'katello_server-host-cert':
+    source  => $certs::katello_server_ca_cert,
+    require => File[$certs::katello_server_ca_cert],
+  }
+
   file { $katello_www_pub_dir:
     ensure => directory,
     owner  => 'apache',
--- a/metadata.json
+++ b/metadata.json
@@ -9,6 +9,10 @@
   "issues_url": "http://projects.theforeman.org/projects/katello/issues",
   "dependencies": [
     {
+      "name": "evenup-trusted_ca",
+      "version_requirement": ">= 1.0.1 < 2.0.0"
+    },
+    {
       "name": "puppetlabs-stdlib",
       "version_requirement": ">= 4.2.0 < 5.0.0"
     },
--- a/spec/classes/certs_katello_spec.rb
+++ b/spec/classes/certs_katello_spec.rb
@@ -1,0 +1,27 @@
+require 'spec_helper'
+
+describe 'certs::katello' do
+  let :facts do
+    {
+      :concat_basedir             => '/tmp',
+      :interfaces                 => '',
+      :operatingsystem            => 'RedHat',
+      :operatingsystemrelease     => '6',
+      :operatingsystemmajrelease  => '6',
+      :osfamily                   => 'RedHat',
+      :fqdn                       => 'pulp.compony.net',
+      :hostname                   => 'pulp',
+    }
+  end
+
+  context 'with parameters' do
+    let :pre_condition do
+      "class {'certs': pki_dir => '/tmp', server_ca_name => 'foo'}"
+    end
+
+    describe 'with katello certs set' do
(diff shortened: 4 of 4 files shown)


manifests/init.pp
```diff
@@ -129,6 +129,7 @@
 
   $katello_server_ca_cert = "${certs::pki_dir}/certs/${server_ca_name}.crt"
 
+  class { 'trusted_ca': } ->
```
ekohl: This should be ```::trusted_ca```

ekohl:
Dependencies should also be in .fixtures.yml, preferably their git versions so we can spot failures sooner.

metadata.json
```diff
@@ -9,6 +9,10 @@
   "issues_url": "http://projects.theforeman.org/projects/katello/issues",
   "dependencies": [
     {
+      "name": "evenup-trusted_ca",
+      "version_requirement": ">= 1.0.1"
```
ehelms: Please add `< 2.0.0` to the end to help us avoid major version breakages in the future.

manifests/katello.pp
```diff
@@ -18,6 +18,12 @@
   $candlepin_consumer_summary     = "Subscription-manager consumer certificate for Katello instance ${::fqdn}"
   $candlepin_consumer_description = 'Consumer certificate and post installation script that configures rhsm.'
 
+  class { 'trusted_ca': }
```
ehelms: Is this needed if you are using the defined type? If it is needed, can you use include instead?

manifests/katello.pp
```diff
@@ -18,6 +18,12 @@
   $candlepin_consumer_summary     = "Subscription-manager consumer certificate for Katello instance ${::fqdn}"
   $candlepin_consumer_description = 'Consumer certificate and post installation script that configures rhsm.'
 
+  class { ::trusted_ca }
+  trusted_ca::ca { 'katello_server-host-cert':
+    source => "${certs::pki_dir}/certs/${certs::server_ca_name}.crt",
+    require => File["${certs::pki_dir}/certs/${certs::server_ca_name}.crt"],
```
ehelms: Is it able to resolve `$::certs::katello_server_ca_cert` here?

parthaa:
> Dependencies should also be in .fixtures.yml, preferably their git versions so we can spot failures sooner.

thanks will look into that

manifests/katello.pp
parthaa: Will test your suggestion. But I got the variables from 2 lines below this.

.fixtures.yml
```diff
@@ -4,5 +4,6 @@ fixtures:
     extlib:        "git://github.com/puppet-community/puppet-extlib.git"
     foreman:       "git://github.com/theforeman/puppet-foreman.git"
     common:        "git://github.com/katello/puppet-common.git"
+    evenup:        "git@github.com:evenup/evenup-trusted_ca.git"
```
ekohl: Please use ```git://``` because it's anonymous.

manifests/katello.pp
```diff
@@ -18,6 +18,12 @@
   $candlepin_consumer_summary     = "Subscription-manager consumer certificate for Katello instance ${::fqdn}"
   $candlepin_consumer_description = 'Consumer certificate and post installation script that configures rhsm.'
 
+  class { '::trusted_ca': } ->
+  trusted_ca::ca { 'katello_server-host-cert':
+    source => $certs::katello_server_ca_cert,
```
ekohl: Please align the parameters.

parthaa:
> thanks will look into that

@ekohl  added

.fixtures.yml
```diff
@@ -4,5 +4,6 @@ fixtures:
     extlib:        "git://github.com/puppet-community/puppet-extlib.git"
     foreman:       "git://github.com/theforeman/puppet-foreman.git"
     common:        "git://github.com/katello/puppet-common.git"
+    evenup:        "git://github.com/evenup/evenup-trusted_ca.git"
```
ehelms: Ideally the name here would be the name of the puppet module and not the provider of the module, i.e. trusted_ca and not evenup

manifests/katello.pp
```diff
@@ -18,6 +18,12 @@
   $candlepin_consumer_summary     = "Subscription-manager consumer certificate for Katello instance ${::fqdn}"
   $candlepin_consumer_description = 'Consumer certificate and post installation script that configures rhsm.'
 
+  class { '::trusted_ca': }
```
ehelms: Can this be done via `include ::trustedca` since we aren't setting any parameters?

parthaa:
Also tested this the following way

- Installed katello-devel
- saw this file ```/etc/pki/ca-trust/source/anchors/katello_server-host-cert.crt```
- tried ```curl https://<fqdn>``` . It responded with a 503 but not a 401, insecure.
- I then did
```$ rm -f /etc/pki/ca-trust/source/anchors/katello_server-host-cert.crt
   $ update-ca-trust
```
- Again tried ```curl https://<fqdn>``` 
- It responded with a 401, insecure error.

manifests/katello.pp
parthaa: Is there a precedent for this ?

parthaa: will try

spec/classes/katello_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'spec_helper'
```
ehelms: Nitpick: typically the test name follows the full manifest name as a convention, i.e. `certs_katello_spec.rb`

ehelms:
Just to double check, we want this to be deployed on both the main server and all Capsules is that correct? Currently, the capsule uses the `katello.pp` manifest -- https://github.com/Katello/puppet-capsule/blob/f05a17386fc7575d7944ea752853696af3ebf6d3/manifests/init.pp#L104

parthaa:
> Just to double check, we want this to be deployed on both the main server and all Capsules is that correct? Currently, the capsule uses the katello.pp manifest -- https://github.com/Katello/puppet-capsule/blob/f05a17386fc7575d7944ea752853696af3ebf6d3/manifests/init.pp#L104

Cert should probably be trusted only the main server for post sync to work afaik.  @jlsherrill with the nodes work do think this will be useful in a capsule?  
One thing running this in a capsule would be if customer wants to turn that capsule into a Docker node.

ehelms:
ACK - thanks @parthaa !

PR #73, opened 2016-02-09
Fixes #13635 - set pulp client cert settings

johnpmitsch:
Removing the hostname from the client cert and key path as that makes it harder to configure in Katello, and setting the pulp client path in the Settings.

Related to https://github.com/Katello/katello/pull/5750

### DIFF
--- a/manifests/pulp_client.pp
+++ b/manifests/pulp_client.pp
@@ -7,7 +7,7 @@ class certs::pulp_client (
   $common_name = 'admin',
   ){
 
-  $client_cert_name = "${hostname}-pulp-client"
+  $client_cert_name = 'pulp-client'
   $client_cert      = "${::certs::pki_dir}/certs/${client_cert_name}.crt"
   $client_key       = "${::certs::pki_dir}/private/${client_cert_name}.key"
   $ssl_ca_cert      = $::certs::ca_cert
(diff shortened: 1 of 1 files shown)


manifests/pulp_client.pp
```diff
@@ -7,7 +7,7 @@
   $common_name = 'admin',
   ){
 
-  $client_cert_name = "${hostname}-pulp-client"
+  $client_cert_name = "pulp-client"
```
ekohl: These should be single quotes because there's no variable anymore.

manifests/pulp_client.pp
```diff
@@ -43,5 +43,14 @@
       mode  => '0440',
     }
   }
-}
 
+  foreman_config_entry { 'pulp_client_cert':
+    value => $client_cert,
```
ekohl: Please align these.

manifests/pulp_client.pp
```diff
@@ -43,5 +43,14 @@
...
+    ignore_missing => false
```
ekohl: We prefer trailing comma's because it gives a smaller diff when we add additional parameters.

manifests/pulp_client.pp
```diff
@@ -43,5 +43,14 @@
...
+  }
+
+  foreman_config_entry { 'pulp_client_key':
+    value => $client_key,
```
ekohl: Same comments as the above config entry.

johnpmitsch:
@ekohl thanks for the feedback, updated

ehelms:
ACK - thanks @johnpmitsch !

PR #74, opened 2016-02-10
[WORK IN PROGRESS] Fixes #13451 - adds squid to the installer

cfouant:


ehelms:
Why does the certs module need squid?

cfouant:
oops wrong repo :-1:

PR #75, opened 2016-02-10
Fixes #13658 - pulp_client_key and pulp_client_cert not being set cor…

johnpmitsch:
…rectly

### DIFF
--- a/manifests/pulp_client.pp
+++ b/manifests/pulp_client.pp
@@ -44,13 +44,4 @@ class certs::pulp_client (
     }
   }
 
-  foreman_config_entry { 'pulp_client_cert':
-    value          => $client_cert,
-    ignore_missing => false,
-  }
-
-  foreman_config_entry { 'pulp_client_key':
-    value          => $client_key,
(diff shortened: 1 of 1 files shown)


manifests/config.pp
```diff
@@ -22,4 +22,16 @@
...
+    value          => $client_cert,
+    ignore_missing => false,
+    before => Exec['foreman-rake-db:migrate'],
+  }
+
+  foreman_config_entry { 'pulp_client_key':
+    value          => $client_key,
+    ignore_missing => false,
+    before => Exec['foreman-rake-db:migrate'],
+  }
```
jlsherrill: @ehelms  was actually suggesting it go here:  https://github.com/Katello/puppet-katello/blob/master/manifests/config.pp

I'll let him comment if he cares one way or the other

johnpmitsch:
@jlsherrill @ehelms moved it to here https://github.com/Katello/puppet-katello/pull/104/files

ehelms:
ACK - thanks @johnpmitsch !

PR #76, opened 2016-02-25
fix bootstrap rpm in katello_devel install

jlsherrill:
in a katello_devel install, the variables were not properly looked up
and the ca certificate had not been written to the filesystem yet


ehelms:
ACK

PR #77, opened 2016-02-27
Fixes #13925: Use concat to build reconfigure script

ehelms:
On Puppet 3.6 the use of File.read in the reconfigure template
causes the certs to be empty given the template is evaluated
at the beginning. This switches to concat to build up the reconfigure
script using the source parameter to read from the cert files.

### DIFF
--- a/manifests/katello.pp
+++ b/manifests/katello.pp
@@ -41,14 +41,9 @@ class certs::katello (
     mode    => '0644',
     require => File[$certs::katello_server_ca_cert],
   } ~>
-  # Generate the the rhsm setup script in the pub dir for rhsm setup
-  file { "${katello_www_pub_dir}/${katello_rhsm_setup_script}":
-    ensure  => file,
-    owner   => 'root',
-    group   => 'root',
-    mode    => '0755',
-    content => template('certs/rhsm-katello-reconfigure.erb'),
-    require => Class['::certs'],
+  certs::rhsm_reconfigure_script { "${katello_www_pub_dir}/${katello_rhsm_setup_script}":
+    ca_cert        => $::certs::ca_cert,
+    server_ca_cert => $::certs::katello_server_ca_cert,
   } ~>
   certs_bootstrap_rpm { $candlepin_consumer_name:
     dir              => $katello_www_pub_dir,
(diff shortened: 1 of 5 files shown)


ehelms:
This is not the cleanest of solutions, but it works and I think feels cleaner than the previous solution given it uses puppet constructs.

ekohl:
I'm having a hard time wrapping my head around the logic here. As I read it it generates a script (with the inline template) on the server side, but the certificate is on the client side. The server doesn't know the CAs in advance. Do I understand it correctly? Is it an option to read the certificate at runtime instead of statically or is the script shipped to others and does it have to be inline?

ehelms:
I'll try my best to explain and let @parthaa fill in the details. On a normal RPM based system, we just needed the RPM that could include a script to reconfigure RHSM and the certificates. On an atomic based system, there are no RPMs so the best we can do is have a script that reconfigures RHSM and we opted to embed the certificates instead of having the script download them or require the user to download them (i.e. an all in one package similar to the RPM).

manifests/katello.pp
```diff
@@ -41,14 +41,9 @@
...
   } ~>
-  # Generate the the rhsm setup script in the pub dir for rhsm setup
-  file { "${katello_www_pub_dir}/${katello_rhsm_setup_script}":
-    ensure  => file,
-    owner   => 'root',
-    group   => 'root',
-    mode    => '0755',
-    content => template('certs/rhsm-katello-reconfigure.erb'),
-    require => Class['::certs'],
+  rhsm_reconfigure_script { "${katello_www_pub_dir}/${katello_rhsm_setup_script}":
```
ekohl: Shouldn't this be certs::rhsm_reconfigure_script?

ehelms:
Updated to include `certs::`

ekohl:
@ehelms I think it makes sense and I agree it's cleaner than the previous solution.

parthaa:
@ehelms I definitely like this solution. Will test and ack if it works

parthaa:
Tested on both Atomic and Non-Atomic Rhel 7 . Once via rpm, and once via script direct.

ACK

PR #83, opened 2016-04-20
Fixes #12266: Handle last RPM sort for more than 10 bootstrap RPMs

ehelms:
Old method of sorting used string based comparisons which after 10
release versions fails to sort properly. This moves to a method of
comparing individual releases via integer comparison to derive the
latest boostrap RPM.

### DIFF
--- a/lib/puppet/provider/certs_bootstrap_rpm/katello_ssl_tool.rb
+++ b/lib/puppet/provider/certs_bootstrap_rpm/katello_ssl_tool.rb
@@ -33,7 +33,17 @@ Puppet::Type.type(:certs_bootstrap_rpm).provide(:katello_ssl_tool) do
   protected
 
   def last_rpm
-    Dir.glob(File.join(resource[:dir], "#{resource[:name]}-*.noarch.rpm")).sort.last
+    rpms = Dir.glob(File.join(resource[:dir], "#{resource[:name]}-*.noarch.rpm"))
+
+    rpms = rpms.collect do |rpm|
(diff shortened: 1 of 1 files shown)


lib/puppet/provider/certs_bootstrap_rpm/katello_ssl_tool.rb
```diff
@@ -33,7 +33,17 @@ def run
...
+
+    rpms = rpms.collect do |rpm|
+      rpm_split = rpm.split("#{resource[:name]}-")[1].split('.noarch.rpm')[0]
+      version = rpm_split.split('-')[0]
+      release = rpm_split.split('-')[1]
+
+      {'release' => release, 'rpm' => rpm}
+    end
+
+    rpms.sort { |a,b| a['release'].to_i <=> b['release'].to_i }.last
```
stbenjam: Does the RPM version never change, always 1.0? Could that change some day?

The last RPM ends up as 1.0.11 when we have: 

```ruby
["foo-1.0-1.noarch.rpm", "foo-2.0-2.noarch.rpm", "foo-1.0-11.noarch.rpm"]
```

ehelms: AFAIK we don't increment the version (https://github.com/Katello/puppet-certs/pull/83/files/e6df6f10f485e0bf283c428c371fa2606b7faaa5#diff-d7c04bc14bf477574cb5d1f4d285675cR13) only the release when we generate a new bootstrap RPM. So possible that logic could change down the road, but I don't forsee it :)

stbenjam:
ACK

PR #86, opened 2016-04-27
Refs #12266 - fixing case where no certs exist

jlsherrill:
undefined method [] for nil:NilClass
/usr/share/katello-installer-base/modules/certs/lib/puppet/provider/certs_bootstrap_rpm/katello_ssl_tool.rb:46:in last_rpm

### DIFF
--- a/lib/puppet/provider/certs_bootstrap_rpm/katello_ssl_tool.rb
+++ b/lib/puppet/provider/certs_bootstrap_rpm/katello_ssl_tool.rb
@@ -20,9 +20,9 @@ Puppet::Type.type(:certs_bootstrap_rpm).provide(:katello_ssl_tool) do
                             '--requires', 'subscription-manager',
                             '--post', post_script_file,
                             *resource[:files])
-      if resource[:alias]
+      if (rpm = last_rpm) && resource[:alias]
         File.delete(resource[:alias]) if File.exists?(resource[:alias])
-        File.symlink(last_rpm, resource[:alias])
+        File.symlink(rpm, resource[:alias])
       end
       system('/sbin/restorecon ./*.rpm')
     end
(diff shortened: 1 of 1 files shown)


lib/puppet/provider/certs_bootstrap_rpm/katello_ssl_tool.rb
```diff
@@ -43,7 +43,8 @@ def last_rpm
       {'release' => release, 'rpm' => rpm}
     end
 
-    rpms.sort { |a,b| a['release'].to_i <=> b['release'].to_i }.last['rpm']
+    rpm = rpms.sort { |a,b| a['release'].to_i <=> b['release'].to_i }.last
+    rpm['rpm'] if rpm
```
stbenjam: I think this will cause an error above - when we make the symlink on L25.  Probably need to gate that with a  `last_rpm.blank?` check

jlsherrill: based on what it used to be:   https://github.com/Katello/puppet-certs/pull/83/files  it seems like it handles nil okay, but i can skip up above if its nill to be more explicit.

jlsherrill: actually after testing File.symlink() does not like nil.  How was this working before?  very strange

jlsherrill:
@stbenjam updated

jlsherrill:
tested and this seemed to get past this error, but other errors still occured

lib/puppet/provider/certs_bootstrap_rpm/katello_ssl_tool.rb
```diff
@@ -20,9 +20,9 @@ def run
                             '--requires', 'subscription-manager',
                             '--post', post_script_file,
                             *resource[:files])
-      if resource[:alias]
+      if rpm = last_rpm && resource[:alias]
```
stbenjam: Nitpick, but usually assignment in a conditional should be in parens.  Do we enforce that?

stbenjam:
Otherwise good for me, were the additional errors still related to this? Or something else?  

From a code perspective I think this change is fine, but didn't test yet.

jlsherrill:
@stbenjam updated with parens.

The errors were unrelated, likely related to https://github.com/theforeman/puppet-foreman/pull/436

stbenjam:
ACK

PR #87, opened 2016-05-12
Use foreman_config_entry instead of manual rake config

ekohl:
I should note that I haven't tested this yet, this is mostly to check if this is even a good idea. https://github.com/theforeman/puppet-foreman/issues/441 also suggested that this should be part of puppet-foreman itself.

### DIFF
--- a/manifests/foreman.pp
+++ b/manifests/foreman.pp
@@ -48,19 +48,19 @@ class certs::foreman (
       mode   => '0400',
     }
 
-    $foreman_config_cmd = "${::foreman::app_root}/script/foreman-rake config --\
-      -k ssl_ca_file -v '${ssl_ca_cert}'\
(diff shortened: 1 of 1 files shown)


ekohl:
https://github.com/theforeman/puppet-foreman/pull/442 is an alternative approach. Using that would mean simply setting the correct answers.

ehelms:
I can test this out with our module-pr tester in kdeploy. As for the parameters, I'd still probably learn towards this approach given we'd have to define the certs in the answers file of the installer and then they are hard-coded as opposed to this approach that will update them if anything changes.

ekohl:
The current certificate locations are already hardcoded in the katello-installer answers file. By implementing this, you reduce coupling which I think is a good thing.

stbenjam:
Closed in favor of https://github.com/Katello/puppet-certs/pull/88

PR #91, opened 2016-06-28
Fixes #15538 - make sure the rpms from ssl-build are used

iNecas:
Before this patch, we were relying on the fact that the latest rpms
are always better. However, people often try to rollback to the
previous state of ssl-build and this behavior of the certs script was
causing more troubles than benefits.

After this change, we always use the latest version we have available
in ssl-build by checking if that's what's already installed on the
system or not.

While trying to rollback to some older version of certs, I was hitting
the nssdb errors, as we were not cleaning the certs in there properly.
Therefore I've reused the resource we already had there for certutil,
to clean up certs first.

Steps to test:

  1 install katello
  2 check the certificiate of web UI
  3 `cp ~/ssl-build{,.1}`
  4 `foreman-installer --certs-update-all`
  5 check the certificiate of web UI
  6 `cp ~/ssl-build{,.2}`
  7 `rm -rf ~/ssl-build`
  8 `cp ~/ssl-build{.1,}`
  9 `foreman-installer`
  10 the certificate of the web UI should change back to the one from step 2
  11 `foreman-installer --certs-update-all`
  12 the certificate of the web UI should be different than the one from step 2 or 5

### DIFF
--- a/lib/puppet/provider/katello_ssl_tool.rb
+++ b/lib/puppet/provider/katello_ssl_tool.rb
@@ -62,7 +62,7 @@ module Puppet::Provider::KatelloSslTool
       return false unless resource[:deploy]
       return true if resource[:regenerate]
       return true if files_to_deploy.any? { |file| ! File.exist?(file) }
-      return true if new_version_available?
+      return true if needs_deploy?
     end
 
     def files_to_deploy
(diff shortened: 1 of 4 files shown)


lib/puppet/provider/katello_ssl_tool.rb
```diff
@@ -81,9 +83,8 @@ def deploy!
 
     def new_version_available?
```
ekohl: I don't think this function covers what it needs to do anymore. Something like ```needs_deploy?``` feels more appropriate now.

iNecas: Good call, updated

ekohl:
Untested, but :+1: for the code.

ehelms:
Running through the test scenario

ehelms:
ACK - thanks @iNecas !

PR #92, opened 2016-07-06
Refs #15538: Check for nssdb cert as the beginning of a line

ehelms:
When using something short and non-specific like 'ca' to grep the
nssdb output, other words in the nssdb can pass the test. For example,

[root@sat-r220-02 ~]# certutil -L -d /etc/pki/katello/nssdb

Certificate Nickname                               Trust Attributes
                                                   SSL,S/MIME,JAR/XPI

amqp-client                                                  ,,
broker                                                       u,u,u

The 'ca' in 'Certificate Nickname' passes the 'grep ca' test.

### DIFF
--- a/manifests/ssltools/certutil.pp
+++ b/manifests/ssltools/certutil.pp
@@ -4,14 +4,14 @@ define certs::ssltools::certutil($nss_db_dir, $client_cert, $cert_name=$title, $
   exec { "delete ${cert_name}":
     path        => ['/bin', '/usr/bin'],
     command     => "certutil -D -d ${nss_db_dir} -n '${cert_name}'",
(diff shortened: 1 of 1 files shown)


stbenjam:
How about a `\s` too just to be safe?  



In the very unlikely event the CA is named  `Certificate` or `SSL` something


```
[root@sat-r220-06 ~]# certutil -L -d /etc/pki/katello/nssdb | grep 'ca\s'
ca                                                           CT,C,c
```

ehelms:
Updated

stbenjam:
APJ

iNecas:
:+1: failure tests unrelated

ekohl:
Tests should pass after a rebase

PR #95, opened 2016-07-28
fixes #15882 - support AIO paths for puppet

stbenjam:


beav:
testing this now

stbenjam:
This makes the ENC work, so `puppet agent --test` works on the Katello

beav:
APJ, works for me. I will remove the hack this clears from https://github.com/Katello/forklift/pull/260 once merged.

ekohl:
For non-AIO installs this changes the location from ```/etc/puppet``` to (the default) ```/var/lib/puppet/ssl```. Will this break existing installs on upgrades?

stbenjam:
> For non-AIO installs this changes the location from /etc/puppet to (the default) /var/lib/puppet/ssl. Will this break existing installs on upgrades?

/etc/puppet/ssl is just a symlink to /var/lib/puppet/ssl, but I can leave it how we have it now, it was mostly just because I copied foreman's code for this

stbenjam:
Oh I see what you mean, katello was putting the certs into just /etc/puppet.  Not sure that makes sense, they should go to the ssl dir.

ekohl:
I'm fine with the code as is it now, just concerned about upgrading users. Sounds like a major version bump with a note in the release notes (and maybe katello-installer needs to migrate the certs for upgrading users).

stbenjam:
There shouldn't be an upgrade issue, the installer will fix it.  We keep generated certs in /root/ssl-build and deploy them to these paths that puppet tells us, and anyone using the paths references the $::certs variables.

ekohl:
ACK then

stbenjam:
Thanks!

PR #96, opened 2016-07-28
Use absolute variable $::candlepin::tomcat

ekohl:


beav:
this appears to not populate in the string:

```
 /Stage[main]/Certs::Candlepin/File[/usr/share//conf/keystore]/ensure: change from absent to link failed: Could not set 'link' on ensure: No such file or directory @ dir_chdir - /usr/share/conf at 61:/usr/share/katello-installer-base/modules/certs/manifests/candlepin.pp
```

ekohl:
Sounds like the candlepin class isn't included.

beav:
I added an include for candlepin to `candlepin.pp`, but got a duplicate class def error.

Is it an OK practice to just redefine the `tomcat/tomcat6` selector logic here, vs calling out to the candlepin module? My puppet is a little rusty.

ekohl:
I think including candlepin is a bad idea so I think I'd prefer a class parameter ```$tomcat_keystore``` which takes it from ```certs::params```. I think we can live with the duplication.

stbenjam:
https://github.com/Katello/puppet-certs/pull/98

PR #97, opened 2016-07-28
Update rhsm config template for puppet 4

beav:
Previously, when generating the client cert RPM, some of these values were
blank due to differences in puppet 3 and 4. This caused clients to not be able
to register.


ekohl:
ACK

PR #98, opened 2016-08-02
Move tomcat name logic to puppet-certs

stbenjam:


ekohl:
It appears the tomcat variables are already present in params.pp.

stbenjam:
Hah, thanks, I didn't notice - fixed.

beav:
that is funny, I did not notice that either.

There is a default of 'tomcat' set by the candlepin module (https://github.com/Katello/puppet-candlepin/blob/master/manifests/params.pp#L11). I don't think it can be hit though, since there's already a default defined in the lines above.

ACK

PR #99, opened 2016-08-04
Use /etc/puppet for client SSL certificates

stbenjam:
Trying a puppet 3 install, and I get this error:

```
[ERROR 2016-08-04 10:36:46 main]  Could not set 'present' on ensure: No such file or directory - /etc/puppet/ssl/client_cert.pem at 39:/usr/share/katello-installer-base/modules/certs/manifests/puppet.pp
```

Looks like I was wrong (https://github.com/Katello/puppet-certs/pull/95#issuecomment-235951864) about /etc/puppet/ssl existing, just reverting it to the original place we were storing these for puppet 3.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -73,7 +73,7 @@ class certs::params {
 
   $puppet_ssldir = $aio_package ? {
     true    => '/etc/puppetlabs/puppet/ssl',
-    default => '/etc/puppet/ssl'
+    default => '/etc/puppet'
   }
 
   $puppet_client_cert = "${puppet_ssldir}/client_cert.pem"


ekohl:
Any particular reason to store this in /etc/puppet rather than the default ```/var/lib/puppet/ssl```? I don't mind either way.

stbenjam:
I'd rather just leave them where it was, since the focus is getting p4 to work. These are the certs used for the foreman node.rb thingy and it's in /etc/puppet too.

ekohl:
ACK, this means it's no change compared to the previous version which is a good thing for admins :)

PR #101, opened 2016-08-23
refs #10283 - mark parameters advanced

stbenjam:
Anything other than "Parameters" are not shown by default in Kafo help.
This makes the installer help shorter, and reduces the chance users will
change options that really shouldn't be changed unless you really really
really know what you're doing.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -6,10 +6,24 @@
 #
 # === Parameters:
 #
-# $log_dir::              When the log files should go
-#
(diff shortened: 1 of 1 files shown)


stbenjam:
Are there any commonly used certs options? This hides them all.

ehelms:
If anything, the server cert options for setting custom certificates

ehelms:
@stbenjam ping

stbenjam:
Updated

PR #105, opened 2016-09-13
Bump foreman dependency

beav:
There are other modules that require `foreman-theforeman > 6.0.0`,
which would break when this module was imported into librarian.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -26,7 +26,7 @@
     },
     {
       "name": "theforeman-foreman",
-      "version_requirement": ">= 5.0.0 < 6.0.0"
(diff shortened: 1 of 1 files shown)


stbenjam:
ACK

metadata.json
```diff
@@ -26,7 +26,7 @@
     },
     {
       "name": "theforeman-foreman",
-      "version_requirement": ">= 5.0.0 < 6.0.0"
+      "version_requirement": ">= 6.0.0 < 7.0.0"
```
ekohl: I think it could have been ```>= 5.0.0 < 7.0.0``` but I don't think we have to change that now.

PR #107, opened 2016-10-14
fixes #16945 - use node_fqdn for consumer cert RPM generation

stbenjam:
In some custom setups (like Red Hat's HA ref arch), users may change the node_fqdn to match something like a load balancer's virtual hostname.  This ensures we use the node_fqdn variable if they did that, but otherwise node_fqdn default value is fqdn anyway.


ehelms:
ACK

stbenjam:
Thanks!

PR #113, opened 2016-11-30
Require puppet server installed before setting file user to puppet

ehelms:


ehelms:
I am testing this change to see if it affects the errors I am seeing on install

stbenjam:
I don't think this would work, `certs::puppet` would come before `puppet` class:

https://github.com/Katello/puppet-capsule/blob/master/manifests/init.pp#L222-L243

But, there should already be a puppet user on the box, as puppet is what's running the code in this class.

ehelms:
So, I wasn't sure this would work either, but I just got a successful install using this and https://github.com/Katello/puppet-katello/pull/156 . If you think the uid is a safer option, I can re-tool the PR and re-run my test. Been so long since I saw success on an install.

stbenjam:
I'm really surprised that works... :-\  Puppet must be able to resolve the dependencies, even though it looks like a loop to me.

IMHO the UID is probably safer, maybe @ekohl has a suggestion

ehelms:
OK - so I tested this with the update to use '52' as the user for the UID and that gave me the same green result testing installs locally.

ekohl:
IIRC puppet 4 client no longer creates the user, just the server.

I'd not be happy with using an ID because not every OS guarantees that to be the same. I know at least Debian doesn't and I don't know if puppetlabs will commit itself to the same UID. Not sure if I have a better idea though.

ehelms:
Since we only support CentOS with this today, I'd like us to just take this as is to fix the issues and we can re-visit this when we get to other OSes, or what I think is more likely to happen this all gets re-factored.

stbenjam:
> I'd not be happy with using an ID because not every OS guarantees that to be the same. I know at least Debian doesn't and I don't know if puppetlabs will commit itself to the same UID. Not sure if I have a better idea though.

UID's are guaranteed on Red Hat-based OS's, puppet is forever enshrined as uid 52: https://git.fedorahosted.org/cgit/setup.git/tree/uidgid#n61

stbenjam:
Although, the long term goal should be to support other OS's, and eventually maybe puppet-certs exists in some form that works with Foreman standalone...   this isn't ideal.

But for now, especially to avoid cross-module dependencies as much as possible, the UID should be good.

APT from me

ekohl:
I agree this is for now the best solution, just not happy that this is the best solution ;)

PR #114, opened 2016-12-01
Change existing Kafo type definitions to Puppet 4 types

stbenjam:


ekohl:
This fixes an issue with nightlies so I'd like to get this merged.

stbenjam:
Ready for another look.  Puppet is not needed at the top level anymore, we did it a different way.

ekohl:
Is the removal of capsule intentional? It feels like it doesn't belong in changing the kafo type definitions. Maybe at least make it a separate commit?

manifests/foreman_proxy_content.pp
```diff
@@ -2,12 +2,15 @@
...
+# $parent_fqdn::                    FQDN of the parent node. Does not usually
 #                                   need to be set.
+#                                   type:Optional[String]
 #
-# $foreman_proxy_fqdn::             fqdn of the foreman proxy. REQUIRED
+# $foreman_proxy_fqdn::             FQDN of the foreman proxy
+#                                   type:String
 #
-# $certs_tar::                      path to tar file with certs to generate. REQUIRED
+# $certs_tar::                      Path to tar file with certs to generate
```
ekohl: A bit odd that this is required but in the code there's ```if $certs_tar {```.

stbenjam: You're right, it is actually optional since you don't need the certs tar on a Katello.  I updated the type

stbenjam: I moved it here: https://github.com/Katello/puppet-certs/pull/123

stbenjam:
Ok ready for another look

PR #115, opened 2016-12-13
fixes #17658 - support restarting goferd on OS's with systemd, too

stbenjam:


templates/rhsm-katello-reconfigure.erb
```diff
@@ -95,6 +95,10 @@ fi
   service goferd status >/dev/null && \
   service goferd restart >/dev/null 2&>1
 
+[ -f /bin/systemctl ] && \
+  systemctl status goferd >/dev/null && \
```
ekohl: status && restart can be simpified to ```try-restart```

stbenjam: Neat

PR #117, opened 2016-12-19
fixes #17714 - use pki dir for puppet client certs

stbenjam:


manifests/puppet.pp
```diff
@@ -43,11 +43,11 @@
...
     } ~>
-    file { $client_key:
-      ensure => file,
-      owner  => '52',
-      mode   => '0400',
+    file { [$client_cert, $client_key, $ssl_ca_cert]:
+      ensure  => file,
+      owner   => 'puppet',
+      mode    => '0400',
+      require => Class['puppet::server'],
```
stbenjam: Using @ehelms original idea from https://github.com/Katello/puppet-certs/pull/113, we can't rely on the UID until https://tickets.puppetlabs.com/browse/SERVER-1381 is fixed.

ekohl:
Strictly speaking I think this refs #17714, not fixes. Other than that I think this is correct. It could even require just ```puppet::server::install``` I think.

stbenjam:
Updated per your suggestion

manifests/params.pp
```diff
@@ -62,24 +62,10 @@
...
-
-  $puppet_ssldir = $aio_package ? {
-    true    => '/etc/puppetlabs/puppet/ssl',
-    default => '/etc/puppet'
-  }
-
-  $puppet_client_cert = "${puppet_ssldir}/client_cert.pem"
-  $puppet_client_key  = "${puppet_ssldir}/client_key.pem"
+  $puppet_client_cert = "${pki_dir}/puppet_client.crt"
+  $puppet_client_key  = "${pki_dir}/puppet_client.key"
```
ehelms: With our other certificates, we'd been following a convention of putting public certs into `${pki_dir}/certs` and keys into `${pki_dir}/private`.

stbenjam: Updated

manifests/puppet.pp
```diff
@@ -43,11 +43,11 @@
...
     } ~>
-    file { $client_key:
-      ensure => file,
-      owner  => '52',
-      mode   => '0400',
+    file { [$client_cert, $client_key, $ssl_ca_cert]:
+      ensure  => file,
+      owner   => 'puppet',
+      mode    => '0400',
+      require => Class['puppet::server::install'],
```
jlsherrill: @stbenjam do you think this would resolve http://projects.theforeman.org/issues/17863?  Or do you think its already resolved by https://github.com/Katello/puppet-certs/commit/67d027e7e00117cc8b51fc389be116e858ec268c

stbenjam: They probably need this if they're using Puppet 4 on a clean box

stbenjam:
@ehelms Can you re-review? I made your requested changes

manifests/puppet.pp
ehelms: Could we pull this change out into its own issue then? If we think users of 3.3 are going to need this.

stbenjam: It's here https://github.com/Katello/puppet-certs/pull/121

ehelms:
I will re-review once #121 is in.

stbenjam:
#121 is in

stbenjam:
@ehelms Putting it into `${pki_dir}/private` as [suggested](https://github.com/Katello/puppet-certs/pull/117#discussion_r93551548) doesn't look like it'll work.  Puppet user needs access to that directory, but it's set root:foreman.

```
drwxr-x---.  2 root      foreman 4096 Jan 16 16:04 private
```

Any opinions where to put it instead?

ehelms:
Hrmm, Apache has the same issue but we put into private directory. We look to be adding Apache user to the foreman user_group (https://github.com/Katello/puppet-katello/blob/master/manifests/init.pp#L167). Could we do that with the puppet user, or is that overkill? 

If the above is overkill, we could add a subdirectory to keep things private:

    /etc/pki/katello/puppet
    /etc/pki/katello/puppet/certs
    /etc/pki/katello/puppet/private

ekohl:
I'd prefer to avoid managing the group so if we can use directory permissions that would be my preference.

stbenjam:
I feel like if everything is going in `/etc/pki/katello/puppet`, having another layer of directories for certs and keys is overkill - only puppet user can read the files in `/etc/pki/katello/puppet`, is there some reason to add those additional dirs? It makes the puppet code more complicated

ehelms:
Fair point, if we use the `puppet` subdirectory and lock it down to puppet user and group then having them in there is good by me. In general, the strategy had been to make certs more widely readable and only lock down keys. But this is fine based on your feedback.

ehelms:
Looks good after the test failure fix

PR #118, opened 2016-12-28
Modified codes to switch tar option

netman2k:
certs::tar_extract option always use -xzf option even though certificate file extension is tar not tar.gz.
so I changed the code to see the extension before run tar command.


ekohl:
Maybe we should just use ```xaf``` so tar can automatically detect the extension?

netman2k:
@ekohl, your idea is better. nice

netman2k:
I'll reopen pull request. thanks

PR #119, opened 2016-12-28
Changed tar option to determine the file is compressed or not

netman2k:
certs::tar_extract always use -xzf option even though the file was not compressed.

When I run the command below, the foreman-installer fails due to this problem.
```
foreman-installer --scenario capsule --capsule-certs-tar /root/katello-httpd-ssl-archive-capsule01.example.net-1.0-10.tar --certs-update-all
```

please change the option.

### DIFF
--- a/manifests/tar_extract.pp
+++ b/manifests/tar_extract.pp
@@ -15,6 +15,6 @@ define certs::tar_extract($path = $title) {
   exec { "extract ${path}":
     cwd     => '/root',
     path    => ['/usr/bin', '/bin'],
-    command => "tar -xzf ${path}",
+    command => "tar -xaf ${path}",
   }
 }


ehelms:
Would you mind explaining the underlying issue a bit? The create command attempts to compress (https://github.com/Katello/puppet-certs/blob/master/manifests/tar_create.pp#L19) but you appear to have a certificate archive without compression?

netman2k:
I did odd things recently and found that the tar_extract.pp does not recognize well.
It's not possibly happened. 

Basically, when we see the **tar** extension of a file, we do not think that it's compressed.
I'm not sure why the file name ends with tar not gz, but we can create a file like capsule-cert-generated generated. 

That's why I wanted to change this code. I think it's better to understand.

PR #120, opened 2016-12-29
refs #15931 - allow passing the cname parameter when generating certs

evgeni:
This PR does not completely fix http://projects.theforeman.org/issues/15931 aka https://bugzilla.redhat.com/show_bug.cgi?id=1160344, but it at least allows to actually generate certificates with multiple names in them which would have prevented the mentioned issue.

The `cert` resource already knows how to handle `katello-ssl-tool`s `--set-cname`, so we just have to pass the right value to it. This is done by creating a default `[]` cname parameter and passing this along to all `cert` invocations. Kafo then can override the value via the command line.

### DIFF
--- a/manifests/foreman_proxy_content.pp
+++ b/manifests/foreman_proxy_content.pp
@@ -9,13 +9,17 @@
 # $foreman_proxy_fqdn::             FQDN of the foreman proxy
 #                                   type:String
 #
+# $foreman_proxy_cname::            additional names of the foreman proxy
+#                                   type:Array
+#
 # $certs_tar::                      Path to tar file with certs to generate
 #                                   type:Optional[Stdlib::Absolutepath]
 #
 class certs::foreman_proxy_content (
-  $parent_fqdn        = $fqdn,
-  $foreman_proxy_fqdn = $certs::node_fqdn,
-  $certs_tar          = $certs::params::certs_tar
+  $parent_fqdn          = $fqdn,
+  $foreman_proxy_fqdn   = $certs::node_fqdn,
+  $foreman_proxy_cname  = $certs::cname,
+  $certs_tar            = $certs::params::certs_tar
   ) inherits certs::params {
 
   # until we support again pushing the cert rpms to the Katello,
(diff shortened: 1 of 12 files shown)


manifests/capsule.pp
```diff
@@ -7,11 +7,15 @@
 #
```
ehelms: This file is not being used anymore.

manifests/params.pp
```diff
@@ -16,6 +16,7 @@
   $ssl_build_dir = '/root/ssl-build'
 
   $node_fqdn = $::fqdn
+  $node_cname = []
```
ehelms: Could we just change this to `cname` ? The `node` noemclature was there for pulp nodes and the old node-installer and I think it helps to not use it now.

manifests/capsule.pp
evgeni: Then it should not exist? ;-)
But it was quite handy, as it allowed me to test on a plain Katello 3.2 install.

manifests/params.pp
evgeni: Sure, will do.

manifests/capsule.pp
ehelms: I agree it should not :)

manifests/params.pp
evgeni: done

ehelms:
Thanks @evgeni !

iNecas:
What's missing to fix 1160344? The BZ mentions ability to set the CNAME on certs as alternative way of addressing the issue?

evgeni:
@iNecas this only does the katello certs, so missing for 1160344 would be (IMHO):
* puppet master certs (RHBZ#1415139)
* the ability for foreman/katello to actually use the alternative names, currently all templates will use the primary name

iNecas:
My point is, I see this already bringing value to the users, and makes their lives easier when needing this. I would suggest treating the other issues as ones blocked by this, but we could start delivering this sooner, and for earlier releases, this could become a documentation bug.

ehelms:
@evgeni before I open a full Redmine issue, I hit this while testing a smart proxy:

```
foreman-proxy-certs-generate
--foreman-proxy-fqdn
bye-puppet-proxy.example.com
--certs-tar
/root/bye-puppet-proxy.example.com.tar.gz

start: 2017-01-24 20:49:08.779575

end: 2017-01-24 20:49:14.890633

delta: 0:00:06.111058

stdout: Error during configuration, exiting

stderr: stty: standard input: Inappropriate ioctl for device
Parameter foreman-proxy-cname invalid: nil is not a valid array
```

evgeni:
@ehelms possible. is that puppet 3 or puppet 4?

ehelms:
Puppet 4

evgeni:
@ehelms can you try with 3? I think my test box for that only ever had 3.

ekohl:
@ehelms maybe katello-installer needs a migration?

stbenjam:
foreman-proxy-certs-generate doesn't have migrations, the answers are embedded in the script.  Maybe certs::params isn't getting loaded somehow?

manifests/foreman_proxy_content.pp
```diff
@@ -9,27 +9,31 @@
...
 # $certs_tar::                      Path to tar file with certs to generate
 #                                   type:Optional[Stdlib::Absolutepath]
 #
 class certs::foreman_proxy_content (
-  $parent_fqdn        = $fqdn,
-  $foreman_proxy_fqdn = $certs::node_fqdn,
-  $certs_tar          = $certs::params::certs_tar
+  $parent_fqdn          = $fqdn,
+  $foreman_proxy_fqdn   = $certs::node_fqdn,
+  $foreman_proxy_cname  = $certs::cname,
```
stbenjam: Maybe `$certs` name space isn't loaded when this class is processed in foreman-proxy-certs-generate?

PR #121, opened 2017-01-09
fixes #17863 - use puppet user not uid

stbenjam:
The UID is not using the assigned one:
https://tickets.puppetlabs.com/browse/SERVER-1381, so we can't do the 52
trick.  We need to require the puppet server install class to ensure the
package is installed first.

### DIFF
--- a/manifests/puppet.pp
+++ b/manifests/puppet.pp
@@ -44,9 +44,10 @@ class certs::puppet (
       key_pair => $::certs::server_ca,
     } ~>
     file { $client_key:
-      ensure => file,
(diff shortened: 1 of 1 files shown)


manifests/puppet.pp
```diff
@@ -45,8 +45,9 @@
     } ~>
     file { $client_key:
       ensure => file,
-      owner  => '52',
+      owner  => 'puppet',
       mode   => '0400',
+      require => Class['puppet::server::install'],
```
ekohl: Alignment is off so puppet-lint fails.

PR #124, opened 2017-01-31
Only reference ::certs::params values in foreman_proxy_content

stbenjam:
Values inside `certs` aren't available when `foreman-proxy-certs-generate` is run:

```
[root@certs certs]# foreman-proxy-certs-generate --certs-tar /root/foo.tar.gz --foreman-proxy-fqdn foo.example.com
Parameter foreman-proxy-cname invalid: nil is not a valid array
```

 so we need to get the defaults from the `params`, instead.  I can't think of a better solution.

In order to generate proxy certs with cnames the command would be:

```
foreman-proxy-certs-generate --certs-tar /root/foo.tar.gz --foreman-proxy-fqdn foo.example.com --foreman-proxy-cname foo-lb.example.com`
```

### DIFF
--- a/manifests/foreman_proxy_content.pp
+++ b/manifests/foreman_proxy_content.pp
@@ -17,8 +17,8 @@
 #
 class certs::foreman_proxy_content (
   $parent_fqdn          = $fqdn,
-  $foreman_proxy_fqdn   = $certs::node_fqdn,
-  $foreman_proxy_cname  = $certs::cname,
+  $foreman_proxy_fqdn   = $certs::params::node_fqdn,
+  $foreman_proxy_cname  = $certs::params::cname,
   $certs_tar            = $certs::params::certs_tar
   ) inherits certs::params {
 


stbenjam:
@evgeni Does this affect anything with how you thought this should work? I don't think so, since the foreman_proxy_content class propagates out the cname to the other certs classes it calls.

evgeni:
@stbenjam does it in any way affect how the internal capsule is installed and configured using foreman-installer --scenario katello?
I am not sure how the proxy stuff is included there.

For standalone proxy it seems reasonable.

stbenjam:
I don't think so, it looks like certs::foreman_proxy_content isn't used anywhere but to generate certs for an external proxy.  The actual proxy setup in the puppet-foreman_proxy_content class calls the individual cert classes (certs::qpid, etc).

evgeni:
Should be fine then. This was also my thinking, but I'm on the road and don't have a setup to give it a quick go.

stbenjam:
Green

PR #126, opened 2017-03-06
Fixes #16911 - cert-req should not require

ahumbe:
foreman-installer should not require --certs-server-cert-req parameter

### DIFF
--- a/lib/puppet/provider/cert/katello_ssl_tool.rb
+++ b/lib/puppet/provider/cert/katello_ssl_tool.rb
@@ -15,7 +15,6 @@ Puppet::Type.type(:cert).provide(:katello_ssl_tool, :parent => Puppet::Provider:
       FileUtils.mkdir_p(build_path)
       FileUtils.cp(resource[:custom_pubkey], build_path(File.basename(pubkey)))
       FileUtils.cp(resource[:custom_privkey], build_path(File.basename(privkey)))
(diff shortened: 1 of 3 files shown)


ekohl:
I don't know the details of it, but removing it sounds like more than just 'not require'.

ahumbe:
The REQ file is passed by the CA server to sign it so I think foreman has no use of req file.  In case of using custom ssl certs signed by thirdpary CA it should not be a required parameter for installer to pass. Please correct me if any thing missing/wrong.

ehelms:
Since we do not currently sign the custom certificates with the default CA, this is a sound change. The bigger question should be whether we should be signing the custom certificates or not with two questions in mind:

 1) would it simplify deployment (currently we pass around two CA crts everywhere)
 2) would it couple the certificates too much

ahumbe:
@ehelms Hi, did you get chance to review this ?

ekohl:
I'm still a bit unfamiliar with the whole certs module and @ehelms is afk for a while. I do wonder why you left the parameter in for foreman_proxy.

ekohl:
There are conflicts now and I'm not sure how to proceed with this.

tbrisker:
@ekohl @ahumbe @ehelms Any chance you can try moving this issue forward?

ekohl:
Closing in favor of #172

PR #128, opened 2017-03-16
fixes #17378 - tomcat has dedicated certificate

timogoebel:
@akofink: Sorry to hijack the issue from you, but we needed an immediate fix for this. I guess you have enough work to do without this or can help testing/improving this.

### DIFF
--- a/manifests/candlepin.pp
+++ b/manifests/candlepin.pp
@@ -8,13 +8,13 @@ class certs::candlepin (
   $deploy                 = $::certs::deploy,
   $ca_cert                = $::certs::ca_cert_stripped,
   $ca_key                 = $::certs::ca_key,
-  $pki_dir                = $::certs::params::pki_dir,
-  $keystore               = $::certs::params::candlepin_keystore,
-  $keystore_password_file = $::certs::params::keystore_password_file,
-  $amqp_truststore        = $::certs::params::candlepin_amqp_truststore,
-  $amqp_keystore          = $::certs::params::candlepin_amqp_keystore,
-  $amqp_store_dir         = $::certs::params::candlepin_amqp_store_dir,
-  $tomcat                 = $::certs::params::tomcat,
+  $pki_dir                = $::certs::pki_dir,
+  $keystore               = $::certs::candlepin_keystore,
+  $keystore_password_file = $::certs::keystore_password_file,
+  $amqp_truststore        = $::certs::candlepin_amqp_truststore,
+  $amqp_keystore          = $::certs::candlepin_amqp_keystore,
+  $amqp_store_dir         = $::certs::candlepin_amqp_store_dir,
+  $tomcat                 = $::certs::tomcat,
 ) inherits certs::params {
 
   Exec {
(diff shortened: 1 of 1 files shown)


akofink:
@timogoebel no worries! Thanks for the fix! I'll let someone else who is more familiar with this module do the review.

manifests/candlepin.pp
```diff
@@ -49,6 +70,13 @@
 
   if $deploy {
 
+    Cert[$tomcat_cert_name] ~>
```
ekohl: A quick spec test tells me that just for a requires this works. The difference is that here's a notify. I'm not sure if a notify is needed, but IMHO we should remove all these relations and add autonotify to the type if it's needed.

ekohl: Thinking about it should be ```autosubscribe```.

timogoebel: I agree. Let's make a list of issues with the certs module. I have a few more items to share. :-)

ehelms: For the above, I'm fine with filing Github issues to track these unless you feel we should push to Redmine to collect them all under a single tracker.

ekohl: For now we mostly went with submitting PRs :)

ehelms:
I have a feeling this change won't propagate on an upgrade into the keystore properly. I'd like to test that out before hand.

timogoebel:
> I have a feeling this change won't propagate on an upgrade into the keystore properly.

You are right. This has actually never worked.
Instead of the `creates` on the exec resource we could use something like this with an `unless`:
```
openssl x509 -in /etc/pki/katello/certs/katello-tomcat.crt -sha1 -noout -fingerprint | grep $(keytool -list -keystore /etc/pki/katello/keystore -storepass ptjcqXxE7pvBw7jNeKgAfgKQVos5oqRv -alias tomcat | grep SHA1 | cut -d" " -f4) && echo 'yes' || echo 'no'
```

However, I'd prefer to use https://github.com/puppetlabs/puppetlabs-java_ks. This does introduce another dependency but is probably a lot more reliable than an exec resource.

@ehelms / @ekohl : What do you guys prefer?

ehelms:
I think I'd prefer to take this as is and then re-factor after the fact to account for that. My installation test with this passed, going to try an update test with this then will comment back.

ekohl:
@timogoebel I agree that dedicated types are better for this. Given there's already a puppet module for this, I think we should add the dependency and make use of it. Perhaps we can even replace more of our custom code with it.

@ehelms I'd really like to start getting the katello-* modules in the same shape as theforeman-*. That means there are useful for use without the installer and allow upgrading. Ideally master should also always work. Currently @timogoebel is doing a lot of the work to move us in that direction. I'd prefer to get the proper unless in right from the start.

ehelms:
@ekohl The only reason I was leaning towards this fix then a re-factoring was due to the fact this change is tied to a real issue that users have hit and is marked against Katello 3.4.0. That is to say, I wanted to ensure a fix for this issue for 3.4.0 that was, as much as possible, minimally invasive at a minimum (if @timogoebel whips up a fix per your request quickly then I'm happy to consider that instead considering branching approaches).

ekohl:
@ehelms that makes sense and I'd accept this fix if getting the proper fix is hard and takes a while.

ehelms:
This did pass install and a basic upgrade test for me.

timogoebel:
Excellent. I'd vote for merging this then. I'll try to write a patch to use keytool and fix this properly nevertheless.

ehelms:
@ekohl you ok with that plan?

manifests/candlepin.pp
```diff
@@ -41,6 +41,27 @@
     password_file => $::certs::ca_key_password_file,
   }
 
+  $tomcat_cert_name = "${hostname}-tomcat"
+  $tomcat_cert = "${certs::pki_dir}/certs/katello-tomcat.crt"
```
ekohl: Please use ```${::certs::pki_dir}``` or even better: fix the class parameters to always use ```$::certs::*``` instead of ```$::certs::params::*``` so you can just use ```$pki_dir```.

timogoebel: Fixed.

PR #129, opened 2017-03-17
Make variable usage and indenting consistent

ekohl:
* Use the absolute $::certs::
* Use local variables where possible
* Correct indenting

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -1,7 +1,7 @@
 # Certs Installation
 class certs::install {
 
-  package{['katello-certs-tools']:
+  package { 'katello-certs-tools':
     ensure  => installed,
   }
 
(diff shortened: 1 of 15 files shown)


ekohl:
@timogoebel I think this is a good PR to get in before we do much more refactoring.

manifests/install.pp
```diff
@@ -1,7 +1,7 @@
 # Certs Installation
 class certs::install {
 
-  package{['katello-certs-tools']:
+  package {['katello-certs-tools']:
```
timogoebel: We can safely remove the array here.

ekohl:
Added the changes in a separate commit to make it easier to review, but should be squashed when merging.

PR #130, opened 2017-03-17
Allow newer versions of dependencies

ekohl:
* puppet-extlib version 1.x has no incompatiblities compared to 0.x.
* evenup-trusted_ca was transfered to jlambert121-trusted_ca.


ekohl:
I should note that I haven't tested this yet, but reading the git logs there should be no incompatibilities.

PR #131, opened 2017-03-18
classes inherit from init

timogoebel:
As promised, this is the first PR to clean up the certs module. It includes some very basic test coverage. I want to extend it when working on the following list:

More issues we should fix:

* Client certs should not have a cname
* This module should not have any dependencies to tomcat, apache or whatever. The resource ordering should be done in the katello module.
* The classes that actually generate certificates should be defined types to allow reuse.

### DIFF
--- a/manifests/apache.pp
+++ b/manifests/apache.pp
@@ -1,18 +1,30 @@
 # Certs configurations for Apache
 class certs::apache (
-
-  $hostname        = $::certs::node_fqdn,
-  $cname           = $::certs::cname,
-  $generate        = $::certs::generate,
-  $regenerate      = $::certs::regenerate,
-  $deploy          = $::certs::deploy,
-) inherits certs::params {
+  $hostname             = $::certs::node_fqdn,
+  $cname                = $::certs::cname,
+  $generate             = $::certs::generate,
+  $regenerate           = $::certs::regenerate,
+  $deploy               = $::certs::deploy,
+  $pki_dir              = $::certs::pki_dir,
+  $server_cert          = $::certs::server_cert,
+  $server_key           = $::certs::server_key,
+  $server_cert_req      = $::certs::server_cert_req,
+  $country              = $::certs::country,
+  $state                = $::certs::state,
+  $city                 = $::certs::city,
+  $org                  = $::certs::org,
+  $org_unit             = $::certs::org_unit,
+  $expiration           = $::certs::expiration,
+  $default_ca           = $::certs::default_ca,
+  $ca_key_password_file = $::certs::ca_key_password_file,
+  $group                = $::certs::group,
+) inherits certs {
 
   $apache_cert_name = "${hostname}-apache"
-  $apache_cert = "${::certs::pki_dir}/certs/katello-apache.crt"
-  $apache_key  = "${::certs::pki_dir}/private/katello-apache.key"
+  $apache_cert = "${pki_dir}/certs/katello-apache.crt"
+  $apache_key  = "${pki_dir}/private/katello-apache.key"
 
-  if $::certs::server_cert {
+  if $server_cert {
     cert { $apache_cert_name:
       ensure         => present,
       hostname       => $hostname,
@@ -20,26 +32,26 @@ class certs::apache (
       generate       => $generate,
       deploy         => $deploy,
       regenerate     => $regenerate,
-      custom_pubkey  => $::certs::server_cert,
-      custom_privkey => $::certs::server_key,
-      custom_req     => $::certs::server_cert_req,
+      custom_pubkey  => $server_cert,
+      custom_privkey => $server_key,
+      custom_req     => $server_cert_req,
     }
   } else {
     cert { $apache_cert_name:
       ensure        => present,
       hostname      => $hostname,
       cname         => $cname,
-      country       => $::certs::country,
-      state         => $::certs::state,
-      city          => $::certs::city,
-      org           => $::certs::org,
-      org_unit      => $::certs::org_unit,
-      expiration    => $::certs::expiration,
-      ca            => $::certs::default_ca,
+      country       => $country,
+      state         => $state,
+      city          => $city,
+      org           => $org,
+      org_unit      => $org_unit,
+      expiration    => $expiration,
+      ca            => $default_ca,
       generate      => $generate,
       regenerate    => $regenerate,
       deploy        => $deploy,
-      password_file => $::certs::ca_key_password_file,
+      password_file => $ca_key_password_file,
     }
   }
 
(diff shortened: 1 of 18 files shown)


manifests/candlepin.pp
```diff
@@ -80,19 +80,14 @@
...
-      notify      => Service['qpidd'],
+      #notify      => Service['qpidd'],
     } ~>
     file { $amqp_store_dir:
       ensure => directory,
       owner  => 'tomcat',
       group  => $::certs::group,
       mode   => '0750',
     } ~>
-    exec { 'create candlepin qpid exchange':
```
timogoebel: Code has moved here:

https://github.com/Katello/puppet-katello/pull/170

manifests/candlepin.pp
```diff
@@ -80,19 +80,14 @@
       client_cert => $client_cert,
       refreshonly => true,
       subscribe   => Exec['create-nss-db'],
-      notify      => Service['qpidd'],
+      #notify      => Service['qpidd'],
```
timogoebel: I don't think, this is necessary. It's a false dependency that causes issues in a distributed setup.

manifests/ca.pp
```diff
@@ -0,0 +1,119 @@
...
+  $group                   = $::certs::group,
+  $katello_server_ca_cert  = $::certs::katello_server_ca_cert,
+  $katello_default_ca_cert = $::certs::katello_default_ca_cert,
+  $nss_db_dir              = $::certs::nss_db_dir,
+  $ca_key                  = $::certs::ca_key,
+  $ca_cert                 = $::certs::ca_cert,
+  $ca_cert_stripped        = $::certs::ca_cert_stripped,
+  $ca_key_password         = $::certs::ca_key_password,
+  $ca_key_password_file    = $::certs::ca_key_password_file,
+) inherits certs {
```
timogoebel: This might not actually work correctly as `certs` includes `certs::ca` (read: this class). Any insights?

manifests/candlepin.pp
```diff
@@ -101,7 +96,7 @@
       # Stupid keytool doesn't allow you to import a keypair.  You can only import a cert.  Hence, we have to
       # create the store as an PKCS12 and convert to JKS.  See http://stackoverflow.com/a/8224863
       command => "openssl pkcs12 -export -name amqp-client -in ${client_cert} -inkey ${client_key} -out /tmp/keystore.p12 -passout file:${password_file} && keytool -importkeystore -destkeystore ${amqp_keystore} -srckeystore /tmp/keystore.p12 -srcstoretype pkcs12 -alias amqp-client -storepass ${keystore_password} -srcstorepass ${keystore_password} -noprompt && rm /tmp/keystore.p12",
-      unless  => "keytool -list -keystore ${amqp_keystore} -storepass ${keystore_password} -alias ${certs::default_ca_name}",
+      unless  => "keytool -list -keystore ${amqp_keystore} -storepass ${keystore_password} -alias amqp-client",
```
timogoebel: This is fixing a but that didn't make this idempotent.

manifests/foreman_proxy_content.pp
```diff
@@ -20,7 +20,7 @@
   $foreman_proxy_fqdn   = $certs::params::node_fqdn,
   $foreman_proxy_cname  = $certs::params::cname,
   $certs_tar            = $certs::params::certs_tar
-  ) inherits certs::params {
+) inherits certs {
```
ekohl: This class is meant to be used from a different context, without the certs class I believe.

manifests/foreman_proxy.pp
```diff
@@ -100,14 +100,13 @@
...
     pubkey { $foreman_ssl_ca_cert:
-      key_pair => $::certs::server_ca,
+      key_pair => $::certs::default_ca,
     } ~>
     key_bundle { $foreman_proxy_ssl_client_bundle:
       key_pair => Cert[$foreman_proxy_client_cert_name],
     } ~>
     file { $foreman_proxy_ssl_client_bundle:
       ensure => file,
-      owner  => $::certs::group,
```
ekohl: Any particular reason to remove this ownership?

timogoebel: This is a bug as well. The file is world readable, setting the group to `qpid` doesn't make any sense for a `foreman_proxy` related file.

manifests/ca.pp
ekohl: This file could be a "private" standalone class that is not designed to be called from other modules. Like ```certs:install``` and ```certs::config```

manifests/foreman_proxy.pp
timogoebel: yes. see the commend above.

timogoebel:
@ekohl : Created separate PR for every change in this.

Test failures are expected. They need some of the other PRs merged first.

timogoebel:
Rebased. This should be ready to merge (as in not having any dependencies), now.

manifests/puppet.pp
```diff
@@ -9,7 +9,7 @@
   $client_cert = $::certs::params::puppet_client_cert,
```
ekohl: These could be ```$::certs::puppet_client_cert```, even when it wasn't inheriting from the certs class.

timogoebel:
> Would a require ::certs as the first line in each class have the same effect in forcing the puppet parser to get the correct ordering?

@ekohl: Unfortunately, that isn't the same. Inheriting makes the puppet parser parse `init.pp` before   the parameters of the class are parsed. That's what we want. If you use `include ::certs` as first line in the class, that is parsed after the parameters are parsed.

timogoebel:
Rebased.

timogoebel:
@ekohl : Rebased and changed all variables to use local scope. Test failures are weird, don't see them locally. Any idea, what's wrong?

manifests/apache.pp
```diff
@@ -1,16 +1,28 @@
...
+  $group                = $::certs::group,
+) inherits certs {
 
   $apache_cert_name = "${hostname}-apache"
-  $apache_cert = "${::certs::pki_dir}/certs/katello-apache.crt"
-  $apache_key  = "${::certs::pki_dir}/private/katello-apache.key"
+  $apache_cert = "${pki_dir}/certs/katello-apache.crt"
+  $apache_key  = "${pki_dir}/private/katello-apache.key"
 
   if $::certs::server_cert {
```
ekohl: This can be local as well.

ekohl:
@timogoebel they're from puppet-lint 2.2. Most likely you have an older version installed. We've been discussing about disabling that check but haven't reached a conclusion.

ekohl:
Needs a rebase. I'm now convinced this is the only solution to the problem and I'd like to get this in.

timogoebel:
@ekohl : Thanks. Rebased.

ekohl:
There is currently an issue where pulp isn't properly installed but that's also present without this patch. I'd like to get that resolved before we merge this.

ekohl:
So we've finally fixed nightlies. Tomorrow I'll test this with fresh nightlies and hopefully we can finally get this merged. If you could rebase then Travis should pass since we dropped Puppet 3 from the matrix.

timogoebel:
Rebased.

ekohl:
Thanks!

PR #133, opened 2017-03-18
fixtures.yml uses https

timogoebel:
... required when checking out the code behind a proxy.


timogoebel:
Needs a rebase once #130 is merged.

PR #136, opened 2017-03-18
foreman_proxy does not need foreman user

timogoebel:
The file is world readable, setting the owner to user
`foreman` causes errors because a user named `foreman` does not exist
on a system with just foreman proxy installed.
Removing the owner causes the file to be owned by root which is fine as
it is world readable.


ehelms:
Thanks @timogoebel !

PR #137, opened 2017-03-18
Introduce certs::keypair

ekohl:
This is an untested experiment to see if the code ends up cleaner.

### DIFF
--- a/manifests/keypair.pp
+++ b/manifests/keypair.pp
@@ -1,0 +1,41 @@
+define certs::keypair (
+  $key_pair,
+  $key_file,
+  $cert_file,
+  $manage_key  = false,
+  $key_owner   = undef,
+  $key_group   = undef,
+  $key_mode    = undef,
+  $manage_cert = false,
+  $cert_owner  = undef,
+  $cert_group  = undef,
+  $cert_mode   = undef,
+) {
+  Cert[$key_pair] ~>
+  privkey { $key_file:
+    key_pair => Cert[$key_pair],
+  } ~>
+  pubkey { $cert_file:
+    key_pair => Cert[$key_pair],
+  }
+
+  if $manage_key {
+    file { $key_file:
+      ensure  => file,
+      owner   => $key_owner,
+      group   => $key_group,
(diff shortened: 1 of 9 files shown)


manifests/keypair.pp
```diff
@@ -0,0 +1,41 @@
...
+  $manage_key  = false,
+  $key_owner   = undef,
+  $key_group   = undef,
+  $key_mode    = undef,
+  $manage_cert = false,
+  $cert_owner  = undef,
+  $cert_group  = undef,
+  $cert_mode   = undef,
+) {
+  Cert[$key_pair] ~>
```
ekohl: I do believe this is redundant but didn't want to drop too much at the same time. This at least puts it in one place and makes it easier to change it.

manifests/keypair.pp
```diff
@@ -0,0 +1,41 @@
+define certs::keypair (
+  $key_pair,
+  $key_file,
+  $cert_file,
+  $manage_key  = false,
+  $key_owner   = undef,
+  $key_group   = undef,
+  $key_mode    = undef,
+  $manage_cert = false,
```
ehelms: Do we want to manage by default? Seems like we should default to manage the certificates and keys since we generate them and let users of the module opt out where necessary. In some cases, we weren't always "managing" the cert file but I think this was largely due to inexperience with puppet and only "managing" it when we needed to set user/mode specifically. Whereas, we probably ought to have been managing it the whole time to ensure puppet recreates the files when the original file changes or we run cert updates.

ekohl: I'm unsure because I didn't know the larger impact of it. I also considered not adding the parameter at all and manage the file if any property is not undef.

Maybe we should do it in stages and start with this refactoring and go from there. I was talking to @timogoebel and we both think we can pull many service integrations from this into other modules. Let's see it on a case-by-case basis.

ekohl:
Rebased to resolve the merge conflict.

ehelms:
@ekohl already another rebase needed

ekohl:
Rebased

PR #139, opened 2017-03-18
extract nssdb creation into separate class

timogoebel:
Both `qpidd` and `candlepin` need the `nssdb` present. This commit extracts the code to a common class.

### DIFF
--- a/manifests/ssltools/nssdb.pp
+++ b/manifests/ssltools/nssdb.pp
@@ -1,0 +1,42 @@
+# Sets up nssdb
+class certs::ssltools::nssdb (
+  $nss_db_dir = $::certs::nss_db_dir,
+  $group = $::certs::qpidd_group,
+)  {
+  Exec { logoutput => 'on_failure' }
+
+  $nss_db_password_file   = "${nss_db_dir}/nss_db_password-file"
+  $nssdb_files            = ["${nss_db_dir}/cert8.db", "${nss_db_dir}/key3.db", "${nss_db_dir}/secmod.db"]
+
+  file { $nss_db_dir:
+    ensure => directory,
+    owner  => 'root',
+    group  => $group,
+    mode   => '0755',
+  } ->
+  exec { 'generate-nss-password':
+    command => "openssl rand -base64 24 > ${nss_db_password_file}",
+    path    => '/usr/bin',
+    umask   => '0027',
+    group   => $group,
+    creates => $nss_db_password_file,
+  } ->
+  file { $nss_db_password_file:
+    ensure => file,
+    owner  => 'root',
+    group  => $group,
+    mode   => '0640',
+  } ->
+  exec { 'create-nss-db':
+    command => "certutil -N -d '${nss_db_dir}' -f '${nss_db_password_file}'",
+    path    => '/usr/bin',
+    umask   => '0027',
+    group   => $group,
+    creates => $nssdb_files,
+  } ->
+  file { $nssdb_files:
+    owner => 'root',
+    group => $group,
+    mode  => '0640',
+  }
+}
--- a/manifests/candlepin.pp
+++ b/manifests/candlepin.pp
@@ -48,6 +48,7 @@ class certs::candlepin (
   $client_key = "${::certs::pki_dir}/private/${java_client_cert_name}.key"
 
   if $deploy {
+    include ::certs::ssltools::nssdb
 
     file { $password_file:
       ensure  => file,
--- a/manifests/qpid.pp
+++ b/manifests/qpid.pp
@@ -29,12 +29,12 @@ class certs::qpid (
   }
 
   if $deploy {
+    include ::certs::ssltools::nssdb
+    $nss_db_password_file = $::certs::ssltools::nssdb::nss_db_password_file
 
-    $nss_db_password_file   = "${::certs::nss_db_dir}/nss_db_password-file"
     $client_cert            = "${::certs::pki_dir}/certs/${qpid_cert_name}.crt"
     $client_key             = "${::certs::pki_dir}/private/${qpid_cert_name}.key"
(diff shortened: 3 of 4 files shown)


manifests/ssltools/create_nssdb.pp
```diff
@@ -0,0 +1,35 @@
...
+
+  $nss_db_password_file   = "${::certs::nss_db_dir}/nss_db_password-file"
+  $nssdb_files            = ["${::certs::nss_db_dir}/cert8.db", "${::certs::nss_db_dir}/key3.db", "${::certs::nss_db_dir}/secmod.db"]
+
+  file { $::certs::nss_db_dir:
+    ensure => directory,
+    owner  => 'root',
+    group  => $::certs::qpidd_group,
+    mode   => '0755',
+  } ~>
```
ekohl: I think this doesn't have to be a notify

manifests/ssltools/create_nssdb.pp
```diff
@@ -0,0 +1,35 @@
...
+    command => "openssl rand -base64 24 > ${nss_db_password_file}",
+    path    => '/usr/bin',
+    creates => $nss_db_password_file,
+  } ->
+  file { $nss_db_password_file:
+    ensure => file,
+    owner  => 'root',
+    group  => $::certs::qpidd_group,
+    mode   => '0640',
+  } ~>
```
ekohl: I don't think this has to be a notify

manifests/ssltools/create_nssdb.pp
```diff
@@ -0,0 +1,35 @@
...
+  exec { 'create-nss-db':
+    command => "certutil -N -d '${::certs::nss_db_dir}' -f '${nss_db_password_file}'",
+    path    => '/usr/bin',
+    creates => $nssdb_files,
+  } ~>
```
ekohl: I don't think this has to be a notify

manifests/ssltools/create_nssdb.pp
```diff
@@ -0,0 +1,35 @@
+# Creates a nssdb
+class certs::ssltools::create_nssdb inherits certs {
```
ekohl: Maybe just name it ```nssdb``? The current name sounds like a command.

ekohl: I'd also prefer if this was just a class with a parameter rather than inheriting from certs. Looks like there are just 2 (nss_db_dir and group).

timogoebel:
@ekohl : Rebased and addressed your comments.

manifests/ssltools/nssdb.pp
```diff
@@ -0,0 +1,35 @@
...
+  $nss_db_password_file   = "${::certs::nss_db_dir}/nss_db_password-file"
+  $nssdb_files            = ["${::certs::nss_db_dir}/cert8.db", "${::certs::nss_db_dir}/key3.db", "${::certs::nss_db_dir}/secmod.db"]
+
+  file { $::certs::nss_db_dir:
+    ensure => directory,
+    owner  => 'root',
+    group  => $::certs::qpidd_group,
+    mode   => '0755',
+  } ->
+  exec { 'generate-nss-password':
```
ekohl: While we're at it, maybe set the umask as well for security reasons?

manifests/ssltools/nssdb.pp
```diff
@@ -0,0 +1,35 @@
...
+    path    => '/usr/bin',
+    creates => $nss_db_password_file,
+  } ->
+  file { $nss_db_password_file:
+    ensure => file,
+    owner  => 'root',
+    group  => $::certs::qpidd_group,
+    mode   => '0640',
+  } ->
+  exec { 'create-nss-db':
```
ekohl: This could use a umask as well.

manifests/ssltools/nssdb.pp
```diff
@@ -0,0 +1,35 @@
+# Creates a nssdb
+class certs::ssltools::nssdb inherits certs {
```
ekohl: Could you implement it as a non-inheriting class with parameters? That would make testing easier and faster.

timogoebel: @ekohl: How do you suggest that we pass the parameters? That would mean we can't call the class with `include` but have to use `class {'':}` although we have to call the class several times?

manifests/ssltools/nssdb.pp
timogoebel: ... to 027?

manifests/ssltools/nssdb.pp
ekohl: If they default to ```$::certs::pki_dir``` and ```$::certs__qpidd_group``` a simple include should work I think.

manifests/ssltools/nssdb.pp
ekohl: Yes, and maybe if we add ```group => $group``` the file resource won't change anything at all (but still should be there to ensure it's correct).

timogoebel:
@ekohl : The class does not inherit from certs anymore. Added the `group` and `umask` parameters. They work nicely.

manifests/ssltools/nssdb.pp
```diff
@@ -0,0 +1,42 @@
+# Sets up nssdb
+class certs::ssltools::nssdb (
+  $nss_db_dir = $::certs::nss_db_dir,
+  $qpidd_group = $::certs::qpidd_group
```
ekohl: Given this is now just about creating the nss db I think it could even be just ```$group = ...``` and use a trailing comma.

timogoebel:
Updated.

timogoebel:
Now that the pipeline is back to green, I think this should be next to merge to unblock the other PRs.

ehelms:
Thanks @timogoebel

PR #140, opened 2017-03-18
qpid does not need apache

timogoebel:
`qpid` has a dependency on `apache` although the former works perfectly fine without the latter.


manifests/qpid.pp
```diff
@@ -37,7 +37,7 @@
     $pfx_path               = "${certs::pki_dir}/${qpid_cert_name}.pfx"
     $nssdb_files            = ["${::certs::nss_db_dir}/cert8.db", "${::certs::nss_db_dir}/key3.db", "${::certs::nss_db_dir}/secmod.db"]
 
-    Package['httpd'] -> Package['qpid-cpp-server'] ->
+    Package['qpid-cpp-server'] ->
```
ehelms: This was most likely some Pulp related dependency an ordering that was perceived to be needed at the time.

PR #141, opened 2017-03-18
move qpidd reload to katello module

timogoebel:
This should go into the katello module. I haven't written the code for the module, yet. Currently, this is just a reminder.

### DIFF
--- a/manifests/candlepin.pp
+++ b/manifests/candlepin.pp
@@ -80,7 +80,6 @@ class certs::candlepin (
       client_cert => $client_cert,
       refreshonly => true,
       subscribe   => Exec['create-nss-db'],
-      notify      => Service['qpidd'],
     } ~>
     file { $amqp_store_dir:
       ensure => directory,


ekohl:
There are many cases like this where we need to move these notifies to other modules.

timogoebel:
@ekohl : https://github.com/Katello/puppet-katello/pull/172 should move this to puppet-katello.

manifests/candlepin.pp
```diff
@@ -80,7 +80,6 @@
       client_cert => $client_cert,
       refreshonly => true,
       subscribe   => Exec['create-nss-db'],
-      notify      => Service['qpidd'],
```
ehelms: I am a little surprised this was needed at all. I haven't tested to verify, but it seems like qpidd shouldn't have to be restarted every time the nssdb is updated.

ehelms:
@ekohl The katello PR was merged, what else do you think this needs?

ekohl:
There is another qpid refresh a bit lower (line 93) that you will run into as well.

ekohl:
@timogoebel could you look at the other refresh as well? After that I think this can be merged.

timogoebel:
@ekohl : I was hoping to get rid of the other qpid refresh by https://github.com/Katello/puppet-certs/pull/134
I think, this one has to go first as the exec actually needs the service running.

PR #145, opened 2017-03-23
Add an acceptance test for candlepin

ekohl:
This is how an acceptance test for candlepin could look like. Currently it doesn't pass yet and we should move more code into common areas to manage the repositories.

This includes https://github.com/Katello/puppet-certs/pull/144, builds on https://github.com/Katello/foreman-installer-modulesync/pull/20 and relates to https://github.com/Katello/puppet-pulp/issues/164.

### DIFF
--- a/manifests/ssltools/nssdb.pp
+++ b/manifests/ssltools/nssdb.pp
@@ -8,6 +8,8 @@ class certs::ssltools::nssdb (
   $nss_db_password_file   = "${nss_db_dir}/nss_db_password-file"
   $nssdb_files            = ["${nss_db_dir}/cert8.db", "${nss_db_dir}/key3.db", "${nss_db_dir}/secmod.db"]
 
+  ensure_packages(['openssl', 'nss-tools'])
+
   file { $nss_db_dir:
     ensure => directory,
     owner  => 'root',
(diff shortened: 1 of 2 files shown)


ekohl:
The acceptance tests now pass for me.

While trying to get a minimal set I also wanted to install candlepin without qpid. Since puppet-candlepin does support that I introduced a parameter to certs::candlepin to mirror this. I can submit this as a separate PR.

ekohl:
Now with an actual API call to ```/candlepin/status```. I only read the API documentation and assumed this is the correct way to check if it's up and correctly initialized. It would be great if someone could confirm that.

sean797:
`/candlepin/status` looks right to me, though I don't really know. This requires a rebase now..

Thanks for doing this BTW @ekohl 🥇

ekohl:
I think I'll first try and finish the acceptance test for candlepin itself and then revisit this. Maybe the more modular puppet-katello would be a better place to test with certificates.

ekohl:
Marking it as WIP for now.

ekohl:
Rebased it and now it verifies just that the files exist with the correct content. This will be the basis for verification on replacing the execs with puppetlabs-java_ks.

PR #150, opened 2017-03-29
Contain classes in init to preserve relationships

ehelms:


timogoebel:
@ehelms : This leads do duplicate declarations. You need to change the lines below as well to reference the class, e.g. `Class['::certs::bla']`

ehelms:
Thanks! I ran a test and this fixed the issue I described with install.

PR #154, opened 2017-04-12
fixes #19259 - apache key should be mode 0440

stbenjam:
Foreman launches websockify process that needs read access to
this key.


stbenjam:
Hm tests are failing because concat dropped puppet 3.  I guess now that 3.4 RC is out, we should branch the puppet modules and start dropping p4 in Katello too.

ekohl:
You're right and I did start with that. There are a few things that I'd like to include and reduce the number of modulesyncs we have to do.

ekohl:
If you rebase this then the tests should pass.

PR #155, opened 2017-04-13
fixes #19271 - do not restart docker service

stbenjam:
Restarting things that aren't ours is *very* bad behavior.

This means that any Katello client at all will have docker restarted - when it possibly runs containers, taking them down.  The only reason for the restart is so docker sees the updated CA trust.  Not everyone wants to manage docker with Katello either, they could just want us for software updates, and here we are taking down all of their containers on them :-(


ekohl:
If there is no docker reload then 👍 on not doing it at all.

stbenjam:
Yea, there's no reload. 

@parthaa do you know if this is going to cause problems with Atomic hosts?

ekohl:
I merge the reload PR instead.

PR #156, opened 2017-05-02
Fixes #19271 - reload docker instead of restart

ahumbe:
Restarting docker service will stop all the running containers, so instead of restart we should reload the service which will not affect any running container.

### DIFF
--- a/templates/rhsm-katello-reconfigure.erb
+++ b/templates/rhsm-katello-reconfigure.erb
@@ -80,13 +80,13 @@ if [ -d $CA_TRUST_ANCHORS ]; then
   cp $KATELLO_CERT_DIR/$KATELLO_SERVER_CA_CERT $CA_TRUST_ANCHORS
   update-ca-trust
 
-  # restart docker if it is installed and running
+  # reload docker if it is installed and running
   if [ -f /usr/lib/systemd/system/docker.service ]; then
     systemctl status docker >/dev/null && \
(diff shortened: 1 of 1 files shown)


ahumbe:
As a part of installation of katello-ca-consumer package we do not make any changes to the existing docker configuration, can any one help me understand why need to restart/reload the docker service?

ekohl:
Interesting. In https://github.com/Katello/puppet-certs/pull/155 @stbenjam said there was no reload.

ahumbe:
The reload option is there for the docker service, here is the output : 

# service docker reload
Redirecting to /bin/systemctl reload  docker.service

# tail -f /var/log/messages
May  2 10:39:53 rhel7u2-2 dockerd-current: time="2017-05-02T10:39:53.184676367-04:00" level=info msg="Got signal to reload configuration, reloading from: /etc/docker/daemon.json"
May  2 10:39:53 rhel7u2-2 dockerd-current: time="2017-05-02T10:39:53.185286790-04:00" level=error msg="open /etc/docker/daemon.json: no such file or directory"
May  2 10:39:53 rhel7u2-2 systemd: Reloaded Docker Application Container Engine.

ekohl:
I assume @stbenjam is busy with Summit right now but I'd like his opinion on this.

stbenjam:
I get an error on my box when I try reload but I guess it was added in 2015 and seems there on a fresh install on CentOS 7.  So this should be OK

PR #158, opened 2017-06-01
Fixes #19734 - enforce proper exection order for Candlepin

evgeni:
otherwise `Exec[import client certificate into Candlepin keystore]` has no relationship with `File[/etc/pki/katello/keystore_password-file]` and Puppet might execute them in any order, but the Exec needs the File.

### DIFF
--- a/manifests/candlepin.pp
+++ b/manifests/candlepin.pp
@@ -105,8 +105,7 @@ class certs::candlepin (
       target => $keystore,
       owner  => 'tomcat',
       group  => $group,
-    }
-
+    } ->
     certs::keypair { 'candlepin':
       key_pair  => $java_client_cert_name,
       key_file  => $client_key,


manifests/candlepin.pp
```diff
@@ -105,8 +105,7 @@
       target => $keystore,
       owner  => 'tomcat',
       group  => $group,
-    }
-
+    } ->
```
ehelms: Does this require the entire stack above it to execute or do they just need the password file to exist per the usage in the `import CA` step?

evgeni: Technically it just needs the file.

ehelms: So might be cleaner to put at requires on the directives below that need the password file.

evgeni: One the one hand yes, on the other the current solution does not hurt and given the current structure of the file, one can be easily tricked into thinking that everything is executed in order (which was not true before my commit).

PR #160, opened 2017-06-16
Fixes #20021 - Optionally generate CA public cert with other CAs certs

sean797:
Requires https://github.com/Katello/katello-certs-tools/pull/1

### DIFF
--- a/lib/puppet/type/ca.rb
+++ b/lib/puppet/type/ca.rb
@@ -5,4 +5,20 @@ Puppet::Type.newtype(:ca) do
 
   instance_eval(&Certs::CERT_COMMON_PARAMS)
 
+  newparam(:other_certs, :array_matching => :all) do
+    validate do |value|
+      unless value.is_a?(Array) || value.is_a?(String) || value.is_a?(Symbol)
+        raise Puppet::Error, "Puppet::Type::Ca: other_certs parameter must be an array or a string,
+          got #{value.class.name}."
+      end
+    end
+
+    munge do |value|
+      return [value] if value.is_a?(String) || value.is_a?(Symbol)
+      value
+    end
+
+    defaultto Array.new
+  end
+
 end
--- a/spec/acceptance/certs_ca_spec.rb
+++ b/spec/acceptance/certs_ca_spec.rb
@@ -1,0 +1,141 @@
+require 'spec_helper_acceptance'
+
+describe 'certs with default params' do
+  context 'with default params' do
+    let(:pp) do
+      <<-EOS
+      $repo = 'latest'
+      $dist = 'el7'
+
+      package { 'epel-release':
+        ensure => installed,
+      }
+
+      yumrepo { 'katello':
+        descr    => 'Katello latest',
+        baseurl  => "https://fedorapeople.org/groups/katello/releases/yum/${repo}/katello/${dist}/\\$basearch/",
+        gpgkey   => 'https://raw.githubusercontent.com/Katello/katello-packaging/master/repos/RPM-GPG-KEY-katello-2015',
+        gpgcheck => '0',
+        enabled  => '1',
+      }
+      class { '::certs':}
+      EOS
+    end
+
+    it_behaves_like 'a idempotent resource'
+
+    describe package('katello-certs-tools') do
+      it { is_expected.to be_installed }
+    end
+
+    describe x509_certificate('/etc/pki/katello-certs-tools/certs/katello-default-ca.crt') do
+      it { should be_certificate }
+      it { should be_valid }
+      it { should have_purpose 'SSL server CA' }
+      its(:issuer) { should eq "/C=US/ST=North Carolina/L=Raleigh/O=Katello/OU=SomeOrgUnit/CN=#{fact('fqdn')}" }
+      its(:subject) { should eq "/C=US/ST=North Carolina/L=Raleigh/O=Katello/OU=SomeOrgUnit/CN=#{fact('fqdn')}" }
+    end
+
+    describe x509_certificate('/etc/pki/katello-certs-tools/certs/katello-server-ca.crt') do
+      it { should be_certificate }
+      it { should be_valid }
+      it { should have_purpose 'SSL server CA' }
+      its(:issuer) { should eq "/C=US/ST=North Carolina/L=Raleigh/O=Katello/OU=SomeOrgUnit/CN=#{fact('fqdn')}" }
+      its(:subject) { should eq "/C=US/ST=North Carolina/L=Raleigh/O=Katello/OU=SomeOrgUnit/CN=#{fact('fqdn')}" }
+    end
+  end
+
+  context 'with custom server certs' do
+    before(:all) do
+      beforepp = <<-EOS
+        # Mark server-ca for update
+        file { '/root/ssl-build/katello-server-ca.update':
+          ensure => file,
+        }
+
+        Exec {
+          path => ['/usr/bin'],
+        }
+
+        file { '/root/custom_cert':
+          ensure => directory,
+        } ~>
+        exec { 'Create CA key':
+          cwd     => "/root/custom_cert",
+          command => "openssl genrsa -out ca.key 2048",
+          creates => "/root/custom_cert/ca.key",
+        } ~>
+        exec { 'Create CA certficates':
+          cwd     => "/root/custom_cert",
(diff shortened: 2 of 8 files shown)


Gemfile
```diff
@@ -27,7 +27,7 @@ gem 'puppet-blacksmith', '>= 3.1.0', {"groups"=>["development"]}
 gem 'json', '~> 1.0', {"platforms"=>["ruby_19"], "groups"=>["test"]}
 gem 'json_pure', '~> 1.0', {"platforms"=>["ruby_19"], "groups"=>["test"]}
 gem 'beaker-rspec', {"groups"=>["system_tests"]}
-gem 'beaker-module_install_helper', {"groups"=>["system_tests"]}
+gem 'beaker-module_install_helper', :git => 'https://github.com/puppetlabs/beaker-module_install_helper.git', :ref => 'f6356', :groups => ["system_tests"]
```
sean797: Tests fail without https://github.com/puppetlabs/beaker-module_install_helper/commit/f635654950506aad81522d45e8de9fd384a36124

spec/acceptance/certs_ca_spec.rb
```diff
@@ -0,0 +1,157 @@
...
+    let(:pp) do
+      <<-EOS
+      $repo = 'latest'
+      $dist = 'el7'
+
+      package { 'epel-release':
+        ensure => installed,
+      }
+
+      yumrepo { 'candlepin':
```
ekohl: Does it actually need the candlepin repo?

Gemfile
ekohl: Looks like that made it into a release now.

.travis.yml
```diff
@@ -9,5 +9,13 @@ env:
   - PUPPET_VERSION=4.6 ONLY_OS=redhat-7-x86_64,centos-7-x86_64
 matrix:
   fast_finish: true
+  include:
```
ekohl: This also needs a change in ```.sync.yml``` to prevent modulesync from overwriting it. See https://github.com/Katello/puppet-pulp/blob/1ad10c1e567f165aeaef847ffde23619a2648e34/.sync.yml#L2-L4

sean797:
@ekohl thanks for your review! Updated 👍 

Yes, that's it. This CA cert is also bundled in the `smart-proxy-certs.tar` when generated. Making sure that the smart proxies install CA certs from all Katello nodes and trust them, currently it'll just trust the one the `smart-proxy-certs.tar` came from.

lib/puppet/type/ca.rb
```diff
@@ -5,4 +5,20 @@
 
   instance_eval(&Certs::CERT_COMMON_PARAMS)
 
+  newparam(:other_certs) do
```
ekohl: ```:array_matching => :all``` might be a good one to add https://docs.puppet.com/puppet/5.0/custom_types.html#customizing-behaviour

lib/puppet/type/ca.rb
```diff
@@ -5,4 +5,20 @@
...
+    validate do |value|
+      unless value.is_a?(Array) || value.is_a?(String) || value.is_a?(Symbol)
```
ekohl: Does this mean you accept an array of arrays?

spec/acceptance/certs_ca_spec.rb
```diff
@@ -0,0 +1,142 @@
...
+        file { '/root/ssl-build/katello-server-ca.update':
+          ensure => file,
+        }
+
+        file { '/root/custom_cert':
+          ensure => directory,
+        } ~>
+        exec { 'Create CA key':
+          cwd     => "/root/custom_cert",
+          path    => ["/usr/bin"],
```
ekohl: Might be a bit shorter to set a default beforehand:
```puppet
Exec {
  path => ['/usr/bin'],
}

lib/puppet/type/ca.rb
sean797: No https://docs.puppet.com/puppet/5.0/custom_types.html#validation-and-munging

ekohl:
Like we discussed on IRC: I'm afraid this pattern won't scale well since every "node" will be a CA. I do like your acceptance tests and I will use it to write some more

sean797:
So we discussed a little more on IRC, putting it here for a wider audience and allow us to think about it a little more:

I don't think we should require an external CA service like freeIPA or others (although that may be a nice optional feature) I think an option may be:

Candlepin node(s) are the default CA(s) and they trust each other if there is more than one.
Other component nodes (qpidd(s), Katello(s), Smart Proxies) would all be setup using a certs.tar or similar. Server CA certs would be optionally supplied when generating Katello(s) or Smart Proxies certs.tar from a candlepin node, just how `foreman-proxy-content-certs-generate` works today.

ekohl:
I could be convinced of merging this even if we're not going to use it as the default multi-server setups but I'd like the opinion of other maintainers because how we deploy certs is still somewhat unfamiliar to me.

ehelms:
Help remind me -- we need to append because we can't rely on defining a directory of CA certs in all cases to point to vs. a single CA file?

I think it would help me to wrap my brain around the scenarios where multiple CAs would come into play. Feel free to walk me through it like a freshman in college. HA concepts are a weakness of mine that I want to understand better :)

sean797:
> Help remind me -- we need to append because we can't rely on defining a directory of CA certs in all cases to point to vs. a single CA file?

Sorry not sure what you mean here? My goal is to have the clients & Foreman Proxies trust either of the CAs on my Candlepin nodes (right now with our supported monolithic install  that also includes app, pulp, qpid, ect..)  

> I think it would help me to wrap my brain around the scenarios where multiple CAs would come into play. Feel free to walk me through it like a freshman in college. HA concepts are a weakness of mine that I want to understand better :)

Sure, so HA means Highly Available, if a system is acting in an active/passive mode then that is normally considered HA, as one node will take over when the other fails, normally with a outage as the traffic fails over, if fail over is manual then the outage will likely be longer (but this is still considered to be HA)

I'm trying to make it possible to load balance our application(s) active/active not active/passsive. This will allow for there to be no _fail over_ process which will allow users to add extra capacity to the system by adding more nodes, as well as make it natively HA (HA without pacemaker).

At the the heart of things its super simple goal. **Be able to add active component nodes independently without sharing anything to increase capacity and resilience.** The nodes shouldn't share any resources, if they require shared storage you just move the Single Point of Failure (SPOF) to the storage system.

-----

This PR allows me to have 2 CAs that each candlepin node uses. I have tested it with 2 monolithic Katello nodes and a single Smart Proxy & Client.

**How does this relate to certificates?**
Certificates need to be stored somewhere, today we probably could put the certs of shared storage and scale nodes but that just moves to SPOF somewhere else, doesn't really solve the problem.


**Any other approaches?**
Yes! I think we could use certmonger and/or freeIPA or a some other similar service but that also moves the SPOF somewhere else, though freeIPA can be made active/active (I think?) it adds yet more complexity to HA Katello and I don't know how that would work when candlepin issues certs for clients.

ehelms:
First off, thanks for the explanation! That helped a ton in wrapping my brain around how to think about this approach.

The last gap of knowledge to help me fill in is "2 CAs that each Candlepin node uses". Are you saying in this case two root CAs, essentially what is katello-default-ca, are generated and given to each Candlepin instance? And that Candlepin is able to take both CAs and anytime a certificate request is made, Candlepin is able to then cut certificates for each CA to hand to clients?

sean797:
I was thinking was:

- CP-node1 would have CA1
- CP-node2 would have CA2

When my client registers or attaches a new sub, one of those nodes would pick up the request and give my client a cert & key. So my client would potentially have:
- client key from CA1
- client cert from CA1
- CA cert contain both CA1 & CA2 (this comes from the katello-ca-consumer RPM)

My Pulp nodes would then provide content to clients with client certs from either CA1 or CA2. Their certs would come from a certs.tar (like Smart Proxies today) or similar (as would Katello app nodes) 

But like I said, this need experimentation, other ideas may also work or be a better architecture to go with.

bkearney:
@ehelms Any update on this?

sean797:
It needs some work to get the tests green, but functionally this PR is fine. I'm happy to get tests passing if we want to merge this - there was some debate previously regarding this approach.

zjhuntin:
Here is an RFC being worked on for an issue like this: https://meet.google.com/linkredirect?authuser=1&dest=https%3A%2F%2Fcommunity.theforeman.org%2Ft%2Frfc-redesign-certificate-handling-within-foreman-deployments%2F17933

PR #161, opened 2017-06-26
#19578 - Switch to custom datatype for path validation

NeilHanlon:
No puppetlabs-stdlib data type exists for any sort of relative url, so we have to roll our own. This could _probably_ go into a more generic module, but I figured for simplicity of not having 3+ chained PRs...

http://projects.theforeman.org/issues/19578

### DIFF
--- a/manifests/foreman_proxy_content.pp
+++ b/manifests/foreman_proxy_content.pp
@@ -13,7 +13,7 @@
 #                                   type:Array
 #
 # $certs_tar::                      Path to tar file with certs to generate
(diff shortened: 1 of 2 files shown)


ekohl:
Do we allow relative paths? If so I wonder what the type adds above a plain string.

NeilHanlon:
All the documentation for generating certs uses `~/proxy-certs.tar`, which `Stdlib::Absolutepath` does not support as it's not absolute.

ekohl:
In that case 👍 with just ```String```.

ekohl:
Thanks!

PR #165, opened 2017-08-21
Simplify variable access

ekohl:
This should effectively be no change.


sean797:
I'm not sure if you want to do something similar to `default_ca_name` in https://github.com/Katello/puppet-certs/blob/master/manifests/candlepin.pp#L127 ?

ekohl:
@sean797 I'll look into that.

ekohl:
@sean797 that looks like a good candidate but I'm working on using puppetlabs-java_ks to replace the execs. I'll fix that as well then.

PR #168, opened 2017-09-07
Drop apache username/password

ekohl:
This reduces coupling to other modules and will allow much easier testing while increasing security.

Apache uses the root user to read certificates and private keys so we can tighten the permissions to ```root:root 0640```.

In puppet-katello we already subscribe to it [here](https://github.com/Katello/puppet-katello/blob/451530b0b161d41ab09f10333f3e849c82cda70d/manifests/application.pp#L52) and that is equal.

In puppet-foreman_proxy_content we subscribe to it [here](https://github.com/Katello/puppet-foreman_proxy_content/blob/c0a476813ecda9cdc0af324faa3c8be05c886364/manifests/init.pp#L116-L125) but I'm not 100% sure yet if this is equal. Ordering should be correct but if Apache is actually reloaded is not guaranteed. https://github.com/Katello/puppet-foreman_proxy_content/pull/136 will fix this.

### DIFF
--- a/manifests/apache.pp
+++ b/manifests/apache.pp
@@ -17,7 +17,6 @@ class certs::apache (
   $expiration           = $::certs::expiration,
   $default_ca           = $::certs::default_ca,
   $ca_key_password_file = $::certs::ca_key_password_file,
-  $group                = $::certs::group,
 ) inherits certs {
 
   $apache_cert_name = "${hostname}-apache"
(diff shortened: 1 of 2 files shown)


ekohl:
Rebased, now with acceptance test.

manifests/apache.pp
```diff
@@ -56,18 +55,14 @@
...
       key_pair   => $apache_cert_name,
       key_file   => $apache_key,
       manage_key => true,
-      key_owner  => $::apache::user,
+      key_owner  => 'root',
+      key_group  => 'root',
       key_mode   => '0440',
-      key_group  => $group,
       cert_file  => $apache_cert,
-      notify     => Service['httpd'],
```
sean797: Why remove this?

ekohl: I'd rather solve this at another layer, namely puppet-katello and puppet-foreman_proxy_content where we apply this. The benefit is that this class becomes totally self-contained. https://github.com/Katello/puppet-foreman_proxy_content/pull/138 should be merged before though. In particular https://github.com/Katello/puppet-foreman_proxy_content/pull/138/commits/199b334caf617d51ce539e18e7d11f3be8ec2e1b

sean797: Okay, lets merge the other one first 👍 I'll review it later today.

PR #172, opened 2017-09-13
Fixes #16911 - Make $server_cert_req optional

ekohl:
Replaces GH-126


lib/puppet/provider/cert/katello_ssl_tool.rb
```diff
@@ -15,7 +15,7 @@ def generate!
       FileUtils.mkdir_p(build_path)
       FileUtils.cp(resource[:custom_pubkey], build_path(File.basename(pubkey)))
       FileUtils.cp(resource[:custom_privkey], build_path(File.basename(privkey)))
-      FileUtils.cp(resource[:custom_req], build_path(File.basename(req_file)))
+      FileUtils.cp(resource[:custom_req], build_path(File.basename(req_file))) if resource[:custom_req]
```
ehelms: Granted, looking back at my own comments, I am not sure this even works without deeper testing. But, I think this might fail due to line 10 above.

ekohl: Good one. I added an extra condition but I hadn't reached this point in the code yet to write some proper tests so I'm relying on @ahumbe or @tbrisker to test this.

ehelms:
This appears to not be as simple as this change, as the underlying tool as some requirements and reliance on the CSR existing:

```
[ERROR 2017-09-13 13:07:52 verbose]  Execution of '/bin/katello-ssl-tool --gen-server --set-hostname sat63-qa-rhel7.woodford.example.com --server-cert sat63-qa-rhel7.woodford.example.com-apache.crt --server-key sat63-qa-rhel7.woodford.example.com-apache.key --server-rpm sat63-qa-rhel7.woodford.example.com-apache --rpm-only' returned 33: ERROR: can't find a file that should have been created during an earlier step:
[ERROR 2017-09-13 13:07:52 verbose]        ./ssl-build/sat63-qa-rhel7.woodford.example.com/server.csr
```

ekohl:
I took a brief look at katello-ssl-tool but I shouldn't because it's just a mess in terms of coding style.

The whole situation confuses me: if you pass in a certificate (that must exist) then it will still require a cert? I would expect it to either require a cert or a csr, not both.

tbrisker:
I took a peek at it and :scream: - who would be the relevant person to talk to about getting this fixed?

ekohl:
image

ekohl:
The current implementation uses the suggestion from Bugzilla and creates an empty CSR file. In my limited testing this works but I'd appreciate some real tests.

ekohl:
Now with an acceptance test

lib/puppet/provider/cert/katello_ssl_tool.rb
```diff
@@ -15,7 +15,11 @@ def generate!
       FileUtils.mkdir_p(build_path)
       FileUtils.cp(resource[:custom_pubkey], build_path(File.basename(pubkey)))
       FileUtils.cp(resource[:custom_privkey], build_path(File.basename(privkey)))
-      FileUtils.cp(resource[:custom_req], build_path(File.basename(req_file)))
+      if resource[:custom_req]
+        FileUtils.cp(resource[:custom_req], build_path(File.basename(req_file)))
+      else
+        File.open(build_path(File.basename(req_file)), 'w') { |f| f.write('') }
```
ehelms: Seems a smidge ugly to write an empty file. Would it not be more appropriate to fix our SSL tool so this parameter is not required?

chris1984: :hammer: but this is so much easier :)

ekohl: @ehelms yes it would, but I'm afraid that's opening such a can of worms that I'll spend days to learn it. On the other hand it might be needed at some point, depending on where we want to go in the long term.

tbrisker:
Ping?

PR #173, opened 2017-09-14
Use puppetlabs-java_ks

ekohl:
Before this change there were a lot of complex execs. This cleans it up by using puppetlabs-java_ks.

We can also remove unused defines that were introduced for gutterball in 2220ef508ad09bd346ba4cf295edb3f9a1f2e1b8 but 7d5473faa13d9394698d98c8277d3b30e5472bfb removed gutterball again.

This includes https://github.com/Katello/puppet-certs/pull/145.


.fixtures.yml
```diff
@@ -6,6 +6,7 @@ fixtures:
     trusted_ca:        "https://github.com/jlambert121/jlambert121-trusted_ca"
     concat:            "https://github.com/puppetlabs/puppetlabs-concat"
     apache:            "https://github.com/puppetlabs/puppetlabs-apache"
+    java_ks:           "https://github.com/puppetlabs/puppetlabs-java_ks"
```
sean797: Interesting that we use the puppetlabs modules for tests but also host our copies. e.g If production releases include https://github.com/theforeman/puppetlabs-apache then we should probably run tests with that. from katello-installer or foreman-installer spec its not clear what the source of these are.

Do we need to make a copy of https://github.com/theforeman/puppetlabs-java_ks in theforeman namespace?

ekohl: We've only used our forked puppetlabs-apache when puppetlabs was too slow with releases but luckily they've started to pick up speed there so we're on regular pl-apache releases.

timogoebel:
@ekohl: This looks pretty good.
It might be because it's Friday evening, but I don't get while `candlepin.pp` still needs `certs::ssltools::certutil`. Can you explain that?

ekohl:
@timogoebel it might be possible that that's no longer required, I'm also somewhat unfamiliar with it. Hence why I was writing tests first.

ekohl:
Running this through the pipeline I get a failure because tomcat doesn't bind to port 8443 while it does if I change the storetype to JKS in /etc/tomcat/server.xml. That surprises me because the acceptance test implies that before this patch it was also generated in that way. If I change the keystore type in server.xml it does work.

sean797:
@ekohl can you rebase please

ekohl:
@sean797 done

ekohl:
@sean797 you didn't run into the JKS/PKCS12 storetype issue with tomcat?

sean797:
@ekohl I did not... Though I am using a slightly fudged nightly, because some RPMs haven't built in a while.

ekohl:
@sean797 I had similar issues so I'm holding back with this till that's solved.

sean797:
Okay, spent some time testing this, I'm not sure exactly what to do to get it working.

I think `storetype   => 'pkcs12'` needs to be set, without it will create JKS. Though I'm not sure what `source_password` should be set to then.

I'm not sure if qpid.pp should be updated to use `java_ks` ?
https://github.com/theforeman/puppet-certs/blob/c5d6d12832b3b4952f046317fdb04dc75b5aee83/manifests/qpid.pp#L73-L82

ekohl:
@sean797 thanks for testing. I updated it, but didn't have time to test it yet and somewhat busy. I didn't migrate qpid since that's NSS rather than openssl or keytool.

manifests/candlepin.pp
```diff
@@ -120,17 +113,26 @@
...
-      creates => $amqp_truststore,
-    } ~>
-    exec { 'import client certificate into Candlepin keystore':
-      # Stupid keytool doesn't allow you to import a keypair.  You can only import a cert.  Hence, we have to
-      # create the store as an PKCS12 and convert to JKS.  See http://stackoverflow.com/a/8224863
-      command => "openssl pkcs12 -export -name amqp-client -in ${client_cert} -inkey ${client_key} -out /tmp/keystore.p12 -passout file:${password_file} && keytool -importkeystore -destkeystore ${amqp_keystore} -srckeystore /tmp/keystore.p12 -srcstoretype pkcs12 -alias amqp-client -storepass ${keystore_password} -srcstorepass ${keystore_password} -noprompt && rm /tmp/keystore.p12",
-      unless  => "keytool -list -keystore ${amqp_keystore} -storepass ${keystore_password} -alias amqp-client",
-    } ~>
+    } ->
+    java_ks { 'candlepin:truststore':
```
ehelms: Note this does this to the `amqp` truststore, we need a similar declaration for the `keystore` (thats what https://github.com/theforeman/puppet-certs/pull/245/files#diff-0a129cc5a540f220caade2da4fc3229bR113 is trying to solve)

ehelms:
We generate two keystores, one for AMQP and one for Tomcat that are also used as truststores. I have a feeling we could trim this down to a single keystore and truststore or a single keystore for all communication. However, that could be a follow up.

ehelms:
Hitting errors:

```
etc/candlepin/certs/candlepin-ca.crt to /etc/candlepin/certs/amqp/candlepin.truststore...
Enter destination keystore password:  Enter source keystore password:  keytool error: java.io.IOException: toDerInputStream rejects tag type 45
Error: /Stage[main]/Certs::Candlepin/Java_ks[candlepin:truststore]/ensure: change from 'absent' to 'latest' failed: Execution of 'keytool -importkeystore -srcstoretype PKCS12 -destkeystore /etc/candlepin/certs/amqp/candlepin.truststore -srckeystore /etc/candlepin/certs/candlepin-ca.crt' returned 1: Importing keystore /etc/candlepin/certs/candlepin-ca.crt to /etc/candlepin/certs/amqp/candlepin.truststore...
Enter destination keystore password:  Enter source keystore password:  keytool error: java.io.IOException: toDerInputStream rejects tag type 45

````

ekohl:
Is that on EL8?

ekohl:
Looks like https://github.com/puppetlabs/puppetlabs-java_ks/commit/a8bdb416059f787fc51294ab2f2a11216b946e2f introduced a change that will require files to be pkcs12 encoded if you want to import them into a pkcs12 encoded keystore. Not sure if this was a regression and what we want to do here is simply not possible. I'd propose we first get your other fix in and then proceed with this.

ekohl:
Closing since things have changed and probably doesn't have a lot of benefit anymore.

PR #174, opened 2017-09-24
candlepin: remove nssdb dependency

timogoebel:
I don't see any reasoning for this code being there. Basically, it adds candlepin's `java-client` cert (the cert candlepin uses to connect to qpidd) to the qpidd truststore.
Candlepin doesn't need this to boot up. And it doesn't need this to connect to qpidd. We don't do anything similar elsewhere. I mean, isn't the idea of a PKI to not whitelist individual certificates but trust the CA?

Anyways, this removes another dependency. Goes along with https://github.com/Katello/puppet-katello/pull/215 to remove the qpidd dependency.

### DIFF
--- a/manifests/candlepin.pp
+++ b/manifests/candlepin.pp
@@ -79,8 +79,6 @@ class certs::candlepin (
   $client_key = "${pki_dir}/private/${java_client_cert_name}.key"
 
   if $deploy {
-    include ::certs::ssltools::nssdb
-
     certs::keypair { 'tomcat':
       key_pair  => $tomcat_cert_name,
       key_file  => $tomcat_key,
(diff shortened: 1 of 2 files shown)


ekohl:
I was a bit unsure about this in the first place and agree with your reasoning but I hadn't looked far enough into this to check why it's there in the first place.

ekohl:
Oh, and I think https://github.com/Katello/puppet-certs/blob/f4152faff300dbd1ff972e5575e3934e4a8f9a98/spec/acceptance/candlepin_spec.rb#L24-L26 can be removed because of this.

ehelms:
@ekohl did you check that the Candlepin listener event is happy and receiving events?

ekohl:
@ehelms if it's covered by bats then yes, otherwise no.

timogoebel:
Removed `qpidd` in acceptance test.

ekohl:
Thanks @timogoebel!

PR #175, opened 2017-10-04
Stop managing tomcat keystore symlink

ekohl:
https://github.com/Katello/puppet-katello/commit/4e1371e8c93d75465b1f5b5f8fd025b90c8b9424 started to manage think symlink as well. That means we can stop managing it here.

### DIFF
--- a/manifests/candlepin.pp
+++ b/manifests/candlepin.pp
@@ -13,7 +13,6 @@ class certs::candlepin (
   $amqp_truststore        = $::certs::candlepin_amqp_truststore,
   $amqp_keystore          = $::certs::candlepin_amqp_keystore,
   $amqp_store_dir         = $::certs::candlepin_amqp_store_dir,
-  $tomcat                 = $::certs::tomcat,
   $country                = $::certs::country,
   $state                  = $::certs::state,
   $city                   = $::certs::city,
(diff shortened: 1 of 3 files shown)


manifests/candlepin.pp
```diff
@@ -97,12 +96,6 @@
       command => "openssl pkcs12 -export -in ${tomcat_cert} -inkey ${tomcat_key} -out ${keystore} -name tomcat -CAfile ${ca_cert} -caname root -password \"file:${password_file}\" -passin \"file:${ca_key_password_file}\" ",
       creates => $keystore,
     } ~>
-    file { "/usr/share/${tomcat}/conf/keystore":
```
sean797: This is a different file than https://github.com/Katello/puppet-katello/commit/4e1371e8c93d75465b1f5b5f8fd025b90c8b9424#diff-0a129cc5a540f220caade2da4fc3229bR55 ?

ekohl: The path is different, but since ```/usr/share/tomcat/conf``` points to ```/etc/tomcat``` this is actually the same file.

sean797: So it is 

```
lrwxrwxrwx. 1 root tomcat 11 Apr 19 09:37 /usr/share/tomcat/conf -> /etc/tomcat

PR #184, opened 2017-11-24
add debian support

mdellweg:


templates/rhsm-katello-reconfigure.erb
```diff
@@ -31,6 +31,17 @@ then
 
   # configure rhsm
   # the config command was introduced in rhsm 0.96.6
+  subscription-manager config \
```
sean797: can we DRY this? Maybe wrap just `BASEURL= foo` around an if else?

templates/rhsm-katello-reconfigure.erb
```diff
@@ -31,6 +31,17 @@ then
...
+    --server.hostname="$KATELLO_SERVER" \
+    --server.prefix="$PREFIX" \
+    --server.port="$PORT" \
+    --rhsm.repo_ca_cert="%(ca_cert_dir)s$KATELLO_SERVER_CA_CERT" \
+    --rhsm.baseurl="$BASEURL"
+elif uname -v | grep -qi debian
```
sean797: Can we use /etc/os-release or a fact for this? Though I don't want to add another dependency.

templates/rhsm-katello-reconfigure.erb
mdellweg: sure

templates/rhsm-katello-reconfigure.erb
mdellweg: ATM we transform the built RPM into a DEB. That way we can install and run the script.

mdellweg:
On a second thought, i'm not so sure, the DRY-ing was worth it, because there are some specialities going on around older rhel versions. I kept it on a separate commit to ease the comparison.

sean797:
> ATM we transform the built RPM into a DEB. That way we can install and run the script.

Are you suggesting that is the official documented way for users to register Debian clients?

mdellweg:
No, i'd prefer a proper debian package of course. Or considering, that this packet really only contains one script that is run on install, we could just run `/pub/katello-rhsm-consumer` from the katello-server directly.

sean797:
Sorry I never got round to looking at this again earlier. Yes I agree, lets remove that second commit.

I think there is a discussion to have around how to we register clients, I'm thinking we should change to running `/pub/katello-rhsm-consumer` on a client and deprecate the RPM method. I think this means that we can stop using the dependence's and tooling we need to generate the RPM without adding more for generating an DEP and keeping the method for registering clients the same.

I'm not suggesting this PR deprecates RPM, just that I think the method on both systems should be the same, if not very similar.

@ehelms @jlsherrill @ekohl others; agree? disagree? any other views?

ekohl:
I'm not that familiar with the subscription manager workflow.

jlsherrill:
@sean797 I think what you said makes sense, I'm just not sure how users would react?  I *think* that if i were an admin i would prefer an rpm, but mainly because i tend to not want to run random scripts from a webserver.  With the rpm I can sign it with a gpg key from my company, i can't as easily do that with a random script. 

In reality I'm not sure if users *actually* care.  The script was added to support atomic hosts (who can't install an rpm), but it does make sense to try to reduce the bootstrap steps to a single set of instructions.

ekohl:
> The RPM method, like you say, can be made more secure be signing the RPM, though I've never seen anyone actually do that.

Isn't that the goal of https://github.com/theforeman/puppet-certs/pull/188?

PaulSD:
No, the goal of #188 is to facilitate signing/verifying the repo metadata that is generated by Pulp, not the RPMs themselves.  It is assumed that the RPMs served by Pulp will already be signed when Pulp retrieves them from the upstream repo ... So Pulp doesn't need to re-sign the RPMs, it only needs to sign the new metadata that Pulp itself generates.

In addition, #188 updates this bootstrap RPM so that it will install the GPG key generated by #188, so clients are unlikely to trust that GPG key until after this bootstrap RPM is installed, so it won't help to sign this bootstrap RPM with that key...

ekohl:
@PaulSD thanks for explaining that distinction. I'll get back to reviewing it tomorrow.

ekohl:
I'm leaning to including this in https://github.com/theforeman/puppet-certs/pull/192. @stbenjam could you have a look as well?

ekohl:
Thanks!

PR #187, opened 2017-12-15
Correct documentation on Candlepin certs

ekohl:


README.md
```diff
@@ -38,7 +38,8 @@ for communication between services inside the Katello deployment.
 | ${hostname}-parent-cert       | a client certificate to read content from Pulp parent (distributed to the child over qpid) | default |
 | ${hostname}-qpid-broker       | a client certificate for qpid broker                                                       | default |
 | ${hostname}-qpid-client-cert  | a client certificate for Pulp to connect to qpid                                           | default |
-| java-client                   | a client certificate for Candlpin to connect to qpid                                       | default |
+| ${hostname}-tomcat            | a server certificate for Candlepin http                                                    | default |
```
ekohl: This may not be right. It looks like this is used as intermediates to import into the NSS db.

ehelms: Without doing a hard check, I'd think this certificate was for building up the trustore and keystore for use by Tomcat server. IIRC nssdb is only used for qpidd.

README.md
```diff
@@ -38,7 +38,8 @@ for communication between services inside the Katello deployment.
...
+| java-client                   | a client certificate for Candlepin to connect to qpid                                      | default |
```
ekohl: @ehelms I do think this line is correct. If so, I'll revert the other one.

ehelms: Agreed.

ekohl:
Since now it was just a trivial typo fix I pushed it to master.

PR #189, opened 2018-01-23
Refs #21873 - Switch warn to fail

chris1984:


chris1984:
QE would like the installer to exit out upon trying to create a proxy with the same name as the Katello box.

ehelms:
I looked at the output that failed verification and I did not see the warning appear in the output at all. Does this fail guarantee it appears? I tend to be hesitant to pull in a full hard stop fail when a warning was requested. Fails like this provide no way for a user to ever perform the action where they match (perhaps they have good reason?). This is not me nack-ing this, but rather wanting to step back and ask why the warning does not appear.

chris1984:
I will post the output tomorrow, it does appear with the fail and exits. With the warn it just keeps going and logs it anyway.

ekohl:
An invisible warning that was ignored anyway was my worry in https://github.com/theforeman/puppet-certs/pull/186#pullrequestreview-81558087 as well so hard failures do make sense to me.

chris1984:
@ehelms before the fix it would not show it and let it run, the only time you would see the warning would be with the -v option like so:

Before fix:

~~~
[ WARN 2018-01-24 10:02:02 verbose]  Scope(Class[Certs::Foreman_proxy_content]): The hostname is the same as the foreman-proxy
[ WARN 2018-01-24 10:02:03 verbose]  Compiled catalog for sat.croberts.lan in environment production in 0.49 seconds
[ INFO 2018-01-24 10:02:03 verbose]  Applying configuration version '1516806122'
[ WARN 2018-01-24 10:02:04 verbose]  /Stage[main]/Certs::Foreman_proxy_content/Certs::Tar_create[/root/blah.tar]/Exec[generate /root/blah.tar]/returns: executed successfully
[ WARN 2018-01-24 10:02:04 verbose]  Finished catalog run in 0.55 seconds
[ INFO 2018-01-24 10:02:04 verbose] Puppet has finished, bye!
[ INFO 2018-01-24 10:02:04 verbose] Executing hooks in group post
  Success!
~~~

With fix
~~~
[root@sat ~]# capsule-certs-generate --certs-tar /root/blah.tar --foreman-proxy-fqdn sat.croberts.lan
 The hostname is the same as the foreman-proxy at /usr/share/katello-installer-base/modules/certs/manifests/foreman_proxy_content.pp:31 on node sat.croberts.lan
~~~

From how I have seen customers use the product in support I see this as a failsafe to prevent customers from creating issues for themselves down the road. Also if want to create a capsule/foreman-proxy with the same name and change the sat/katello to a different name then they can use sat-clone and then create new certs correctly.

manifests/foreman_proxy_content.pp
```diff
@@ -24,7 +24,7 @@
   validate_present($foreman_proxy_fqdn)
 
   if $foreman_proxy_fqdn == $::fqdn {
-    warning('The hostname is the same as the foreman-proxy')
+    fail('The hostname is the same as the foreman-proxy')
```
ehelms: I'd consider one slight modification to the wording -- `The hostname is the same as the provided hostname for the foreman-proxy`

chris1984:
@ehelms updated the wording

PR #191, opened 2018-01-25
Add "set -e" to consumer script

mdellweg:
Not sure if you want this, but whenever this option is not set for a shell script, i loose days searching for the line where something failed.

### DIFF
--- a/manifests/rhsm_reconfigure_script.pp
+++ b/manifests/rhsm_reconfigure_script.pp
@@ -8,13 +8,13 @@ define certs::rhsm_reconfigure_script($ca_cert, $server_ca_cert) {
 
   concat::fragment { "${title}+script_start":
     target  => $title,
-    content => "#!/bin/bash\n\n",
+    content => "#!/bin/bash\n\nset -e\n",
     order   => '01',
   }
 
   concat::fragment { "${title}+default_ca_data":
     target  => $title,
-    content => "read -r -d '' KATELLO_DEFAULT_CA_DATA << EOM\n",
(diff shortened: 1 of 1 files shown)


manifests/rhsm_reconfigure_script.pp
```diff
@@ -8,13 +8,13 @@
...
     target  => $title,
-    content => "#!/bin/bash\n\n",
+    content => "#!/bin/bash\n\nset -e\n",
     order   => '01',
   }
 
   concat::fragment { "${title}+default_ca_data":
     target  => $title,
-    content => "read -r -d '' KATELLO_DEFAULT_CA_DATA << EOM\n",
+    content => "read -r -d '' KATELLO_DEFAULT_CA_DATA << EOM || true\n",
```
ekohl: In which case does this fail?

mdellweg: I think, every time. Do not know why, though
```
root@steve-ratzloff:~# read -d '' -r TEST << EOF
> asdf
> EOF
root@steve-ratzloff:~# echo $TEST $?
asdf 1
```

ekohl: You're right: from ```man read```: 
> The exit status is zero, unless end-of-file is encountered

mdellweg: Thanks for diving into this.
But now i remember, that is why `while read LINE` works.

ekohl:
Thanks!

PR #193, opened 2018-02-20
Fix debian OS determination

jturel:
```
Running Transaction                             
  Installing : katello-ca-consumer-centos7-dev2.strangeways.example.com-1.0-1.noarch                                                                                                           1/1 
sed: can't read /etc/os-release: No such file or directory                                       
warning: %post(katello-ca-consumer-centos7-dev2.strangeways.example.com-1.0-1.noarch) scriptlet failed, exit status 2                                                                              
Non-fatal POSTIN scriptlet failure in rpm package katello-ca-consumer-centos7-dev2.strangeways.example.com-1.0-1.noarch                                                                              Verifying  : katello-ca-consumer-centos7-dev2.strangeways.example.com-1.0-1.noarch                                                                                                           1/1 

Installed:                                      
  katello-ca-consumer-centos7-dev2.strangeways.example.com.noarch 0:1.0-1 
```

Hmm, what happened?

```
[root@rhel6 vagrant]# katello-rhsm-consumer
sed: can't read /etc/os-release: No such file or directory
```

Ah! We don't have os-release in EL6.. I think I've fixed it. I also corrected the regex by adding a '^' in front of 'ID' since it was matching too much.

### DIFF
--- a/templates/rhsm-katello-reconfigure.erb
+++ b/templates/rhsm-katello-reconfigure.erb
@@ -8,7 +8,6 @@ PREFIX=<%= scope['certs::katello::deployment_url'] %>
 CFG=/etc/rhsm/rhsm.conf
 CFG_BACKUP=$CFG.kat-backup
 CA_TRUST_ANCHORS=/etc/pki/ca-trust/source/anchors
-OSFAMILY=$(sed -n -e "s/ID\s*=\s*\(.*\)/\1/p" /etc/os-release)
 
 # exit on non-RHEL systems or when rhsm.conf is not found
 test -f $CFG || exit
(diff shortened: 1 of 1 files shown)


jturel:
@mdellweg you might wanna confirm my regex change :)

mdellweg:
Looks good. If you like, you can also add $ to the end of the regex to be really sure.

mdellweg:
Looks good. If you like, you can also add $ to the end of the regex to be really sure.

mdellweg:
Looks good. If you like, you can also add $ to the end of the regex to be really sure.

ekohl:
> Looks good. If you like, you can also add $ to the end of the regex to be really sure.

Isn't that a bit redundant with a greedy regex?

mdellweg:
Sure. I tend to forget, which regex implementations are greedy, and this redundancy might not hurt. But i'm fine with either ways.

mdellweg:
@ekohl  I'm fine.

Sorry for breaking CentOS6 in the first place.

PR #195, opened 2018-03-13
Fixes #22884 - Check RHSM version for hostname-override

johnpmitsch:
The adding of a subscription-manager fact is failing on
stock SLES12 installs as 'hostname -f' doesn't work "out of
the box":

`hostname: Name or service not known`

This is due to /etc/hosts not having the FQDN associated with
127.0.0.1

Its an easy fix, but for automated provisioning/scripting we
want to be sure the bootstrap rpm script will run without
issues.

The subscription manager hostname-override fact isn't
needed for subscription manager versions past 1.18.1-1,
so we can skip this block of code for later versions.

This commit checks the subscription manager version before
adding the hostname-override fact. This is something we
should do anyways and it avoids the error for SLES.
SLES will use subscription-manager-1.20.5-7.1.x86_64 or
later, so this block of code isn't even needed.

### DIFF
--- a/templates/rhsm-katello-reconfigure.erb
+++ b/templates/rhsm-katello-reconfigure.erb
@@ -106,13 +106,19 @@ fi
   service goferd status >/dev/null && \
   service goferd restart >/dev/null 2&>1
 
+[ -f /usr/lib/systemd/system/goferd.service ] && \
 [ -f /bin/systemctl ] && \
   systemctl try-restart goferd >/dev/null 2>&1
 
-
-FQDN=`hostname -f`
-if [ $? == "0" ] && [ "$FQDN" != "localhost" ] && [ -d /etc/rhsm/facts/ ]; then
-  echo "{\"network.hostname-override\":\"$FQDN\"}" > /etc/rhsm/facts/katello.facts
+# EL5 systems and subscription-manager versions before 1.18.1-1 don't have the network.fqdn fact.
+# For these cases, we have to update the "hostname-override" fact
+if (test -f /etc/redhat-release && grep -q -i "Red Hat Enterprise Linux Server release 5" /etc/redhat-release) || \
+   (test -f /etc/centos-release && grep -q -i "CentOS Linux release 5" /etc/centos-release) || \
+   test ${RHSM_VERSION[0]:-0} -lt 1 -o ${RHSM_VERSION[1]:-0} -lt 18 -o \( ${RHSM_VERSION[1]:-0} -eq 18 -a ${RHSM_VERSION[2]:-0} -lt 2 \); then
+  FQDN=`hostname -f`
+  if [ $? == "0" ] && [ "$FQDN" != "localhost" ] && [ -d /etc/rhsm/facts/ ]; then
+    echo "{\"network.hostname-override\":\"$FQDN\"}" > /etc/rhsm/facts/katello.facts
+  fi
 fi
 
 exit 0


templates/rhsm-katello-reconfigure.erb
```diff
@@ -106,13 +106,19 @@ fi
   service goferd status >/dev/null && \
   service goferd restart >/dev/null 2&>1
 
+[ -f /etc/init.d/goferd ] && \
```
ekohl: If goferd is a native systemd unit then this is breaking the restarting. It should be checking for a .service file I think.

templates/rhsm-katello-reconfigure.erb
```diff
@@ -106,13 +106,19 @@ fi
...
 
-
-FQDN=`hostname -f`
-if [ $? == "0" ] && [ "$FQDN" != "localhost" ] && [ -d /etc/rhsm/facts/ ]; then
-  echo "{\"network.hostname-override\":\"$FQDN\"}" > /etc/rhsm/facts/katello.facts
+# If RHSM version is less than 1.18.1-1, we have to update the hostname-override fact, since those versions don't
+# have the network.fqdn fact. This doesn't apply to EL5 systems, where we always want to add the hostname-override
+if (test -f /etc/redhat-release && grep -q -i "Red Hat Enterprise Linux Server release 5" /etc/redhat-release) || \
+   (test -f /etc/centos-release && grep -q -i "CentOS Linux release 5" /etc/centos-release) || \
+   test ${RHSM_VERSION[0]:-0} -lt 1 -o ${RHSM_VERSION[1]:-0} -lt 18 -o \( ${RHSM_VERSION[1]:-0} -eq 18 -a ${RHSM_VERSION[2]:-0} -lt 2 \); then
```
ekohl: This is slightly terrifying and knowing it's already used in another place doesn't soothe me.

templates/rhsm-katello-reconfigure.erb
```diff
@@ -106,13 +106,19 @@ fi
...
+[ -f /etc/init.d/goferd ] && \
 [ -f /bin/systemctl ] && \
   systemctl try-restart goferd >/dev/null 2>&1
 
-
-FQDN=`hostname -f`
-if [ $? == "0" ] && [ "$FQDN" != "localhost" ] && [ -d /etc/rhsm/facts/ ]; then
-  echo "{\"network.hostname-override\":\"$FQDN\"}" > /etc/rhsm/facts/katello.facts
+# If RHSM version is less than 1.18.1-1, we have to update the hostname-override fact, since those versions don't
+# have the network.fqdn fact. This doesn't apply to EL5 systems, where we always want to add the hostname-override
```
ekohl: The ```This doesn't apply to EL5``` is slightly misleading. I'd rephrase it as ```EL5 systems and RHSM before 1.18.1-1 are missing the network.fqdn fact so need a hostname-override fact``` or something along those lines.

templates/rhsm-katello-reconfigure.erb
johnpmitsch: I was copying [a few lines above this](https://git.io/vxUp2), should that change as well? If so, what do you recommend?

johnpmitsch:
Some results from my testing:

### On SLES 12sp3 w/ subscription-manager-1.20.5-7.1.x86_64:
rpm script runs successfully
```
suse12:~/bootstraprpm # rpm -Uvh http://bacon.jomitsch.example.com/pub/katello-ca-consumer-latest.noarch.rpm
Retrieving http://bacon.jomitsch.example.com/pub/katello-ca-consumer-latest.noarch.rpm
Preparing...                          ################################# [100%]
Updating / installing...
   1:katello-ca-consumer-bacon.jomitsc################################# [100%]
suse12:~/bootstraprpm # echo $?
0
```
nothing in katello.facts as expected (the file is only there because I had created it in my testing)
```
suse12:~/bootstraprpm # cat /etc/rhsm/facts/katello.facts 

suse12:~/bootstraprpm # 
```

### On RHEL 5 w/ subscription-manager-1.11.3-14.el5_11
rpm script runs successfully
```
][root@host-8-249-169 ~]# rpm -Uvh katello-ca-consumer-latest.noarch.rpm 
Preparing...                ########################################### [100%]
   1:katello-ca-consumer-bac########################################### [100%]
[root@host-8-249-169 ~]# echo $?
0
```
katello facts is updated as expected
```
[root@host-8-249-169 ~]# cat /etc/rhsm/facts/katello.facts 
{"network.hostname-override":"host-8-249-169.host.centralci.eng.rdu2.redhat.com"}
```

### On centos 7 w/ subscription-manager-1.19.23-1.el7.centos.x86_64
rpm script runs successfully
```
[root@centos7 ~]# rpm -Uvh http://bacon.jomitsch.example.com/pub/katello-ca-consumer-latest.noarch.rpm
Retrieving http://bacon.jomitsch.example.com/pub/katello-ca-consumer-latest.noarch.rpm
Preparing...                          ################################# [100%]
Updating / installing...
   1:katello-ca-consumer-bacon.jomitsc################################# [100%]
[root@centos7 ~]# echo $?
0
```
katello.facts does not exist (as expected)
```
[root@centos7 ~]# cat /etc/rhsm/facts/katello.facts
cat: /etc/rhsm/facts/katello.facts: No such file or directory
```
### On RHEL7 with subscription-manager-1.19.21-1.el7.x86_64
rpm script runs successfully
```
[root@rhel7 ~]# rpm -Uvh http://bacon.jomitsch.example.com/pub/katello-ca-consumer-latest.noarch.rpm
Retrieving http://bacon.jomitsch.example.com/pub/katello-ca-consumer-latest.noarch.rpm
Preparing...                          ################################# [100%]
Updating / installing...
   1:katello-ca-consumer-bacon.jomitsc################################# [100%]
[root@rhel7 ~]# echo $?
0
```
katello.facts is not created as expected
```
[root@rhel7 ~]# cat /etc/rhsm/facts/katello.facts
cat: /etc/rhsm/facts/katello.facts: No such file or directory
```

templates/rhsm-katello-reconfigure.erb
johnpmitsch: We discussed converting this to python, I think that would be a good long term goal and would make complicated logic easier

johnpmitsch:
@ekohl your comment about checking the `goferd.service` file made sense right after I commented. I've updated accordingly :smile:

templates/rhsm-katello-reconfigure.erb
```diff
@@ -106,13 +106,19 @@ fi
...
 [ -f /bin/systemctl ] && \
   systemctl try-restart goferd >/dev/null 2>&1
 
-
-FQDN=`hostname -f`
-if [ $? == "0" ] && [ "$FQDN" != "localhost" ] && [ -d /etc/rhsm/facts/ ]; then
-  echo "{\"network.hostname-override\":\"$FQDN\"}" > /etc/rhsm/facts/katello.facts
+# EL5 systems and subscription-manager versions before 1.18.1-1 don't have the network.fqdn fact.
+# For these cases, we have to update the "hostname-override" fact
+if (test -f /etc/redhat-release && grep -q -i "Red Hat Enterprise Linux Server release 5" /etc/redhat-release) || \
```
parthaa: can you add something like a function for these comparisions. Might be easier to grok whats happening 
```
is_rhel5() {
[ -f /etc/redhat-release && grep -q -i "Red Hat Enterprise Linux Server release 5" /etc/redhat-release ]
}

is_centos5() {
[ -f /etc/centos-release && grep -q -i "CentOS Linux release 5" /etc/centos-release ]
}


```

templates/rhsm-katello-reconfigure.erb
```diff
@@ -106,13 +106,19 @@ fi
...
+   (test -f /etc/centos-release && grep -q -i "CentOS Linux release 5" /etc/centos-release) || \
+   test ${RHSM_VERSION[0]:-0} -lt 1 -o ${RHSM_VERSION[1]:-0} -lt 18 -o \( ${RHSM_VERSION[1]:-0} -eq 18 -a ${RHSM_VERSION[2]:-0} -lt 2 \); then
```
parthaa: Since this comparison is repeated twice. Might be an opportunity to functionify this also

templates/rhsm-katello-reconfigure.erb
```diff
@@ -106,13 +106,19 @@ fi
...
+  FQDN=`hostname -f`
```
parthaa: @jlsherrill @johnpmitsch dont we want to still handle ```hostname -f``` erroring out ?

johnpmitsch: The issue is with `set -e`, `hostname -f` failing will always cause the script to error out, even with `2> /dev/null` or as part of a conditional. I tried multiple ways wihout luck and came to the conclusion that the best way to handle it is to avoid running it completely if its not necessary and that avoids the vast majority of the occurrences since most systems are running a newer subman version

lpramuk:
This will also fix the following issue:
```
# rpm -Uvh http://$SAT/pub/katello-ca-consumer-latest.noarch.rpm
Retrieving http://$SAT/pub/katello-ca-consumer-latest.noarch.rpm
Preparing...                          ################################# [100%]
Updating / installing...
   1:katello-ca-consumer-$SAT################################# [100%]
warning: %post(katello-ca-consumer-$SAT-1.0-1.noarch) scriptlet failed, exit status 5
```
It happens when goferd is not installed and scripts is trying to restart it:
```
# systemctl try-restart goferd ; echo $?
Failed to try-restart goferd.service: Unit not found.
5
```
This part was problematic (since there is newly "set -e" at the beginning of the script)
```
[ -f /bin/systemctl ] && \
  systemctl try-restart goferd >/dev/null 2>&1
```

templates/rhsm-katello-reconfigure.erb
```diff
@@ -106,13 +106,19 @@ fi
   service goferd status >/dev/null && \
   service goferd restart >/dev/null 2&>1
 
+[ -f /usr/lib/systemd/system/goferd.service ] && \
```
lpramuk: +1 you saved me from filing issue by this change

templates/rhsm-katello-reconfigure.erb
lpramuk: or ```FQDN=$(hostname -f || :)``` ?
but then the check ```[ $? == "0" ]``` is useless and can be omitted

parthaa:
ack

templates/rhsm-katello-reconfigure.erb
johnpmitsch: @lpramuk I tried using hostname -f in a conditional as I mentioned above, it is still evaluated and fails the script

PR #197, opened 2018-05-09
Remove dependency on puppet-common

jturel:
there wasn't a dependency on puppet-common in the manifest. that seems unexpected!

### DIFF
--- a/spec/classes/certs_foreman_proxy_content_spec.rb
+++ b/spec/classes/certs_foreman_proxy_content_spec.rb
@@ -7,12 +7,20 @@ describe 'certs::foreman_proxy_content' do
 
   let :params do
     {
-      :certs_tar => '/tmp/tar',
-      :foreman_proxy_fqdn => 'bar.example.com'
+      certs_tar: '/tmp/tar',
+      foreman_proxy_fqdn: 'bar.example.com'
     }
   end
 
   describe 'with default parameters' do
     it { should compile.with_all_deps }
+  end
+
+  context 'with empty certs_tar' do
+    let(:params) { { certs_tar: '' } }
+
+    it do
+      should compile.and_raise_error(/\'certs_tar\' expects a String\[1/)
+    end
   end
 end
--- a/spec/functions/validate_file_exists_spec.rb
+++ b/spec/functions/validate_file_exists_spec.rb
@@ -1,0 +1,8 @@
+require 'spec_helper'
+require 'puppetlabs_spec_helper/puppetlabs_spec/puppet_internals'
+
+describe 'validate_file_exists' do
+  it { should run.with_params('foo_doesnt_exist').and_raise_error(Puppet::ParseError) }
+  it { should run.with_params('foo_doesnt_exist').and_raise_error(/foo_doesnt_exist does not exist/) }
+  it { should run.with_params('/').and_return(true) }
+end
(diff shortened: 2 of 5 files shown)


lib/puppet/parser/functions/validate_present.rb
```diff
@@ -0,0 +1,15 @@
+module Puppet::Parser::Functions
```
ekohl: I think we can get rid of this since we have puppet data types. It means changing `Optional[String] $certs_tar` to `String[1] $certs_tar`. The fqdn check should already be redundant. It also means we can remove the `if $certs_tar` since that looks like a useless check.

jturel: According to the docs it's letting us use `newfunction` https://docs.puppet.com/puppet/4.0/yard/Puppet/Parser/Functions.html#newfunction-class_method

Do you mean it's somehow redundant?

ekohl: I mean the whole function can be removed since we can simply replace all function calls with Puppet 4 data types.

spec/classes/certs_foreman_proxy_content_spec.rb
```diff
@@ -15,4 +15,16 @@
...
     it { should compile.with_all_deps }
   end
+
+  context 'with empty certs_tar' do
+    let(:params) { { certs_tar: '' , foreman_proxy_fqdn: 'example.com' } }
+
+    it { should_not compile.with_all_deps }
+  end
+
+  context 'missing foreman_proxy_fqdn' do
```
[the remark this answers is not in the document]
jturel: makes sense!

spec/classes/certs_foreman_proxy_content_spec.rb
```diff
@@ -15,4 +15,10 @@
   describe 'with default parameters' do
     it { should compile.with_all_deps }
   end
+
+  context 'with empty certs_tar' do
+    let(:params) { { certs_tar: '' , foreman_proxy_fqdn: 'example.com' } }
```
ekohl: We can probably leave out the fqdn parameter (and in the basic test too).

spec/functions/validate_file_exists_spec.rb
```diff
@@ -0,0 +1,7 @@
+require 'spec_helper'
+require 'puppetlabs_spec_helper/puppetlabs_spec/puppet_internals'
+
+describe 'validate_file_exists' do
```
ekohl: This can probably use a test and verify that `/` does exist.

spec/classes/certs_foreman_proxy_content_spec.rb
jturel: Looks like basic tests needs the default fqdn for one scenario:
```
    Failure/Error: it { should compile.with_all_deps }
       error during compilation: Evaluation Error: Error while evaluating a Function Call, The hostname is the same as the provided hostname for the foreman-proxy at /home/travis/build/theforeman/puppet-certs/spec/fixtures/modules/certs/manifests/foreman_proxy_content.pp:22:5 on node travis-job-theforeman-puppet-certs-377315866.travisci.net
```
```
certs::qpid_client
  on centos-7-x86_64
    without parameters
    should compile into a catalogue without dependency cycles (FAILED - 1)
```

ekohl: Ah right :)

jturel:
> @jturel Can you also remove puppet-common from https://github.com/Katello/katello-installer/blob/master/Puppetfile#L6?

ah ok! I was missing where the dependency was actually included

ekohl:
Thanks!

PR #199, opened 2018-05-15
allow creating certs_tar for Katello app

sean797:
For https://github.com/Katello/katello-installer/pull/628

### DIFF
--- a/.sync.yml
+++ b/.sync.yml
@@ -9,6 +9,7 @@
 Rakefile:
   param_docs_pattern:
     - manifests/foreman_proxy_content.pp
+    - manifests/generate_archive.pp
     - manifests/init.pp
 spec/spec_helper_acceptance.rb:
   install_epel: true
--- a/manifests/generate_archive.pp
+++ b/manifests/generate_archive.pp
@@ -1,0 +1,46 @@
+# == Class: certs::generate_archive
+#
+# === Parameters:
+#
+# $server_fqdn::                   FQDN of the Foreman app
+#
+# $server_cname::                  Additional names of the foreman app
+#
+# $certs_tar::                     Path to tar file with certs to generate
+#
+# $foreman_application::           Create certificate bundle for a Foreman node
+#
+# $foreman_proxy::                 Create certificate bundle for a Foreman Proxy node
+#
+class certs::generate_archive (
+  String $server_fqdn          = undef,
+  Array[String] $server_cname  = $::certs::params::cname,
+  String[1] $certs_tar         = $::certs::params::certs_tar,
+  Boolean $foreman_application = false,
+  Boolean $foreman_proxy       = false,
+) inherits certs::params {
+
+  if $server_fqdn == $::fqdn {
+    fail('The hostname is the same as the provided hostname for the Foreman app')
+  }
+
+  unless $foreman_application or $foreman_proxy {
+    fail('What type of certs bundle do you want to generate? Either $foreman_application or $foreman_proxy must be true')
+  }
+
+  if $foreman_application {
+    $class_list = ['::certs::puppet', '::certs::foreman', '::certs::apache', '::certs::qpid_client']
+  } elsif $foreman_proxy {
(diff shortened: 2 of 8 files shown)


manifests/tar_create.pp
```diff
@@ -10,11 +10,11 @@
 #
 define certs::tar_create(
   $path               = $title,
-  $foreman_proxy_fqdn = $::certs::foreman_proxy_content::foreman_proxy_fqdn,
+  $fqdn               = $::certs::foreman_proxy_content::foreman_proxy_fqdn,
```
ekohl: I wonder if this could overlap with the fqdn fact.

sean797:
@ekohl I cleaned up the code, removed `foreman_proxy-content` class and added some tests. This is a breaking change for https://github.com/Katello/katello-installer/blob/master/bin/foreman-proxy-certs-generate#L65. I'll update https://github.com/Katello/katello-installer/pull/628 with all the required changes.

manifests/generate_archive.pp
```diff
@@ -0,0 +1,46 @@
...
+#
+# $foreman_application::           Create certificate bundle for a Foreman node
+#
+# $foreman_proxy::                 Create certificate bundle for a Foreman Proxy node
+#
+class certs::generate_archive (
+  String $server_fqdn          = undef,
+  Array[String] $server_cname  = $::certs::params::cname,
+  String[1] $certs_tar         = $::certs::params::certs_tar,
+  Boolean $foreman_application = false,
```
ekohl: This is the general direction I'm going for but I'm not sure this is the granularity I'm aiming for. My goal is to split the node with candlepin and pulp into two separate nodes. Ideally qpid could run on a third. That means we'd bootstrap with candlepin as a node with the CA, then generate the certs for each additional node with the role they have. Do you think this makes sense?

sean797: Yes, and I agree its what we should aim for, well my end goal is being able to scale each component independently of each other.

Initially, I think we should try to make small(er) changes and iterate on them. Fewer chances of breakages and easier to review, we can extend the scope in other PRs later.

ekohl: I've marked it as WIP for now so we don't accidentally merge it. I agree we should iterate on this but since it's a breaking change I'd like to keep it open until we're really happy with it.

sean797:
I've made this backwards compatible as I want to protect `foreman-proxy-certs-generate`, If we were to make it use the new manifest then `--foreman-proxy-fqdn` arg would change to `--server-fqdn`.

manifests/generate_archive.pp
```diff
@@ -0,0 +1,46 @@
+# Prepare the certificates for the node from the parent node
```
ehelms: Could you maintain backwards compatibility for the time being by having the foreman_proxy_content class call this class with parameters?

sean797:
Thanks @ehelms, updated.

.sync.yml
```diff
@@ -8,7 +8,8 @@
   test_all_rvms: false
 Rakefile:
   param_docs_pattern:
-    - manifests/foreman_proxy_content.pp
+    - manifests/foreman_proxy_content.pp
```
ekohl: Can you remove the trailing whitespace?

manifests/tar_create.pp
```diff
@@ -9,12 +9,12 @@
 # $foreman_proxy_fqdn::         FQDN of the foreman proxy
 #
 define certs::tar_create(
-  $path               = $title,
-  $foreman_proxy_fqdn = $::certs::foreman_proxy_content::foreman_proxy_fqdn,
+  $path    = $title,
+  $sub_dir = $::certs::foreman_proxy_content::foreman_proxy_fqdn,
```
ekohl: Given we always pass this in I think we should remove the default value here. We're breaking the API anyway by changing the name.

sean797:
Updated, I agree about `certs::tar_create` being an internal API ;)  

Thanks!

PR #201, opened 2018-05-19
Clean up $nss_db_dir handling

ekohl:
Both in candlepin.pp and in params.pp they are unused so we can remove them. For qpid it shouldn't be a class parameter because one could override it in the nssdb class and things would break.

Technically we should remove it from init.pp too and only define it in nssdb.pp as a class parameter default, but puppet-katello uses it on the main class so we keep it in for now.


ekohl:
Related: https://github.com/theforeman/puppet-katello/pull/242

PR #204, opened 2018-07-09
fixes #24210 - certs group must be overridable

stbenjam:
For Katello, we override the certs ownership so Foreman can read it.
This is required for WebVNC: the foreman user launches websockify
using the certificate.

Partially reverts https://github.com/theforeman/puppet-certs/pull/168


stbenjam:
Yup, easy to miss :-\ It doesn't help websockify.py has no logs to find out the problem.

Thanks for quick review

ekohl:
Don't we log that to production.log?

stbenjam:
I don't think so, but even if it did, websockify itself doesn't seem to have any output when it can't read the certs, it still starts.  Maybe I can fix that, too.  Will have a look

PR #208, opened 2018-07-27
set empty-password for nssdb as FIPS fails on password file

amitkarsale:


ekohl:
Odd that the type output changes from type `JKS` to `jks`. Any idea?

ekohl:
Looks like it also changed on master to lowercase. That's slightly unexpected.

amitkarsale:
@ekohl sorry am unaware what could have caused it, can you share your thoughts over the cause of failure.

PR #209, opened 2018-07-30
msync: changelog user & custom facts in tests

ekohl:


evgeni:
for some reason the keystore type is printed in lowercase sometimes. should we make the tests case-insensitive?

ekohl:
I was thinking about this. I can't quite figure out why this would be changed to lowercase but i guess making it lower case would be good enough now.

PR #210, opened 2018-08-03
[WIP] Allow split deployments

ekohl:
Includes https://github.com/theforeman/puppet-certs/pull/199

### DIFF
--- a/.sync.yml
+++ b/.sync.yml
@@ -9,6 +9,7 @@
 Rakefile:
   param_docs_pattern:
     - manifests/foreman_proxy_content.pp
+    - manifests/generate_archive.pp
     - manifests/init.pp
 spec/spec_helper_acceptance.rb:
   install_epel: true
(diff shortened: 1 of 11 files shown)


sean797:
@ekohl Is this still WIP? I think its ready. From a Katello release perspective, I think it's totally okay to say, this is in 'tech preview' or alike, maybe not mention it at all.

Maybe we should remove the deprecation warning and merge, once we are more happy with the split deployments and it's had more testing, probably in a couple of releases time, we can add the deprecation warning and announce the feature.

ekohl:
I don't think I've tested this very much but you're right that 1.20 is coming up and we should start moving into a direction. I do want to modify it to a future proof API. That means a single parameter for the aspects instead of 2 booleans.

ekohl:
To get as much code I've split off https://github.com/theforeman/puppet-certs/pull/214 which can be merged already.

ekohl:
Another part is https://projects.theforeman.org/issues/24947 which I think we can get in standalone as well. I have created the installer PRs and I'm going to do a scratch build to test with.  We'll probably also need to update a lot of docs.

ekohl:
@timogoebel suggested that you want this available as a define so you can generate multiple certs. Other than that I still consider this WIP.

ekohl:
There's an RFC open to redesign things and this needs a major rebase now. Closing for now.

PR #211, opened 2018-08-07
set empty-password for nssdb as FIPS fails on password file

amitkarsale:


iNecas:
@ekohl seems like test failures are unrelated.

ekohl:
Can you rebase on top of master? https://github.com/theforeman/puppet-certs/commit/2e40c3aa458c8553dbf735048b47c4cdec0d4c4c has fixed the current master.

amitkarsale:
@ekohl can we merge this ?

ekohl:
I'm wondering if this doesn't actually lower security. My intuition says that a file with the plaintext password is still more secure than no password at all. Have you verified that not any user on the system can read the content of the password store?

iNecas:
Seems like we might not need the `--empty-password` if we passed `-f '${::certs::ssltools::nssdb::nss_db_password_file}'` option to https://github.com/theforeman/puppet-certs/blob/b3e8918f6ca553b750fbcba14013cb86a33df9ed/manifests/ssltools/certutil.pp#L15

@amitkarsale - mind giving this change a try?

upadhyeammit:
@iNecas @amitkarsale I just tested this change and I do see installation works without any problem after doing changes in certutil.pp file, I can also see that /etc/pki/katello/nssdb/nss_db_password-file file is actually there and it has password in ASCII format. Is this expected result ?

@ekohl here password store means the /etc/pki/katello/nssdb/nss_db_password-file which is having read permissions for root and qpidd user only I do not see its accessible to any other user,

ls -l /etc/pki/katello/nssdb/nss_db_password-file 
-rw-r-----. 1 root qpidd 33 Sep 12 15:19 /etc/pki/katello/nssdb/nss_db_password-file

ekohl:
It sounds like a real bug was found and @iNecas has the correct patch.

The NSS DB is only used by qpidd so those permissions look correct.

iNecas:
@upadhyeammit would you mind sending a PR with your fix to `certutil.pp` - I assume you've been testing it with FIPS enabled mode

amitkarsale:
@iNecas the changes you mentioned works fine with the FIPS mode on. Have resubmitted the PR

have verified the installation with the new code change and the installation was success.

upadhyeammit:
> @upadhyeammit would you mind sending a PR with your fix to `certutil.pp` - I assume you've been testing it with FIPS enabled mode

Yes tested with FIPS mode on, discussed with @amitkarsale and he will be sending new change.

iNecas:
@amitkarsale ping on @ekohl question

ekohl:
Could you also look at the deletion command above it? It looks like it'll also need a password

amitkarsale:
Opps I missed delete command.

amitkarsale:
I've tried the deleting the certificates manually and it allows to delete without passing the password file. Deletion is not needing password to delete the certs.

manifests/ssltools/certutil.pp
```diff
@@ -12,8 +12,8 @@
   } ->
   exec { $cert_name:
     path        => ['/bin', '/usr/bin'],
-    command     => "certutil -A -d '${nss_db_dir}' -n '${cert_name}' -t '${trustargs}' -a -i '${client_cert}'",
-    unless      => "certutil -L -d ${nss_db_dir} | grep '^${cert_name}\\b'",
+    command     => "certutil -A -d '${nss_db_dir}' -n '${cert_name}' -t '${trustargs}' -a -i '${client_cert}' -f '${::certs::ssltools::nssdb::nss_db_password_file}'",
+    unless      => "certutil -L -d ${nss_db_dir} | grep '^${cert_name}\\b' -f '${::certs::ssltools::nssdb::nss_db_password_file}'",
```
iNecas: Please not that you're passing the `-f` option to `grep`, not to `certutil`

amitkarsale:
@ekohl @iNecas are there any open point to fix you can think of?

manifests/ssltools/certutil.pp
```diff
@@ -12,8 +12,8 @@
   } ->
   exec { $cert_name:
     path        => ['/bin', '/usr/bin'],
-    command     => "certutil -A -d '${nss_db_dir}' -n '${cert_name}' -t '${trustargs}' -a -i '${client_cert}'",
-    unless      => "certutil -L -d ${nss_db_dir} | grep '^${cert_name}\\b'",
+    command     => "certutil -A -d '${nss_db_dir}' -n '${cert_name}' -t '${trustargs}' -a -i '${client_cert}'  -f '${::certs::ssltools::nssdb::nss_db_password_file}'",
+    unless      => "certutil -L -d ${nss_db_dir}  -f '${::certs::ssltools::nssdb::nss_db_password_file}' | grep '^${cert_name}\\b'",
```
iNecas: I've tested this and `certutil -L` doesn't even have the `-f` option, and it doesn't need it: I suggest not providing it "just in case" , as it might be just misleading.

I've tested this and with the `-A` update, everything worked just fine

iNecas:
Test failures seem unrelated. Tested and it works well. @ekohl mind merging this one?

ekohl:
Thanks!

iNecas:
Thanks @ekohl. @ekohl or @ehelms when can we expect this changes to land in nighties?

ekohl:
If all goes well tonight.

PR #215, opened 2018-09-14
Refs #24947 - Allow extract tar files at the top level

ekohl:
This moves the parameter from puppet-katello and puppet-foreman_proxy_content into puppet-certs. This will need an installer migration as well.


ekohl:
Rebased

PR #220, opened 2018-10-24
Add support for debian derivatives

m-bucher:
Correctly detect debian derivatives like Ubuntu, Linux Mint, etc.


ekohl:
Failure is unrelated and I'll need to look into it.

PR #223, opened 2018-10-31
Fixes #25359 - Add name flag to openssl pkcs12 nsddb key/cert convert.

chris1984:


chris1984:
@ekohl im undecided if we should add broker as a hardcoded value or add a param that can be used so a user can define a name? I am up for whatever you want to go with.

ekohl:
> @ekohl im undecided if we should add broker as a hardcoded value or add a param that can be used so a user can define a name? I am up for whatever you want to go with.

It's already hardcoded and so far it hasn't been an issue. I'm also not sure if candlepin cares about the name. Let's leave it for the future.

PR #225, opened 2018-11-16
Move keystore to Candlepin default location

ehelms:


ekohl:
Looks like you do need to update the acceptance tests to expect it in a different location now.

PR #227, opened 2018-11-19
Fixes #25512 fixes incorrect editing of rhsm.conf

patilsuraj767:
rhsm-katello-reconfigure.erb template had incorrect sed syntax due to which katello-ca-consumer rpm was creating with incorrect /usr/bin/katello-rhsm-consumer.

### DIFF
--- a/templates/rhsm-katello-reconfigure.erb
+++ b/templates/rhsm-katello-reconfigure.erb
@@ -98,7 +98,7 @@ else
     sed -i "s/full_refresh_on_yum\s*=.*$/full_refresh_on_yum = 1/g" $CFG
   else
     full_refresh_config="#config for on-premise management\nfull_refresh_on_yum = 1"
-    sed -i "/baseurl/a\n$full_refresh_config" $CFG
+    sed -i "/baseurl/a\ $full_refresh_config" $CFG
   fi
 fi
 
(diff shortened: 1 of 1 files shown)


patilsuraj767:
@ekohl  In case of  RHEL 5 subscription-manager package is of older version. Code first checks for the subscription-manager version if it is older it uses the sed command to edit the rhsm.conf instead of "subscription-manager config"  as it is not present in older version of subscription manager. 

sed command in the code tries to add the comment in the rhsm.conf file but because of "n" typed in the code, it enters the line the rhsm.conf as "n#config for on-premise management" due to subscription-manger command throws error.

patilsuraj767:
@ekohl  This was changed in 771e11671371da5c53689bbd0ddbd506843f9f34

templates/rhsm-katello-reconfigure.erb
```diff
@@ -98,7 +98,7 @@ else
     sed -i "s/full_refresh_on_yum\s*=.*$/full_refresh_on_yum = 1/g" $CFG
   else
     full_refresh_config="#config for on-premise management\nfull_refresh_on_yum = 1"
-    sed -i "/baseurl/a\n$full_refresh_config" $CFG
+    sed -i "/baseurl/a\ $full_refresh_config" $CFG
```
ekohl: You can also remove the `\`. Otherwise the result is a leading space for the comment:
```console
$ cat test.repo 
[x]
baseurl = bla
sdfsd
$ sed "/baseurl/a\ $full_refresh_config" test.repo 
[x]
baseurl = bla
 #config for on-premise management
full_refresh_on_yum = 1
sdfsd
$ sed "/baseurl/a $full_refresh_config" test.repo 
[x]
baseurl = bla
#config for on-premise management
full_refresh_on_yum = 1
sdfsd
```

patilsuraj767: Done.

PR #228, opened 2018-11-19
 Deploy CA cert and key to Candlepin default locations

ehelms:


manifests/candlepin.pp
```diff
@@ -78,6 +78,27 @@
   $client_key = "${pki_dir}/private/${java_client_cert_name}.key"
 
   if $deploy {
+    privkey { $ca_key:
+      key_pair      => $default_ca,
+      unprotect     => true,
```
ekohl: Would it make sense to enhance https://github.com/theforeman/puppet-certs/blob/master/manifests/keypair.pp with these parameters to reduce duplication?

The alternative is to pass in `candlepin::ca_key_password`

ehelms:
@ekohl Green now. Note that I had to change the definition of the keypair type a bit to fit this model update.

manifests/candlepin.pp
```diff
@@ -106,7 +119,7 @@
       mode   => '0750',
     } ~>
     exec { 'import CA into Candlepin truststore':
-      command => "keytool -import -v -keystore ${amqp_truststore} -storepass ${keystore_password} -alias ${::certs::default_ca_name} -file ${ca_cert} -noprompt",
+      command => "keytool -import -v -keystore ${amqp_truststore} -storepass ${keystore_password} -alias candlepin-ca -file ${ca_cert} -noprompt",
```
ekohl: It'd still be nice to store `candlepin-ca` in a variable, even if it's just a local one in the class. That way we can refer to it in puppet-katello instead of copying it.

ehelms: I don't think we will reference this naming anywhere else.

ehelms:
Updated @ekohl

ekohl:
I do wonder if we'll see issues on upgrade because of the alias change.

ehelms:
I'll do a pre-emptive upgrade pipeline run with this. If it passes I will merge it, if not I will report back here.

PR #148, opened 2017-03-28
Deploy qpid client certificate and key

ehelms:

PR #229, opened 2018-11-26
Move qpid client cert to /etc/pki/pulp

ehelms:


ehelms:
Updated

manifests/qpid_client.pp
```diff
@@ -39,16 +40,25 @@
 
   if $deploy {
 
+    file { $certs::pulp_pki_dir:
```
ekohl: I like explicit ownership and permissions. Otherwise things are affected by umask and such.

ehelms:
You are mistakening the qpid client certs and the Pulp client certs. This only touches the qpid client certificates which are not used by Pulp. Qpid client certificates are for talking to qpid, Pulp client certificates are used for Katello talking to Pulp.

Updated with file perms

manifests/qpid_client.pp
```diff
@@ -39,16 +40,31 @@
...
+      ensure => directory,
+      owner  => 'apache',
```
ekohl: Does apache need to write here? Otherwise I'd prefer root

manifests/qpid_client.pp
```diff
@@ -39,16 +40,31 @@
...
+      group  => 'apache',
+      mode   => '0640',
+    }
+
+    file { "${certs::pulp_pki_dir}/qpid":
+      ensure => directory,
+      owner  => 'apache',
```
ekohl: Same here

manifests/qpid_client.pp
```diff
@@ -39,16 +40,31 @@
...
+      mode   => '0640',
+    } ~>
     Cert[$qpid_client_cert_name] ~>
-    key_bundle { $messaging_client_cert:
+    key_bundle { $qpid_client_cert:
       key_pair => Cert[$qpid_client_cert_name],
     } ~>
-    file { $messaging_client_cert:
+    file { $qpid_client_cert:
       owner => 'apache',
```
ekohl: Guess the same here

ekohl:
Does this now replace https://github.com/theforeman/puppet-certs/pull/148?

ehelms:
Updated ownerships

manifests/qpid_client.pp
```diff
@@ -39,16 +40,31 @@
...
   if $deploy {
 
+    file { $certs::pulp_pki_dir:
+      ensure => directory,
+      owner  => 'root',
+      group  => 'apache',
+      mode   => '0640',
+    }
+
+    file { "${certs::pulp_pki_dir}/qpid":
```
ekohl: Not 100% happy with the assumption that `$qpid_client_cert` and `$qpid_client_ca_cert` are not changed. Should we use a flat directory and name the files `qpid-client.crt` and `qpid-ca.crt`?

ehelms: We can go that route. I based this on the defaults and documentation from `pulp` itself:

https://github.com/pulp/pulp/blob/2.18-release/server/etc/pulp/server.conf#L309-L317

ekohl:
Could you fix the merge conflicts? After that I think this is ready for merge.

ehelms:
Rebased

ehelms:
@ekohl Do you want to target getting this into nightly now that we've branched?

manifests/qpid_client.pp
```diff
@@ -39,16 +40,31 @@
 
   if $deploy {
 
+    file { $certs::pulp_pki_dir:
+      ensure => directory,
+      owner  => 'root',
+      group  => 'apache',
```
ekohl: Could you add a class parameter for this? it's fine to hardcode the default but I've found that it can be very useful to not code these values inside the body.

ehelms:
@ekohl Added `cert_group` parameter, does that cover what you are asking?

PR #230, opened 2018-11-26
Check the file is present in tar_extract

ekohl:


ekohl:
FYI: I intend to use rspec mocks in all modules and will modulesync that.

manifests/tar_extract.pp
```diff
@@ -8,10 +8,9 @@
 # Actions:
 # - Extracts a tarball
 #
-# Requires:
-# - The certs class
-#
 define certs::tar_extract($path = $title) {
+  validate_file_exists($path)
```
chris1984: Is there a way we can add a message to say, the file does not exist here, and let them know how to regen a certs tar?

ekohl:
I'm leaning to merging this as it is now. Currently it's still a redundant check and the docs are in the installer.

manifests/tar_extract.pp
chris1984: +1 was just curious, I don't consider this a blocker.

PR #231, opened 2018-11-27
Move to using root for config files

ehelms:


ehelms:
This was inspired by failures seen when making this change https://github.com/theforeman/puppet-katello/pull/268

ehelms:
I am testing this further. Installation passed but running tests to check for other issues.

ekohl:
The Foreman user does need to read the certs - including the private - for the console proxy process. Would that break here or do we copy the certs somewhere where the foreman user can read them?

ehelms:
The client certificates used by Foreman, as well as the CA cert are all put into `/etc/foreman` (https://github.com/theforeman/puppet-certs/blob/master/manifests/foreman.pp).

ehelms:
I am now wondering if you meant things like websockets which would be broken by this. Is there a way to ensure the specified group is created and/or exists?

ehelms:
Stream of thought, the alternative would be to move the websockets certificates (or a copy of them) to a more natural Foreman location. Right now the websockets certificates and key are the same as the Apache  certificates.

ekohl:
I was talking about the foreman server certs. When spawning websockets proxy it uses those. https://github.com/theforeman/puppet-foreman/blob/acdaea2739a3d71c84b58f95af2e5924ae1b9841/manifests/params.pp#L195-L196 & https://github.com/theforeman/puppet-foreman/blob/acdaea2739a3d71c84b58f95af2e5924ae1b9841/templates/settings.yaml.erb#L22-L25 & https://github.com/theforeman/foreman/blob/e40dee592777e710279c7e6c2cb56422789ac1d4/lib/ws_proxy.rb#L44-L50

ekohl:
And I do agree that moving them to some place else makes sense. The whole `/etc/pki` directory containing the certs is nice if all services are root but it breaks down in reality so you're back to their own directories in `/etc/<service>`.

ehelms:
@ekohl Is there a clean way to also ensure that the group exists? since that still has to be set on the certs even when they go into `/etc/foreman` ?

PR #236, opened 2018-12-26
Use FQDNs data types where appropriate

ekohl:
This also moves back to the global facts as parameter defaults because it looks like kafo has issues with extracting the proper default.


ehelms:
Odd error, maybe related to this release of bundler 2.0?

ekohl:
Yes, this looks like a bundler 2.0 issue.

ehelms:
ACK from a runtime perspective once unit tests are passing

PR #240, opened 2019-01-17
Fixes #25873 - Set rhsm package upload options

parthaa:
Subscription manager adds the ability to turn on and off package  uploads  on a DNF transaction.
This is accomplished by setting the "package_profile_on_trans" flag in /etc/rhsm/rhsm.conf

This commit accomplishes that.

### DIFF
--- a/templates/rhsm-katello-reconfigure.erb
+++ b/templates/rhsm-katello-reconfigure.erb
@@ -83,6 +83,13 @@ else
       --server.port="$PORT" \
       --rhsm.repo_ca_cert="%(ca_cert_dir)s$KATELLO_SERVER_CA_CERT" \
       --rhsm.baseurl="$BASEURL"
+
+    # Older version of subscription manager may not recognize
+    # report_package_profile and package_profile_on_trans options.
+    # So set them separately and redirect out & error to /dev/null
+    # to fail silently.
+    subscription-manager config --rhsm.package_profile_on_trans=1 > /dev/null 2>&1 || true
+    subscription-manager config --rhsm.report_package_profile=1 > /dev/null 2>&1 || true
   else
     sed -i "s/^hostname\s*=.*/hostname = $KATELLO_SERVER/g" $CFG
     sed -i "s/^port\s*=.*/port = $PORT/g" $CFG
--- a/templates/rhsm-katello-reconfigure.erb
+++ b/templates/rhsm-katello-reconfigure.erb
@@ -84,7 +84,7 @@ else
       --rhsm.repo_ca_cert="%(ca_cert_dir)s$KATELLO_SERVER_CA_CERT" \
       --rhsm.baseurl="$BASEURL"
 
-    # Older version of subscription manager may not recognize
+    # Older versions of subscription manager may not recognize
     # report_package_profile and package_profile_on_trans options.
     # So set them separately and redirect out & error to /dev/null
     # to fail silently.


templates/rhsm-katello-reconfigure.erb
```diff
@@ -83,6 +83,8 @@ else
       --server.port="$PORT" \
       --rhsm.repo_ca_cert="%(ca_cert_dir)s$KATELLO_SERVER_CA_CERT" \
       --rhsm.baseurl="$BASEURL"
+    subscription-manager config --rhsm.package_profile_on_trans=1 &> /dev/null
```
beav: is the `&` needed? i think it would be OK to just let it run and complete without backgrounding it

parthaa: So older versions of subscription manager will not support these options. Moreover subscription manager errors out if it finds a config entry that doesn't exist. For example
```
# subscription-manager config --rhsm.FOO=1 
Usage: subscription-manager config [OPTIONS]

subscription-manager: error: no such option: --rhsm.FOO
```
I do not want the entire script to fail if 'rhsm.package_profile_on_trans' is not supported.
The best way I saw to accomplish this is to do the  ```&> /dev/null`` so that it continues along on failure.

beav: @parthaa that sounds good to me. Can you add a comment above saying its OK if the command fails? that way people wont get confused if they are debugging it later on.

parthaa: Updated

templates/rhsm-katello-reconfigure.erb
```diff
@@ -83,6 +83,13 @@ else
       --server.port="$PORT" \
       --rhsm.repo_ca_cert="%(ca_cert_dir)s$KATELLO_SERVER_CA_CERT" \
       --rhsm.baseurl="$BASEURL"
+
+    # Older version of subscription manager might not recogniza
```
beav: typo

templates/rhsm-katello-reconfigure.erb
```diff
@@ -83,6 +83,13 @@ else
...
+    # report_package_profile and package_profile_on_trans options
+    # set them separately and redirect out & error to /dev/null
+    # to fail silently
+    subscription-manager config --rhsm.package_profile_on_trans=1 &> /dev/null
```
beav: you may want to do 2&>1 to send all output (both stderr and stdout) to /dev/null

templates/rhsm-katello-reconfigure.erb
```diff
@@ -83,6 +83,13 @@ else
       --server.port="$PORT" \
       --rhsm.repo_ca_cert="%(ca_cert_dir)s$KATELLO_SERVER_CA_CERT" \
       --rhsm.baseurl="$BASEURL"
+
+    # Older version of subscription manager may not recognize
+    # report_package_profile and package_profile_on_trans options.
+    # So set them separately and redirect out & error to /dev/null
+    # to fail silently.
+    subscription-manager config --rhsm.package_profile_on_trans=1 > /dev/null 2>&1
```
ekohl: IIRC this script is running with `-e` so it'll exit if it fails. You probably want to use `|| true` if it's allowed to fail

parthaa: yes it has `-e`. Good catch

parthaa: updated

templates/rhsm-katello-reconfigure.erb
```diff
@@ -83,6 +83,13 @@ else
       --server.port="$PORT" \
       --rhsm.repo_ca_cert="%(ca_cert_dir)s$KATELLO_SERVER_CA_CERT" \
       --rhsm.baseurl="$BASEURL"
+
+    # Older version of subscription manager may not recognize
```
chris1984: version should be versions

chris1984: 1 nitpick from me but other than that ack.

parthaa: updated

beav:
@parthaa I didnt test this, but no additional comment from me. looks good :eyeglasses:

jturel:
gonna give this a test.

jturel:
Tested that these commands fail silently where they should and affect the subman config in the right way when they should

PR #243, opened 2019-02-20
Fixes #26088 - ensure RSA word for SSLProxyMachineCertificateFile

iNecas:
Apaches's SSLProxyMachineCertificateFile is not able to find a key
wrapped in `-----BEGIN PRIVATE KEY-----`, which is the format that is
generated when running the server in FIPS mode. It seems that just making
sure the RSA word is there make Apache happy to find the key.

I've added `force_rsa` option for the cert files resources to enforce the
RSA word to be there when wrapping the private key + used that to
generate the client bundle for SSL proxy to use it.

### DIFF
--- a/lib/puppet/provider/katello_ssl_tool.rb
+++ b/lib/puppet/provider/katello_ssl_tool.rb
@@ -175,14 +175,22 @@ module Puppet::Provider::KatelloSslTool
 
     def exists?
       return false unless File.exists?(resource[:path])
-      checksum(expected_content) == checksum(current_content)
+      checksum(expected_content_processed) == checksum(current_content)
     end
 
     def create
-      File.open(resource[:path], "w", mode) { |f| f << expected_content }
+      File.open(resource[:path], "w", mode) { |f| f << expected_content_processed }
     end
 
     protected
+
+    def expected_content_processed
+      content = expected_content
+      if resource[:force_rsa]
+        content.gsub!(/(BEGIN|END) (PRIVATE KEY)/, '\1 RSA \2')
+      end
+      content
+    end
 
     def expected_content
       File.read(source_path)
--- a/lib/puppet/type/certs_common.rb
+++ b/lib/puppet/type/certs_common.rb
@@ -63,6 +63,9 @@ module Certs
 
     newparam(:password_file)
 
+    # ensure RSA string is present in -----(BEGIN/END) (RSA )?PRIVATE KEY-----
+    newparam(:force_rsa)
+
     # make ensure present default
     define_method(:managed?) { true }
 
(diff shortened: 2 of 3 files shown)


iNecas:
Test failures unrelated

ehelms:
I'm good with this change to address the issue. I've not had a chance to test it myself yet.

lib/puppet/provider/katello_ssl_tool.rb
```diff
@@ -175,15 +175,23 @@ class CertFile < Puppet::Provider
 
     def exists?
       return false unless File.exists?(resource[:path])
-      checksum(expected_content) == checksum(current_content)
+      checksum(expected_content_processed) == checksum(current_content)
```
ekohl: This checksum does an `md5()`. Isn't that a problem in itself for FIPS?

iNecas: Good catch. So apparently puppet by default vendors it's own openssl, so not using the RHEL fips one. As a consequence, the puppet's ruby can still use FIPS non-compliant algorithms (see https://tickets.puppetlabs.com/browse/PUP-7510)

To compare, running on FIPS enabled machine:

```
$ ruby -e 'require "digest/md5"; puts Digest::MD5.digest("eat this")'
md5_dgst.c(82): OpenSSL internal error, assertion failed: Digest MD5 forbidden in FIPS mode!
Aborted
$ /opt/puppetlabs/puppet/bin/ruby -e 'require "digest/md5"; puts Digest::MD5.digest("eat this")'
}T+5��m�$��y�=
```

They have started building a version without bundling openssl for this particular reason (https://github.com/puppetlabs/puppet-agent/commit/7001813f0d8a736f7e8f21a60151ea705f865aff), and in puppet enterprise, there is FIPS listed as separate architecture (https://puppet.com/docs/pe/2017.3/supported_operating_systems.html#puppet-agent-platforms). However, it doesn't look like this builds being available in the community version.

I will open separate PR to replace the md5 here, but it will require additional testing with the puppet's FIPS-enabled version, as their might be dragons hidden in other parts of our installer modules.

pondrejk:
Did some sanity testing with this patch applied on downstream capsule on fips-enabled rhel7, installation, services, host registration and content sync worked as expected

lib/puppet/provider/katello_ssl_tool.rb
ekohl: I'm not sure we actually need to compare checksums. Can't ruby just compare strings? But I'm fine with doing that in a separate patch

iNecas: here you go https://github.com/theforeman/puppet-certs/pull/246

ekohl:
Merging since @ehelms had no objections and @pondrejk tested it.

PR #244, opened 2019-02-20
Add nssdb files for EL8

ehelms:


manifests/ssltools/nssdb.pp
```diff
@@ -6,7 +6,15 @@
   Exec { logoutput => 'on_failure' }
 
   $nss_db_password_file   = "${nss_db_dir}/nss_db_password-file"
-  $nssdb_files            = ["${nss_db_dir}/cert8.db", "${nss_db_dir}/key3.db", "${nss_db_dir}/secmod.db"]
+
+  case $operatingsystemrelease {
```
ekohl: You could also use `$operatingsystemmajrelease` which is only the major version. Nowadays we should also use `$facts['operatingsystemmajrelease']`.

ehelms: Updated

manifests/ssltools/nssdb.pp
```diff
@@ -6,7 +6,15 @@
   Exec { logoutput => 'on_failure' }
 
   $nss_db_password_file   = "${nss_db_dir}/nss_db_password-file"
-  $nssdb_files            = ["${nss_db_dir}/cert8.db", "${nss_db_dir}/key3.db", "${nss_db_dir}/secmod.db"]
+
+  case $facts['operatingsystemmajrelease'] {
```
ekohl: I think I agree with @evgeni that we should be future compatible. How about:
```puppet
$nssdb_files = $facts['operatingsystemmajrelease'] ? {
  '7'     => ["${nss_db_dir}/cert8.db", "${nss_db_dir}/key3.db", "${nss_db_dir}/secmod.db"],
  default => ["${nss_db_dir}/cert9.db", "${nss_db_dir}/key4.db", "${nss_db_dir}/pkcs11.txt"],
}
```

ehelms:
Updated to change the default

PR #245, opened 2019-02-20
Add CA cert to keystore as a trustedcert

ehelms:


ekohl:
https://github.com/theforeman/puppet-certs/pull/173 should do this as well. Shall we revisit that?

ehelms:
#173 Does not actually do what this does. I'm happy to re-visit that if you are signing on to keep that PR up to date as I test and updates are needed. 

This part is also needed -- https://github.com/theforeman/puppet-certs/pull/245/files#diff-0a129cc5a540f220caade2da4fc3229bR113

manifests/candlepin.pp
```diff
@@ -122,7 +128,11 @@
       mode   => '0750',
     } ~>
     exec { 'import CA into Candlepin truststore':
-      command => "keytool -import -v -keystore ${amqp_truststore} -storepass ${keystore_password} -alias ${alias} -file ${ca_cert} -noprompt",
+      command => "keytool -import -trustcacerts -v -keystore ${keystore} -storepass ${keystore_password} -alias ${alias} -file ${ca_cert} -noprompt",
+      unless  => "keytool -list -keystore ${keystore} -storepass ${keystore_password} -alias ${alias}",
+    } ~>
+    exec { 'import CA into Candlepin truststore':
```
ekohl: This should be the Candlepin AMQP truststore so the name is unique

ehelms:
Updated and passing tests

manifests/candlepin.pp
```diff
@@ -122,7 +128,11 @@
       mode   => '0750',
     } ~>
     exec { 'import CA into Candlepin truststore':
-      command => "keytool -import -v -keystore ${amqp_truststore} -storepass ${keystore_password} -alias ${alias} -file ${ca_cert} -noprompt",
+      command => "keytool -import -trustcacerts -v -keystore ${keystore} -storepass ${keystore_password} -alias ${alias} -file ${ca_cert} -noprompt",
+      unless  => "keytool -list -keystore ${keystore} -storepass ${keystore_password} -alias ${alias}",
+    } ~>
+    exec { 'import CA into Candlepin AMQP truststore':
+      command => "keytool -import -v -keystore ${amqp_truststore} -storepass ${keystore_password} -alias ${alias} -file ${ca_cert} -trustcacerts -noprompt",
       creates => $amqp_truststore,
```
ekohl: It wasn't here before, but maybe this also needs an unless instead of creates so it's also updated. I like your new approach.

PR #267, opened 2019-12-19
Fixes #25683 - Do not run postun script if upgrading

jlsherrill:


templates/postun.erb
```diff
@@ -0,0 +1,3 @@
+if [ $1 -eq 0 ]; then
```
ehelms: RPM post uninstall passes arguments?

manifests/katello.pp
```diff
@@ -55,7 +55,7 @@
     # (especially in the custom certs scenario)
     files            => ["${katello_rhsm_setup_script_location}:755=${katello_www_pub_dir}/${katello_rhsm_setup_script}"],
     bootstrap_script => "/bin/bash ${katello_rhsm_setup_script_location}",
-    postun_script    => 'test -f /etc/rhsm/rhsm.conf.kat-backup && command cp /etc/rhsm/rhsm.conf.kat-backup /etc/rhsm/rhsm.conf',
+    postun_script    => template('certs/postun.erb'),
```
ekohl: If there's no template I'd prefer this to be a static file. You'll need to move it to `files/postun.sh` as well.
```suggestion
    postun_script    => file('certs/postun.sh'),
```

templates/postun.erb
ekohl: Yes. https://docs.fedoraproject.org/en-US/packaging-guidelines/Scriptlets/#_syntax

jlsherrill: was news to me too :)

manifests/katello.pp
jlsherrill: Learned something new!  Done!

ekohl:
I believe this fixes https://projects.theforeman.org/issues/25683. Should be move it to the installer project and properly use `Fixes #25683` in the commit message here?

jlsherrill:
updated issue association

ekohl:
Thanks!

PR #270, opened 2020-01-13
Ensure pulp-server package present before qpid client certs created

ehelms:


ehelms:
Devel installs suffer from this error:

```
[ERROR 2020-01-13T19:38:46 main]  Could not set 'directory' on ensure: Could not find group pulp (file: /usr/share/foreman-installer/modules/certs/manifests/qpid_client.pp, line: 45)
```

ekohl:
Should we fix it [here](https://github.com/theforeman/puppet-katello_devel/blob/a9d0ba1889a9147197696ffc2efa6d8f336f278b/manifests/init.pp#L181-L206)? This class has no knowledge of the pulp-server package and thus will fail to compile with all dependencies.

ehelms:
Maybe? I am not always sure where something should live. This class for example won't work *unless* the pulp group exists. Is ensuring that exists the purview of this puppet class? or somebody using this class? I think that production likely gets lucky that this works right now.

ehelms:
Just the edits, as this current idea doesnt work:

```
    [ERROR 2020-01-13T20:11:50 verbose] (Cert[candlepin-migration.war.example.com-qpid-client-cert] => Class[Certs::Qpid_client] => Class[Pulp] => Class[Pulp::Install] => Package[pulp-server] => File[/etc/pki/pulp] => Class[Certs::Qpid_client])\nTry the '--graph' option and opening the resulting '.dot' file in OmniGraffle or GraphViz
    [ERROR 2020-01-13T20:11:50 verbose]  Failed to apply catalog: One or more resource dependency cycles detected in graph
```

ekohl:
Regarding design: most modules take care of placing their own files to be able to properly set the dependencies. In puppet-pulpcore we define the pulp user and group so Puppet would automatically find this. That means we can't define them again in puppet-pulp.

ehelms:
The virtual resource idea failed with:

```
    [ERROR 2020-01-13T20:29:55 verbose]  Could not find group pulp
```

This is likely due to nothing actually declaring this resource in the devel scenario yet?

manifests/qpid_client.pp
```diff
@@ -42,27 +42,29 @@
 
   if $deploy {
 
+    Package <| name == 'pulp-server' |> -> File <| group == 'pulp' |>
```
ekohl: Alternative suggestion:
```suggestion
    Package <| name == 'pulp-server' |> -> Class['qpid_client']
```

ehelms:
This appears to be the crux of the problem (https://github.com/theforeman/puppet-katello/blob/master/manifests/pulp.pp#L127). We need to generate the messaging certificates before setting up Pulp, but those certificates need pulp-server package in order to have the user and group.

Is there any puppet magic to help?

manifests/qpid_client.pp
```diff
@@ -42,27 +42,29 @@
 
   if $deploy {
 
+    Package <| name == 'pulp-server' |> -> Class['qpid_client']
```
ekohl: Looks like I wasn't fully awake when I made the suggestion:
```suggestion
    Package <| name == 'pulp-server' |> -> Class['certs::qpid_client']
```

ehelms:
I believe this has been fixed by other updates and PRs. Thanks @ekohl !

PR #273, opened 2020-02-04
Refs #28922: Generate tomcat certificates for localhost

ehelms:


ekohl:
Looks like the tests are unhappy.

I think we also already do this for qpid, but generating certs for localhost is not a very good practice due to security concerns. Let's discuss this in the 2.1 timeframe.

ehelms:
There is more context on the associated PR -- https://github.com/theforeman/puppet-candlepin/pull/142. Let's discuss overall deployment strategy there.

manifests/candlepin.pp
```diff
@@ -1,6 +1,6 @@
 # Constains certs specific configurations for candlepin
 class certs::candlepin (
-  $hostname               = $certs::node_fqdn,
+  $hostname               = 'localhost',
```
jturel: Should this reflect the value of $candlepin::params::host or whatever value was giving to puppet-candlepin?

ehelms: That's a fair question. We'd handle this by whatever calls this class passing this value in. That could mean we should change (https://github.com/theforeman/puppet-katello/blob/master/manifests/candlepin.pp#L31) to be:

```
class { 'certs::candlepin':
  hostname => $candlepin::host
}
```

@ekohl am I thinking correctly here or is there some other method we'd expect?

jturel: I was asking in reference to a comment on the related PR: 

https://github.com/theforeman/puppet-candlepin/pull/142#issuecomment-587504985

ekohl: These are the defaults, but I do feel you're on the right path. IMHO [katello::candlepin](https://github.com/theforeman/puppet-katello/blob/master/manifests/candlepin.pp) is the right place to connect these two. There we can do something like:
```puppet
include certs::candlepin
class { 'candlepin':
  host => $certs::candlepin::hostname,
}
```

Now the end users' experience will be that they can use `--certs-candlepin-hostname` to set the hostname. The alternative is to add a `$hostname` parameter to `katello::candlepin`. The benefit to that is that it feels a bit more natural for the end user but we need to look at the configuration of the various certs classes and how to expose it. In our team we've been talking at evaluating the certs usage in the installer.

ekohl:
I don't think this change is really needed given we'll set this parameter in puppet-katello.

Another reason not to have this default is that a localhost certificate is IMHO a bad practice since it's undefined what the host actually is.

ehelms:
These questions are for my own education:

 1) Can you expand on the part about what the host actually is? Are you implying that if using localhost you shouldn't use SSL?
 2) Is there other ways than using localhost to expose a service without exposing it publically?
 3) Given this module is isolated to Katello's scenario, I assumed it made sense to have the defaults align with the default deployment. That is more of a statement on why I added this PR then a question.

ekohl:
> Can you expand on the part about what the host actually is? Are you implying that if using localhost you shouldn't use SSL?

I would consider that. In favor of using SSL:
* Easier configuration on the Katello side
Against:
* Generation of certificates is more complex
* Additional overhead of SSL at runtime

Does Katello use client side certificates?

> Is there other ways than using localhost to expose a service without exposing it publically?

Technically we always expose it on port 443 so it feels like we're doing something too complex:

```
candlepin (https://localhost:8443) <- apache (https://katello.example.com) <- apache (https://proxy.example.com)
```

It feels to me that candlepin should stay on its own port and (content) proxies should connect directly to candlepin instead of pushing it though another reverse proxy on the katello host.

Am I missing something?

> Given this module is isolated to Katello's scenario, I assumed it made sense to have the defaults align with the default deployment. That is more of a statement on why I added this PR then a question.

On the one hand I get that, but it means a breaking change for no additional benefit. That means there's a greater chance that cherry picking is needed. In the future I'd like to increase coverage so that puppet-katello has sufficient coverage.

ehelms:
> Technically we always expose it on port 443 so it feels like we're doing something too complex:
> 
> ```
> candlepin (https://localhost:8443) <- apache (https://katello.example.com) <- apache (https://proxy.example.com)
> ```
> 
> It feels to me that candlepin should stay on its own port and (content) proxies should connect directly to candlepin instead of pushing it though another reverse proxy on the katello host.
> 
> Am I missing something?

You are. All requests go through Katello to talk to Candlepin. Katello handles auth for incoming requests.


Today Katello does not use client certificates to talk to Candlepin, but I am pushing this conversation forward to move away from oauth.

ekohl:
Right, that piece of the puzzle I was missing.

ekohl:
So how do we proceed with this? Merging this isn't needed since puppet-katello already sets the right hostname when calling this so I'm leaning to closing the PR.

PR #275, opened 2020-03-06
Refs #28924: Drop amqp key and truststore

ehelms:


manifests/candlepin.pp
```diff
@@ -121,31 +118,15 @@
...
       # create the store as an PKCS12 and convert to JKS.  See http://stackoverflow.com/a/8224863
-      command => "openssl pkcs12 -export -name amqp-client -in ${client_cert} -inkey ${client_key} -out /tmp/keystore.p12 -passout file:${password_file} && keytool -importkeystore -destkeystore ${amqp_keystore} -srckeystore /tmp/keystore.p12 -srcstoretype pkcs12 -alias amqp-client -storepass ${keystore_password} -srcstorepass ${keystore_password} -noprompt && rm /tmp/keystore.p12",
-      unless  => "keytool -list -keystore ${amqp_keystore} -storepass ${keystore_password} -alias amqp-client",
-    } ~>
-    file { $amqp_keystore:
-      ensure => file,
-      owner  => 'tomcat',
-      group  => $group,
-      mode   => '0640',
+      command => "openssl pkcs12 -export -name artemis-client -in ${client_cert} -inkey ${client_key} -out /tmp/keystore.p12 -passout file:${password_file} && keytool -importkeystore -destkeystore ${artemis_keystore} -srckeystore /tmp/keystore.p12 -srcstoretype pkcs12 -alias artemis-client -storepass ${keystore_password} -srcstorepass ${keystore_password} -noprompt && rm /tmp/keystore.p12",
```
ekohl: `$artemis_keystore` is undefined and making the tests very unhappy

manifests/candlepin.pp
```diff
@@ -121,31 +118,15 @@
       key_file  => $client_key,
       cert_file => $client_cert,
     } ~>
-    file { $amqp_store_dir:
```
ekohl: I think we should ensure it's absent to clean up after ourselves, but I don't recall if we can ensure directories are absent in Puppet.

spec/acceptance/candlepin_spec.rb
```diff
@@ -21,7 +21,7 @@
         ensure => present,
       }
 
-      ['/usr/share/tomcat/conf', '/etc/candlepin/certs/amqp'].each |$dir| {
+      ['/usr/share/tomcat/conf'].each |$dir| {
```
ekohl: I think we can simplify this a bit since a loop with a single item isn't all that useful :)

jturel:
Seeing this:

```
[ WARN 2020-03-09T18:11:46 verbose]  Unknown variable: 'artemis_keystore'. (file: /usr/share/foreman-installer/modules/certs/manifests/candlepin.
pp, line: 128, column: 200)
```

So the subsequent keytool command fails since -destkeystore is missing

jturel:
@ehelms just pinging to ensure you saw my last update :ping_pong:

manifests/candlepin.pp
ehelms: I wondered about that. Should the puppet module clean this up or installer post install task clean this up?

spec/acceptance/candlepin_spec.rb
```diff
@@ -91,6 +89,7 @@
       its(:stdout) { should match(/^Your keystore contains 2 entries$/) }
       its(:stdout) { should match(/^tomcat, .+, PrivateKeyEntry, $/) }
       its(:stdout) { should match(/^candlepin-ca, .+, trustedCertEntry, $/) }
+<<<<<<< HEAD
```
ekohl:
```suggestion
```

manifests/candlepin.pp
ekohl: Normally I'd say Puppet but if Puppet can't clean up directories with the `file` type then I'm ok with a post install task rather than an exec here.

manifests/candlepin.pp
```diff
@@ -121,31 +118,15 @@
...
       # create the store as an PKCS12 and convert to JKS.  See http://stackoverflow.com/a/8224863
-      command => "openssl pkcs12 -export -name amqp-client -in ${client_cert} -inkey ${client_key} -out /tmp/keystore.p12 -passout file:${password_file} && keytool -importkeystore -destkeystore ${amqp_keystore} -srckeystore /tmp/keystore.p12 -srcstoretype pkcs12 -alias amqp-client -storepass ${keystore_password} -srcstorepass ${keystore_password} -noprompt && rm /tmp/keystore.p12",
-      unless  => "keytool -list -keystore ${amqp_keystore} -storepass ${keystore_password} -alias amqp-client",
-    } ~>
-    file { $amqp_keystore:
-      ensure => file,
-      owner  => 'tomcat',
-      group  => $group,
-      mode   => '0640',
+      command => "openssl pkcs12 -export -name artemis-client -in ${client_cert} -inkey ${client_key} -out /tmp/keystore.p12 -passout file:${password_file} && keytool -importkeystore -destkeystore ${keystore} -srckeystore /tmp/keystore.p12 -srcstoretype pkcs12 -alias artemis-client -storepass ${keystore_password} -srcstorepass ${keystore_password} -noprompt && rm /tmp/keystore.p12",
```
ekohl: I'd like to store `artemis-client` in a variable, even if it's a local one.

spec/acceptance/candlepin_spec.rb
```diff
@@ -10,11 +10,9 @@
         ensure => present,
       }
 
-      ['/usr/share/tomcat/conf', '/etc/candlepin/certs/amqp'].each |$dir| {
```
ekohl: Something is failing in the acceptance tests:
```
  Error: Could not set 'present' on ensure: No such file or directory @ rb_sysopen - /etc/candlepin/certs/candlepin-ca.key (file: /etc/puppetlabs/code/modules/certs/manifests/keypair.pp, line: 18)
```

Looks like the certs directory is still used for the regular files (`candlepin_certs_dir`)

spec/acceptance/candlepin_spec.rb
```diff
@@ -110,7 +110,7 @@
       its(:exit_status) { should eq 0 }
```
ekohl: This whole store doesn't exist anymore. Same with the truststore above.

manifests/candlepin.pp
ehelms: From my googling you can do a regular file set to absent with recursive set to true. This just doesn't feel as clean. That weird line where sometimes we cleanup in the puppet module and sometimes we use the puppet module to simply represent the state and the installer does cleanup. I thought we were universally doing the latter. Its hard to know which way to go.

jturel:
I did some testing and was able to connect with SSL client auth from Katello to Artemis.

I basically followed the steps [here](https://github.com/apache/activemq-artemis/blob/d8ecf24733049f0ff686beaa6ed507f87766ecde/examples/protocols/stomp/stomp-dual-authentication/readme.md), extracted the client cert+key from the stores, and used them to connect. Seems like we should be able to approximate that here, and let me know if I can provide any more detail.

ehelms:
@jturel Apologies, I'm not following what the action item is regarding server and client keystores.

jturel:
> @jturel Apologies, I'm not following what the action item is regarding server and client keystores.

I need my own clarification then :)

The `artemis_client` entry you've added to the keystore: what is the purpose of that? Is it for Katello to use? If so, I suppose we'll need puppet-katello to extract the cert and key in PEM format in order to connect to the broker.

Or is it for Candlepin to use? If so, Candlepin is connecting directly to the embedded artemis without SSL since it's running within the same JVM which means it's not necessary, unless you're thinking forward to candlepin connecting to an external Artemis

ehelms:
Ahh yes. This client certificate is being added to the keystore/truststore so that it can be used by Katello. The client certificate is already deployed to the system in a place that Katello can consume it. This change is just ensuring its in the truststore. We abuse keystore/truststore by creating one and using it for both.

/etc/pki/katello/certs/java-client.crt
/etc/pki/katello/private/java-client.key

manifests/candlepin.pp
```diff
@@ -31,6 +28,8 @@
   }
 
   $java_client_cert_name = 'java-client'
+  $artemis_alias = 'artemis-client'
+  $artemis_client_dn = "C=${city}, ST=${state}, O=candlepin, OU=${org_unit}, CN=${cname}"
```
jturel, line 32: I think the CN should be $hostname rather than $cname

ehelms: To be safe, I think this is correct. See https://github.com/theforeman/puppet-certs/pull/275/files#diff-0a129cc5a540f220caade2da4fc3229bR37

jturel: Hmm, well here's the contents of the file:

```
katelloUser=C=Raleigh, ST=North Carolina, O=candlepin, OU=SomeOrgUnit, CN=[]
```

Hard to discern what you wanted me to see at that URL..

ehelms: Ohh, yea, I see now how this gets handled.

ehelms: Updated to hostname, I forget cname is *additional* cnames to add.

ekohl: In SSL you have `Common Name` (`cn`) and `subject Alternative Name` (`subjectAltName`). `cname` is DNS terminology.

manifests/candlepin.pp
```diff
@@ -31,6 +28,8 @@
   }
 
   $java_client_cert_name = 'java-client'
+  $artemis_alias = 'artemis-client'
+  $artemis_client_dn = "C=${city}, ST=${state}, O=candlepin, OU=${org_unit}, CN=${hostname}"
```
jturel, line 32: It actually seems like the order of these matters.... (most specific -> least specific?)

Should be like this: `CN=localhost, OU=SomeOrgUnit, O=candlepin, ST=North Carolina, C=US
`

Note also that C is country, not city

ehelms: Ahh yea, sorry you are right on the country.

I modeled this off of "C=US, ST=North Carolina, O=candlepin, OU=SomeOrgUnit, CN=dhcp-8-29-228.lab.eng.rdu2.redhat.com"

```
openssl x509 -noout -text -in /etc/pki/katello/certs/java-client.crt
```

jturel: All good. When I re-order them in login.config it prevents Katello from connecting with an auth failure :shrug:

PR #283, opened 2020-05-04
Modulesync: Require voxpupuli-test 1.4

ekohl:


mmoll:
I guess the fixture setup is broken on integration tests now...

ekohl:
I think the module is built now and doesn't include fixtures.

PR #284, opened 2020-05-07
Refs #29215: Add RHEL 8 to metadata and run tests on all OSes

ehelms:


spec/classes/certs_apache_spec.rb
```diff
@@ -1,24 +1,28 @@
 require 'spec_helper'
 
 describe 'certs::apache' do
-  let :facts do
-    on_supported_os['redhat-7-x86_64']
-  end
+  on_supported_os.each do |os, facts|
+    context "on #{os}" do
+      let :facts do
+        facts
```
ekohl, line 7: This is the old pattern that we still have all over the place. Recently the best practice has changed to:
```suggestion
  on_supported_os.each do |os, os_facts|
    context "on #{os}" do
      let(:facts) do
        os_facts
```

The benefit is that you don't have overlap between the variable and let block naming which makes things less confusing.

spec/classes/certs_qpid_router_spec.rb
```diff
@@ -1,11 +1,13 @@
 require 'spec_helper'
 
 describe 'certs::qpid_router' do
-  let :facts do
-    on_supported_os['redhat-7-x86_64']
-  end
+  on_supported_os.each do |os, facts|
+    context "on #{os}", if: facts[:operatingsystemmajrelease] == '7' do
+      let :facts do facts end
```
ekohl, line 6: Slightly easier to read:
```suggestion
      let(:facts) { facts }
```

spec/classes/certs_puppet_spec.rb
```diff
@@ -1,11 +1,19 @@
...
-  end
+  on_supported_os.each do |os, os_facts|
+    context "on #{os}" do
+      let :facts do
+        os_facts
+      end
+
+      let :facts do
+        on_supported_os['redhat-7-x86_64']
+      end
```
ekohl, line 12:
```suggestion
```

PR #285, opened 2020-05-11
Refs #29715: Add mongodb server and client certs

ehelms:


manifests/mongodb.pp
```diff
@@ -0,0 +1,105 @@
...
+
+  if $deploy {
+    file { "${pki_dir}/mongodb":
+      ensure => directory,
+      mode   => '0750',
+      owner  => 'mongodb',
+      group  => 'mongodb',
+    }
+
+    file { '/etc/pulp/mongodb':
```
ekohl, line 73: Just to show the relation, I'd store this in a variable (`$mongodb_client_cert_dir`) and use it in `$mongodb_client_{cert,key}` as well. It doesn't have to be a class parameter, but I think it might even make sense.

manifests/mongodb.pp
```diff
@@ -0,0 +1,105 @@
...
+    regenerate    => $regenerate,
+    deploy        => $deploy,
+    password_file => $ca_key_password_file,
+  }
+
+  if $deploy {
+    file { "${pki_dir}/mongodb":
+      ensure => directory,
+      mode   => '0750',
+      owner  => 'mongodb',
```
ekohl, line 69: Please introduce a variable for this user.

manifests/mongodb.pp
```diff
@@ -0,0 +1,105 @@
...
+      group  => 'mongodb',
+    }
+
+    file { '/etc/pulp/mongodb':
+      ensure => directory,
+      owner  => 'root',
+      group  => 'pulp',
```
ekohl, line 76: Please introduce a variable for this

ehelms:
I do appreciate the idea and power behind splitting, this class is intended for a rather specific purpose and will be short lived as Pulp 2 is not long left of life in our ecosystem. I prefer to keep this is together and simple.

manifests/mongodb.pp
ehelms: I might still make that change, I can't get these tests to run locally so some of the code pushes you see are just to use Travis for acceptance testing.

ekohl:
My reasoning is that we will need them in Pulpcore to import. There the parameters (`/etc/pulpcore/mongodb` + `pulpcore` user) will make sense. In that case there will not be a server.

spec/setup_acceptance_node.pp
```diff
@@ -8,3 +8,17 @@
...
+user { 'mongodb':
+  ensure => present
+}
+
+user { 'pulp':
+  ensure => present
+}
+
+file { '/etc/pulp':
+  ensure => present,
```
ekohl, line 21:
```suggestion
  ensure => directory,
```

ehelms:
Tis green now and split into two classes.

ehelms:
@ekohl Ready for re-review please

manifests/mongodb_client.pp
```diff
@@ -0,0 +1,67 @@
...
+  $server_cert          = $certs::server_cert,
+  $country              = $certs::country,
+  $state                = $certs::state,
+  $city                 = $certs::city,
+  $org                  = $certs::org,
+  $org_unit             = $certs::org_unit,
+  $expiration           = $certs::expiration,
+  $default_ca           = $certs::default_ca,
+  $ca_key_password_file = $certs::ca_key_password_file,
+  $group                = $certs::group,
```
ekohl, line 18: Should this be
```suggestion
  $group                = 'pulp',
```

And then use that instead of `$pulp_user`? The variable `$group` is not used elsewhere.

manifests/mongodb_client.pp
```diff
@@ -0,0 +1,67 @@
...
+    regenerate    => $regenerate,
+    deploy        => $deploy,
+    password_file => $ca_key_password_file,
+  }
+
+  if $deploy {
+    file { $pki_dir:
+      ensure => directory,
+      owner  => 'root',
+      group  => 'pulp',
```
ekohl, line 50: Missed one here

manifests/mongodb.pp
```diff
@@ -0,0 +1,62 @@
...
+    regenerate    => $regenerate,
+    deploy        => $deploy,
+    password_file => $ca_key_password_file,
+  }
+
+  if $deploy {
+    file { "${pki_dir}/mongodb":
+      ensure => directory,
+      mode   => '0750',
+      owner  => $mongodb_user,
```
ekohl, line 47: Normally I'd say this should be root since mongodb shouldn't alter it (same below for the bundle)
```suggestion
      owner  => 'root',
```

manifests/mongodb.pp
```diff
@@ -0,0 +1,62 @@
...
+  $server_cert          = $certs::server_cert,
+  $country              = $certs::country,
+  $state                = $certs::state,
+  $city                 = $certs::city,
+  $org                  = $certs::org,
+  $org_unit             = $certs::org_unit,
+  $expiration           = $certs::expiration,
+  $default_ca           = $certs::default_ca,
+  $ca_key_password_file = $certs::ca_key_password_file,
+  $group                = $certs::group,
```
ekohl, line 18: Should this be `mongodb` and drop `$mongodb_user`?
```suggestion
  $group                = 'mongodb',
```

PR #289, opened 2020-07-02
Fixes #30312: Drop docker, atomic and goferd support from consumer RPM

ehelms:
@jlsherrill @parthaa Could y'all confirm that these can safely be removed for Katello 4.0 and point to anything else that might could be removed? I left in the EL5 host override fact for now given EL5 EOLs in November of this year which more closely aligns with 4.1.

### DIFF
--- a/templates/rhsm-katello-reconfigure.erb
+++ b/templates/rhsm-katello-reconfigure.erb
@@ -40,21 +40,7 @@ chmod 644 $KATELLO_CERT_DIR/$KATELLO_SERVER_CA_CERT
 echo "$KATELLO_DEFAULT_CA_DATA" > $KATELLO_CERT_DIR/$KATELLO_DEFAULT_CA_CERT
 chmod 644 $KATELLO_CERT_DIR/$KATELLO_DEFAULT_CA_CERT
 
-# if atomic or debian machine handle it the special way, else handle the regular rhel way
-if [ -n "${IS_ATOMIC+1}" ] || [ -e "/run/ostree-booted" ]
-then
-  # atomic setup
-  BASEURL=https://$KATELLO_SERVER/pulp/ostree/web/
-
-  # configure rhsm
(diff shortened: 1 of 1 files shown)


jturel:
Looks sane to me

parthaa:
I am ok with the changes here. But have  a question ->  https://github.com/theforeman/puppet-certs/commit/3028e1f5c3c9e9b87840953084bfa0dbc45ac236 meant for atomic changed a bunch of things . 

- It created a new katello-rhsm-reconfigure  script (may have the wrong name)
- It required the cert to be embedded in the reconfigure script

Wonder if we should take the approach of splitting certs into a separate file since this kinda hack is no longer necessary. 
I am a bit conflicted about this though. Ok with what is there.

ehelms:
@parthaa you are sniffin around changes I plan to make as a follow up to simplify things. I want to clean up any unsupported workflows, then move creation to a different module (https://github.com/theforeman/puppet-foreman_proxy_content/pull/272) and then further re-factor to simplify the script and drop the reliance on katello-certs-tools.

parthaa:
> @parthaa you are sniffin around changes I plan to make as a follow up to simplify things. I want to clean up any unsupported workflows, then move creation to a different module ([theforeman/puppet-foreman_proxy_content#272](https://github.com/theforeman/puppet-foreman_proxy_content/pull/272)) and then further re-factor to simplify the script and drop the reliance on katello-certs-tools.

sounds good to me!

ekohl:
@ehelms I think this would be a major version change, right?

ehelms:
Yes - since this will correspond with the Pulp 2 removal and Katello 4.0 effort.

ehelms:
This change is already captured in https://github.com/theforeman/puppet-foreman_proxy_content/pull/272 as well

ekohl:
Right, labeled it as such.

PR #291, opened 2020-07-08
Refs #30346: Generate a separate truststore for Candlepin

ehelms:


ehelms:
Requires https://github.com/theforeman/puppet-candlepin/pull/160

manifests/candlepin.pp
```diff
@@ -128,15 +132,26 @@
...
+    file { $truststore_password_path:
+      ensure  => file,
+      content => $truststore_password,
+      owner   => $user,
+      group   => $group,
+      mode    => '0440',
+    } ~>
+    exec { 'Create Candlepin truststore with CA':
+      command => "keytool -import -v -keystore ${truststore} -alias ${alias} -file ${ca_cert} -noprompt -storetype pkcs12 -storepass ${truststore_password}",
+      unless  => "keytool -list -keystore ${truststore} -alias ${alias} -storepass ${truststore_password}",
```
ekohl, line 144: This was already a problem before, but passing it in as a command line option means it's visible in `ps aux`. Can keytool read `$truststore_password_path` like openssl does with `file:<path>`?

manifests/candlepin.pp
```diff
@@ -128,15 +132,26 @@
...
+    } ~>
+    file { $truststore:
+      ensure => file,
+      owner  => 'tomcat',
```
ekohl, line 148: Why does the user need to own and modify this? Isn't reading sufficient? If it does, should it use `$user`?

spec/acceptance/candlepin_spec.rb
```diff
@@ -88,16 +103,29 @@
...
       its(:stdout) { should match(/^Owner: CN=#{host_inventory['fqdn']}, OU=SomeOrgUnit, O=Katello, L=Raleigh, ST=North Carolina, C=US$/) }
       its(:stdout) { should match(/^Issuer: CN=#{host_inventory['fqdn']}, OU=SomeOrgUnit, O=Katello, L=Raleigh, ST=North Carolina, C=US$/) }
     end
+
+    describe command("keytool -list -keystore /etc/candlepin/certs/truststore -storepass $(cat #{truststore_password_file})") do
+      its(:exit_status) { should eq 0 }
+      its(:stdout) { should match(/^Keystore type: PKCS12$/i) }
+      its(:stdout) { should match(/^Your keystore contains 2 entries$/) }
+      its(:stdout) { should match(/^candlepin-ca, .+, trustedCertEntry, $/) }
+      its(:stdout) { should match(/^artemis-client, .+, PrivateKeyEntry, $/) }
```
ekohl, line 121: Looks like this is not a private key but a trusted cert
```suggestion
      its(:stdout) { should match(/^artemis-client, .+, trustedCertEntry, $/) }
```

manifests/candlepin.pp
ehelms: There is and I fixed that everywhere

PR #293, opened 2020-07-10
Refs #30320: Do not manage certificate files for Candlepin

ehelms:
This moves management of the files to puppet-candlepin and focuses
just on creation of the certificates, see:

  https://github.com/theforeman/puppet-candlepin/pull/159

### DIFF
--- a/manifests/candlepin.pp
+++ b/manifests/candlepin.pp
@@ -82,17 +82,9 @@ class certs::candlepin (
 
   if $deploy {
     certs::keypair { 'candlepin-ca':
-      manage_cert   => true,
-      manage_key    => true,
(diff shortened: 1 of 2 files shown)


manifests/candlepin.pp
```diff
@@ -106,19 +98,13 @@
       ensure  => file,
       content => $keystore_password,
       owner   => 'root',
-      group   => $group,
-      mode    => '0440',
+      group   => 'root',
```
ekohl, line 101: Is this file only temporary for the Puppet commands?

ehelms: Effectively yes, and for debugging the keystore/truststore.

PR #296, opened 2020-08-04
Release 9.0.0

wbclark:


CHANGELOG.md
```diff
@@ -1,6 +1,22 @@
 # Changelog
 
-## [8.0.0](https://github.com/theforeman/puppet-certs/tree/8.0.0) (2020-05-13)
+## [9.0.0](https://github.com/theforeman/puppet-certs/tree/9.0.0) (2020-08-04)
+
+[Full Changelog](https://github.com/theforeman/puppet-certs/compare/8.0.0...9.0.0)
+
+**Breaking changes:**
+
+- Refs [\#30346](https://projects.theforeman.org/issues/30346): Generate a separate truststore for Candlepin [\#291](https://github.com/theforeman/puppet-certs/pull/291) ([ehelms](https://github.com/ehelms))
```
wbclark, line 9: I wasn't sure whether to label this one Breaking or Enhancement. I went with Breaking Change here because of the ownership changes but perhaps it is really more of an Enhancement.

ekohl: I think breaking is good since it changes the behavior.

ehelms: I think this is technically an enhancement.

wbclark:
> Looks good, but do you mind updating the date to today? I do that by hand rather than rerunning the generator. Doing it now prevents a change log with 8.0.0 where it was generated one day and then merged + released the next.

It is done. Thanks!

CHANGELOG.md
ekohl: I consider it breaking because you can no longer use the two modules as you used to. That means as a consumer of the module you need to make changes. Breaking changes can be enhancements/improvements but it's more about signalling to the user.

I'll agree it's sometimes a bit arbitrary, but in this case there are already other breaking changes and not a lot of people read this changelog. However, I still think it's worth it because it helps at least me when considering cherry picks or other bigger rebases in an existing release. Doesn't happen often, but it's a good habit and having it available when you do need it is always nice.

PR #300, opened 2020-08-25
Change foreman-proxy file ownership to 440

ekohl:
This changes the file owner to be root. This prevents the service for being able to modify the files. Because it still needs to read the files, the group is changed to foreman-proxy.

The public key permissions are also modified to 640. While 644 would work, it's common in PCI compliant setups to restrict this. Only foreman-proxy should read these files so it should have no impact. However, it may be used by humans who debug a service. They may notice a difference.

I'm not entirely sure the public key needs to be limited. I'd like to find a citation of the PCI requirements that dictates this.

### DIFF
--- a/manifests/foreman_proxy.pp
+++ b/manifests/foreman_proxy.pp
@@ -1,4 +1,10 @@
 # Handles Foreman Proxy cert configuration
+#
+# @param $private_key_mode
+#   The mode used on private key files (which are owned by root:$group).
+#
+# @param $public_key_mode
+#   The mode used on public key files (which are owned by root:$group).
 class certs::foreman_proxy (
   $hostname             = $certs::node_fqdn,
   $cname                = $certs::cname,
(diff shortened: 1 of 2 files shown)


manifests/foreman_proxy.pp
```diff
@@ -82,25 +88,34 @@
...
-      manage_key => true,
-      key_owner  => 'foreman-proxy',
-      key_mode   => '0400',
-      key_group  => $group,
-      cert_file  => $proxy_cert,
+      key_pair    => Cert[$proxy_cert_name],
+      key_file    => $proxy_key,
+      manage_key  => true,
+      key_owner   => 'root',
+      key_mode    => '0640',
```
ehelms, line 95: I think we should keep this at `0400` as it was previously. There is not a situation I can perceive where this should be writable and can prevent accidental debug errors where a file gets edited by someone opening it.

ekohl: I disagree. 0400 is bad because the service itself can modify the file. Remember the owner of a file can just chmod 600. By using the group, we can prevent it from doing so. I'd also be fine with 0440, but the critical change here is that we prevent foreman-proxy from changing the config.

ehelms: I vote `0440` then.

ekohl:
Updated to 0440/0444. The original reporter about the PCI issue couldn't verify it with the customer and there's little reason to keep those certs private. A parameter is provided if it's really needed.

ehelms (approved):

PR #305, opened 2020-11-04
Use shorter names in acceptance tests

ekohl:
Draft because I don't know if this is even a valid syntax.


ekohl:
This will happen as part of the larger GH actions update.

PR #306, opened 2020-12-21
Removed unused rhsm_reconfigure_script

ehelms:


ekohl (approved):
I would have expected at least a test for this, but given the mentioned template was removed previously I don't think this is backwards incompatible.

PR #308, opened 2020-12-21
Convert parameters to static defaults

ehelms:


manifests/params.pp
```diff
@@ -1,14 +1,10 @@
 # Certs Parameters
 class certs::params {
 
-  $pki_dir = '/etc/pki/katello'
```
ekohl, line 5: This removal looks accidental

manifests/params.pp
```diff
@@ -2,6 +2,7 @@
 class certs::params {
 
   $pki_dir = '/etc/pki/katello'
+  $nss_db_dir   = "${pki_dir}/nssdb"
```
ekohl, line 5: This means it's no longer possible to change `$pki_dir` for `$nss_db_dir`, which it previously was. That's a regression.

ekohl (changes requested):
Small nit about `$pki_dir` which you remove and later re-add. The `$nss_db_dir` change looks like a regression though. Overall :+1:

ehelms:
Fixed up pki_dir and put back the nss_db setting.

ekohl (approved):

PR #311, opened 2021-01-06
Fixes #31574: Ensure truststore certificates get updated when they ch…

ehelms:
…ange

### DIFF
--- a/lib/puppet/provider/truststore_certificate/truststore_certificate.rb
+++ b/lib/puppet/provider/truststore_certificate/truststore_certificate.rb
@@ -1,0 +1,81 @@
+Puppet::Type.type(:truststore_certificate).provide(:truststore_certificate) do
+
+  commands :keytool => 'keytool'
+  commands :openssl => 'openssl'
+
+  def create
+    add_certificate unless certificate_present?
+    if certificate_changed?
+      delete_certificate
+      add_certificate
+    end
+  end
+
+  def destroy
+    delete_certificate if truststore_exists? && certificate_present?
+  end
+
+  def exists?
+    truststore_exists? && !certificate_changed?
+  end
+
+  private
+
+  def add_certificate
+    keytool(
+      '-import',
+      '-v',
+      '-noprompt',
+      '-storetype',
+      'pkcs12',
+      '-keystore',
+      resource[:truststore],
+      '-alias',
+      resource[:name],
+      '-file',
+      resource[:certificate],
+      '-storepass:file',
+      resource[:password_file]
+    )
+  end
+
+  def delete_certificate
+    delete = keytool(
+      '-delete',
+      '-v',
+      '-noprompt',
+      '-keystore',
+      resource[:truststore],
+      '-alias',
+      resource[:name],
+      '-storepass:file',
+      resource[:password_file]
+    )
+  end
+
+  def truststore_contents
+    keytool(
+      '-list',
+      '-keystore',
+      resource[:truststore],
+      '-storepass:file',
+      resource[:password_file]
+    )
+  end
+
+  def truststore_exists?
+    File.exist?(resource[:truststore])
+  end
+
+  def certificate_present?
+    truststore_exists? && truststore_contents.include?(resource[:name])
+  end
+
+  def certificate_changed?
+    return true unless certificate_present?
+
+    fingerprint = openssl('x509', '-noout', '-fingerprint', '-in', resource[:certificate]).strip.split('=')[1]
+    !truststore_contents.include?(fingerprint)
+  end
+
+end
--- a/spec/acceptance/candlepin_spec.rb
+++ b/spec/acceptance/candlepin_spec.rb
@@ -163,4 +163,47 @@ describe 'certs' do
     its(:stdout) { should match(/^Owner: CN=localhost, OU=SomeOrgUnit, O=Katello, ST=North Carolina, C=US$/) }
     its(:stdout) { should match(/^Issuer: CN=#{host_inventory['fqdn']}, OU=SomeOrgUnit, O=Katello, L=Raleigh, ST=North Carolina, C=US$/) }
   end
+
+  context 'updates java-client certificate in truststore if it changes' do
+    let(:pp) do
+      <<-EOS
+      user { 'tomcat':
+        ensure => present,
+      }
+
+      ['/usr/share/tomcat/conf', '/etc/candlepin/certs'].each |$dir| {
+        exec { "mkdir -p ${dir}":
+          creates => $dir,
+          path    => ['/bin', '/usr/bin'],
+        }
+      }
+
+      package { 'java-1.8.0-openjdk-headless':
+        ensure => installed,
+      }
+
+      include certs::candlepin
+      EOS
+    end
+
+    it "checks that the fingerprint matches" do
+      apply_manifest(pp, catch_failures: true)
+
+      initial_fingerprint_output = on default, 'openssl x509 -noout -fingerprint -in /etc/pki/katello/certs/java-client.crt'
+      initial_fingerprint = initial_fingerprint_output.output.strip.split('=').last
+      initial_truststore_output = on default, "keytool -list -keystore /etc/candlepin/certs/truststore -storepass $(cat #{truststore_password_file})"
+      expect(initial_truststore_output.output.strip).to include(initial_fingerprint)
+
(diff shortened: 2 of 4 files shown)


lib/puppet/provider/truststore_certificate/truststore_certificate.rb
```diff
@@ -0,0 +1,77 @@
...
+  def truststore_exists?
+    File.exist?(resource[:truststore])
+  end
+
+  def certificate_present?
+    truststore_exists? && truststore_contents.include?(resource[:name])
+  end
+
+  def certificate_changed?
+    return true if !certificate_present?
```
ekohl, line 71:
```suggestion
    return true unless certificate_present?
```

lib/puppet/provider/truststore_certificate/truststore_certificate.rb
```diff
@@ -0,0 +1,77 @@
...
+    add_certificate unless certificate_present?
+    if certificate_changed?
+      delete_certificate
+      add_certificate
+    end
+  end
+
+  def exists?
+    truststore_exists? && !certificate_changed?
+  end
```
ekohl, line 16: I think it would be better to override `insync?` on the property. https://github.com/theforeman/puppet-foreman/blob/c3c090ce8645a3f3da9cf7290ac13cab1503d8c6/lib/puppet/type/foreman_smartproxy.rb#L28-L43 is an example of how you can compare the certificate.

I think this part is what actually implements the property: https://github.com/theforeman/puppet-foreman/blob/c3c090ce8645a3f3da9cf7290ac13cab1503d8c6/lib/puppet/provider/foreman_smartproxy/rest_v3.rb#L43-L60

So `truststore_contents` would actually be the property `certificate` (in the provider) and then in the type you would compare those 2. `certificate=` (in the provider) would call `add_certificate` and `delete_certificate`.

spec/acceptance/candlepin_spec.rb
```diff
@@ -163,4 +163,49 @@ class { 'certs::candlepin':
...
+    end
+
+    before(:all) do
+      pp_setup = <<-EOS
+        exec { "rm -rf /root/ssl-build/localhost":
+          path => "/bin:/usr/bin",
+        }
+      EOS
+
+      apply_manifest(pp_setup, catch_failures: true)
```
ekohl, line 196: You don't need to do this via Puppet, I think this also works:
```suggestion
  on default, 'rm -rf /root/ssl-build/localhost'
```

spec/acceptance/candlepin_spec.rb
```diff
@@ -163,4 +163,49 @@ class { 'certs::candlepin':
...
+    it_behaves_like 'a idempotent resource'
+
+    it "checks that the fingerprint matches" do
+      keytool = <<-EOS
+        exec { "keytool -list -keystore /etc/candlepin/certs/truststore -storepass $(cat #{truststore_password_file}) | grep $(openssl x509 -noout -fingerprint -in /etc/pki/katello/certs/java-client.crt | cut -d '=' -f 2)":
+          path => "/bin:/usr/bin",
+        }
+      EOS
+
+      apply_manifest(keytool, catch_failures: true)
```
ekohl, line 208: Is the [command](https://serverspec.org/resource_types.html#command) resource perhaps better?

ekohl (commented):
Overall :+1: but some inline comments.

spec/acceptance/candlepin_spec.rb
ehelms: It did not seem to handle pipes

lib/puppet/provider/truststore_certificate/truststore_certificate.rb
ehelms: I opted to update ensurable's `insync` to decide if something should be changed. I found tacking this onto a single property to be confusing when it's the entirety of the resource we care about.

jturel (approved):
ACK from a functionality perspective. Tested with the PR in place on a fresh 3.18 install - truststore certs worked. Then I changed the hostname with katello-change-hostname and the certs still worked (and I had one artemis-client cert in the truststore at the end).

Thanks @ehelms ! :+1:

jturel:
It looks like there are two discourse threads seemingly related to the issue here. Is this going to automatically reconcile cases where the artemis-client cert doesn't match the current CA? If not, I'll file a separate issue.

[1] https://community.theforeman.org/t/katello-candlepineventlistener-not-starting-after-upgrade-to-foreman-2-2-2-katello-3-17/21857
[2] https://community.theforeman.org/t/candlepin-events-fail/21239

ehelms:
Yes, with this fix present, users who run the installer will get an automatic fix. For users with this issue today, they should be able to:

 1) rm /etc/candlepin/certs/truststore
 2) foreman-installer

I will look over the threads and comment.

lib/puppet/provider/truststore_certificate/truststore_certificate.rb
```diff
@@ -0,0 +1,81 @@
...
+  def create
+    add_certificate unless certificate_present?
+    if certificate_changed?
+      delete_certificate
+      add_certificate
+    end
+  end
+
+  def destroy
+    delete_certificate if trustore_exists? && certificate_present?
```
ekohl, line 15:
```suggestion
    delete_certificate if truststore_exists? && certificate_present?
```

spec/acceptance/candlepin_spec.rb
ekohl: Then the more correct thing is
```ruby
it "checks that the fingerprint matches" do
  fingerprint_output = on default, 'openssl x509 -noout -fingerprint -in /etc/pki/katello/certs/java-client.crt'
  fingerprint = fingerprint_output.split('=').last
  expect(on default, "keytool -list -keystore /etc/candlepin/certs/truststore -storepass file:#{truststore_password_file}").to include(fingerprint)
end
```

ekohl (changes requested):

ehelms:
@ekohl could you give this another review? thanks

ehelms:
At least on this version, if I revert `manifests/candlepin.pp` but leave the test in place:

```
Failures:

  1) certs updates java-client certificate in truststore if it changes checks that the fingerprint matches
     Failure/Error: expect(truststore_output.output).to include(fingerprint)
       expected "Keystore type: PKCS12\nKeystore provider: SUN\n\nYour keystore contains 2 entries\n\nartemis-client,...ry, \nCertificate fingerprint (SHA1): 85:36:10:06:AD:63:74:10:3B:3C:FE:9F:35:2D:B7:15:0C:88:0C:9A\n" to include "02:02:69:AE:3B:85:36:62:CB:F5:CD:72:28:AD:1F:A5:36:43:AA:FE\n"
       Diff:
       @@ -1,2 +1,10 @@
       -02:02:69:AE:3B:85:36:62:CB:F5:CD:72:28:AD:1F:A5:36:43:AA:FE\n
       +Keystore type: PKCS12
       +Keystore provider: SUN
       +
       +Your keystore contains 2 entries
       +
       +artemis-client, Jan 25, 2021, trustedCertEntry, 
       +Certificate fingerprint (SHA1): 3B:4C:03:71:86:7A:7A:7B:0E:D3:CF:B4:4F:79:95:28:7E:A4:8B:AB
       +candlepin-ca, Jan 25, 2021, trustedCertEntry, 
       +Certificate fingerprint (SHA1): 85:36:10:06:AD:63:74:10:3B:3C:FE:9F:35:2D:B7:15:0C:88:0C:9A
       
```

ehelms:
I'd like to wait on releasing this module for Foreman 2.4 until we have settled on a version of fixing this issue between the two PRs and getting it merged.

spec/acceptance/candlepin_spec.rb
```diff
@@ -163,4 +163,47 @@ class { 'certs::candlepin':
...
+
+      package { 'java-1.8.0-openjdk-headless':
+        ensure => installed,
+      }
+
+      include certs::candlepin
+      EOS
+    end
+
+    it "checks that the fingerprint matches" do
```
ehelms, line 189: @ekohl Check out this version of the test that has more assertions to help ensure that what was before is not now present after cleanup. We can use this test in your PR, or I can update my PR ti incorporate your changes. Just want to get a version of it in to fix this.

ehelms:
Closing in favor of https://github.com/theforeman/puppet-certs/pull/320

PR #313, opened 2021-01-15
Add more tests

ekohl:
See the individual commits for details.

### DIFF
--- a/lib/puppet_x/certs/common.rb
+++ b/lib/puppet_x/certs/common.rb
@@ -64,9 +64,7 @@ module PuppetX
         end
 
         autorequire(:ca) do
-          if @parameters.has_key?(:ca)
-            catalog.resource(@parameters[:ca].value.to_s).to_hash[:name]
-          end
+          [self[:ca]&.title].compact
         end
 
         autorequire(:file) do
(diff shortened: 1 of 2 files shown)


manifests/init.pp
```diff
@@ -73,7 +73,7 @@
   Boolean $generate = true,
   Boolean $regenerate = false,
   Boolean $deploy = true,
-  String $ca_common_name = $certs::params::ca_common_name,
+  Stdlib::Fqdn $ca_common_name = $certs::params::ca_common_name,
```
ehelms, line 76: Given this is the CA's common name, the previous string requirement should remain, as you should be able to put whatever descriptor you want for the CA here. Take puppet for example:

```
        Subject: CN=Puppet CA: pipe-katello-server-nightly-centos7.woodford.example.com
```

lib/puppet_x/certs/common.rb
```diff
@@ -43,25 +43,37 @@ module Common
...
 
         newparam(:ca) do
+          isrequired
+
           validate do |value|
             ca_resource = resource.catalog.resource(value.to_s)
-            if ca_resource && ca_resource.class.to_s != 'Puppet::Type::Ca'
-              raise ArgumentError, "Expected Ca resource, got #{ca_resource.class} #{ca_resource.inspect}"
+            if ca_resource
+              # rspec-puppet presents Puppet::Resource instances
```
ehelms, line 51: What does this comment mean? Is this like "we have to do the next line because of our test infrastructure" ?

ekohl:
Splitting the PR: https://github.com/theforeman/puppet-certs/pull/422

lib/puppet_x/certs/common.rb
ekohl: Pretty much yes.

PR #315, opened 2021-02-11
Remove qpid_client class and avoid resource defaults in qpid class

ehelms:


ehelms:
The qpid_client class was targeted at supporting Pulp 2 use of Qpid.

ekohl (commented):
For the changelog I think it would be clearer to write `Remove cert::qpid_client class` and the other commit `Avoid resource defaults and set explicit parameters`. Other than that this looks ok to me.

ehelms:
Updated to two commits

ekohl (commented):
It mentions `qpid::client` but really it removes `certs::qpid_client`.

wbclark (commented):
The commit message (both short and full) still references `qpid::client` when it should be `qpid_client` or `certs::qpid_client`

LGTM otherwise

ekohl (approved):

PR #316, opened 2021-02-16
Refs #31878 - Split qpid router server and client certificates

ehelms:
This splits the qpid router server and client certificates into separate classes so that they can be used independently. In particular, we want to allow the puppet-foreman_proxy_content usage of this to be able to set a special CN when used as a hub configuration (https://github.com/theforeman/puppet-foreman_proxy_content/pull/327/files#diff-b52d4cdda63edbb25bbbcb2e7c293e30f0fa39b9fadab9edf9f070b2c27fc00dR23).


jturel (approved):
ACK - tested working in conjunction w/ my Katello PR

manifests/qpid_router_client.pp
```diff
@@ -0,0 +1,57 @@
+# Constains certs specific configurations for qpid dispatch router
+class certs::qpid_router_client (
```
ekohl, line 2: How about:
```suggestion
class certs::qpid_router::client (
```

manifests/qpid_router.pp
```diff
@@ -22,7 +20,6 @@
 ) inherits certs {
 
   $server_keypair = "${hostname}-qpid-router-server"
```
ekohl, line 22: It feels like we should rename this class to `certs::qpid_router::server`

ekohl (commented):
Overall :+1: for splitting the functionality. And yes, I do realize you could also argue for `certs::qpid::server` and `certs::qpid::client` and ideally I'd like to see that, but I'm fine with keeping them to keep this PR focused on just qpid router.

jlsherrill (approved):
ACK based on @jturel's testing

PR #312, opened 2021-01-14
Fixes #31574: Ensure truststore certificates get updated when they change

ekohl:
Alternative to https://github.com/theforeman/puppet-certs/pull/311. This implementation takes the approach of trying to figure out the fingerprints for the certificate property. This means that the changed value is also more readable.

There is a bit of overhead since keytool and openssl are called multiple times with the same parameters. These were pretty fast and I figured that caching was more likely to break something than to improve it.

Some manual verification. If we decide to go this route, I can convert them to automated tests.

```console
# ls /etc/candlepin/certs/truststore2
ls: cannot access /etc/candlepin/certs/truststore2: No such file or directory
# puppet resource truststore_certificate candlepin-ca2 truststore=/etc/candlepin/certs/truststore2 password_file=/etc/pki/katello/truststore_password-file
truststore_certificate { 'candlepin-ca2':
  ensure   => 'absent',
  provider => 'truststore_certificate',
}
# puppet resource truststore_certificate candlepin-ca2 truststore=/etc/candlepin/certs/truststore2 password_file=/etc/pki/katello/truststore_password-file certificate=/etc/candlepin/certs/candlepin-ca.crt
Notice: /Truststore_certificate[candlepin-ca2]/ensure: created
truststore_certificate { 'candlepin-ca2':
  ensure      => 'present',
  certificate => '5E:3F:23:30:46:5F:1A:70:E4:45:85:87:B9:6A:7B:39:20:9B:DF:83',
  provider    => 'truststore_certificate',
}
# puppet resource truststore_certificate candlepin-ca2 truststore=/etc/candlepin/certs/truststore2 password_file=/etc/pki/katello/truststore_password-file certificate=/etc/pki/katello-certs-tools/certs/centos7-64.example.com-tomcat.crt
Notice: /Truststore_certificate[candlepin-ca2]/certificate: certificate changed '5E:3F:23:30:46:5F:1A:70:E4:45:85:87:B9:6A:7B:39:20:9B:DF:83' to '02:7D:1F:42:86:22:54:27:A8:20:5C:43:4F:87:0D:E8:06:50:AD:4B'
truststore_certificate { 'candlepin-ca2':
  ensure      => 'present',
  certificate => '02:7D:1F:42:86:22:54:27:A8:20:5C:43:4F:87:0D:E8:06:50:AD:4B',
  provider    => 'truststore_certificate',
}
[...]

PR #320, opened 2021-03-17
Fixes #31574: Ensure truststore certificates get updated when they change

ehelms:
This combines https://github.com/theforeman/puppet-certs/pull/312 with the updated testing from https://github.com/theforeman/puppet-certs/pull/311/files#diff-5afea89f9f4123c60bf87969c1a02b0d75c0687a14758480314204b44ca679c1R189 to get to a PR that resolves this issue and we can pull in. There have been additional reports of users hitting this issue.

### DIFF
--- a/lib/puppet/type/truststore_certificate.rb
+++ b/lib/puppet/type/truststore_certificate.rb
@@ -1,0 +1,47 @@
+Puppet::Type.newtype(:truststore_certificate) do
+  desc 'adds a certificate to a pkcs12 truststore'
+
+  ensurable
+
+  def self.title_patterns
+    [ [ /(.+):(.+)/m, [ [:truststore], [:alias] ] ] ]
+  end
+
+  newparam(:alias, :namevar => true) do
+    desc "The certificate alias used to store inside the truststore"
+  end
+
+  newparam(:truststore, :namevar => true) do
+    desc "Path to the truststore to use or create when importing the certificate"
+    isrequired
+  end
+
+  newproperty(:certificate) do
+    desc "Path to the certificate to add to the truststore"
+
+    def fingerprint(file)
+      provider.fingerprint(file)
+    end
+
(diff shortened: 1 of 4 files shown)


ehelms:
@ekohl @wbclark Could one of you look at and help me prioritize this? I've seen another two user reports this week where its being hit.

ekohl (approved):
Thanks!

ekohl:
Not entirely sure how to classify it. It feels like a bugfix, but at the size that I wouldn't feel comfortable cherry picking it without careful consideration. Given the next is major anyway, it's not too critical but that's why I first chose enhancement and ended up with bug.

ehelms:
:+1: to bug in that it definitely solves a problem that users run into -- thank you for getting back to this (and the original enhancement to my original PR that made it cleaner)

lib/puppet/type/truststore_certificate.rb
```diff
@@ -0,0 +1,47 @@
...
+  newparam(:password_file) do
+    desc "Path to file containing the truststore password"
+    isrequired
+  end
+
+  autorequire(:file) do
+    [self[:password_file], File.dirname(self[:truststore]), self[:certificate]]
+  end
+
+  autonotify(:file) do
```
ekohl, line 44: Linking back: it was impossible to write a test for this using rspec-puppet. https://github.com/rodjek/rspec-puppet/commit/7484feb56ffa1380ffe778c7f9f89daa06c07570 has been merged but is unreleased. Once it's merged, it should be possible to test this:
```ruby
it { is_expected.to contain_truststore_certificate('the-cert').that_notifies('File[/truststore]') }
```

ehelms: Maybe file an issue to remember this? Or just open a draft PR with this added?

PR #322, opened 2021-04-22
Refs #32381: Drop the generation of client certificates for Artemis c…

ehelms:
…ommunication

Services interfacing with Artemis should generate their own
client certificates and pass the client cert to to the certs::candlepin
class for inclusion in the truststore.

Should not be merged until puppet-katello no longer relies on this client certificate.

### DIFF
--- a/manifests/candlepin.pp
+++ b/manifests/candlepin.pp
@@ -22,29 +22,10 @@ class certs::candlepin (
   $ca_key_password_file     = $certs::ca_key_password_file,
   $user                     = $certs::user,
   $group                    = 'tomcat',
-  $client_keypair_group     = 'tomcat',
+  $client_cert              = undef,
 ) inherits certs {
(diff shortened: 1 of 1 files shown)


manifests/candlepin.pp
```diff
@@ -22,29 +22,10 @@
...
   $group                    = 'tomcat',
-  $client_keypair_group     = 'tomcat',
+  $client_cert              = undef,
 ) inherits certs {
-  $java_client_cert_name = 'java-client'
-  $artemis_alias = 'artemis-client'
-  $artemis_client_dn = "CN=${hostname}, OU=${org_unit}, O=candlepin, ST=${state}, C=${country}"
-
-  cert { $java_client_cert_name:
-    ensure        => present,
```
ekohl, line 32: Should we also clean it up with an `ensure => absent`

PR #323, opened 2021-04-22
Refs #31574: Compare SHA256 fingerprints when checking truststore

ehelms:
I'm very confused. Between the last time https://github.com/theforeman/puppet-certs/pull/320 ran which as March 17th, something has changed that causes the idempotency of this new type and provider to break.  This only seems to occur on EL7. I have added two debug statements here that highlight so far what I know which is that the `is` value is nil when doing the `insync?` check.

### DIFF
--- a/lib/puppet/provider/truststore_certificate/keytool.rb
+++ b/lib/puppet/provider/truststore_certificate/keytool.rb
@@ -26,7 +26,7 @@ Puppet::Type.type(:truststore_certificate).provide(:keytool) do
   def fingerprint(file)
     return unless File.exist?(file)
 
-    openssl('x509', '-noout', '-fingerprint', '-in', file).strip.split('=')[1]
+    openssl('x509', '-sha256', '-noout', '-fingerprint', '-in', file).strip.split('=')[1]
   end
 
   def file_readable?(file)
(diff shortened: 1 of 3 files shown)


ehelms:
All I can think is some newer Puppet release introduced something that broke this -- hoping you see the magic to fix this @ekohl

wbclark (commented):
I think the `unless =>` here is not working / should never have worked: https://github.com/ehelms/puppet-certs/blob/9481972d0039db21313ef19760279107e3f39e22/manifests/candlepin.pp#L109

Docs for the puppet Exec resource state "Puppet runs this test command first, then runs the main command unless the test has an exit code of 0 (success)"

It looks like `cut` always has an exit code of `0`:

```
$ echo "foo=foo" | grep "foo" | cut -d '=' -f 2
foo
$ echo $?
0
$ echo "bar=bar" | grep "foo" | cut -d '=' -f 2
$ echo $?
0 
```

ehelms:
Thanks for looking at it. Note the issue is with the truststore and not the
keystore. Additionally that cut line is attached to the openssl command
which is ultimately sent to the grep statement (check the parens).

wbclark:
> Thanks for looking at it. Note the issue is with the truststore and not the keystore. Additionally that cut line is attached to the openssl command which is ultimately sent to the grep statement (check the parens).

Thanks for the explanation, that makes a lot more sense.

ehelms:
What I was seeing locally was that the truststore on EL7 switched to SHA-256 by default:

```
Certificate fingerprint (SHA-256): 7C:0F:C2:F3:1E:F7:73:C9:4D:27:0F:8A:4B:FE:A0:75:D1:68:2E:1E:F7:08:E5:77:4D:4C:2A:B1:1B:AC:1C:EA
```

I could not see a way to influence the reporting of the fingerprint from the keystore. I also am testing and need to update the keystore check as well to `-sha256`

ekohl (approved):
It's so weird that this passed before.

PR #324, opened 2021-04-30
Refs #32383: Drop pulp_client certificates

ehelms:
Katello has switched to using the Foreman client certificates
to communicate with Pulp. Generating these Pulp client certificates
is no longer necessary.

Requires:
 - https://github.com/theforeman/puppet-foreman_proxy_content/pull/350
 - https://github.com/theforeman/puppet-katello/pull/411


ekohl (commented):
Should we still have some code to ensure they're absent?

ehelms:
> Should we still have some code to ensure they're absent?

If you have a suggestion on how we do that without having to have this linger around and constrain our code to this class. And I'd ask what do we ensure is absent?

 * the final path
 * the installed path from the RPM
 * the installed RPM
 * the certs in ssl-build

ekohl:
> * the final path
> 
> * the installed path from the RPM
> 
> * the installed RPM

I think these make sense

> the certs in ssl-build

I wouldn't care that much about this one.

ehelms:
Is there a strategy to include this cleanup without having to have https://github.com/theforeman/puppet-katello/pull/411/files#diff-9e68d1f4597ae6d5dd2e74efc344b09bb14d98648b05f0630419cf8bebc82d1dL18 tied to this class for an extended period of time?

ekohl:
Not that I know of. We can adopt a policy where we keep the cleanup in place for 2 or 3 releases. It would allow users to skip an intermediate upgrade and still benefit from the cleanup.

ehelms:
Could we do something where instead of puppet-katello having to carry around the `include certs::pulp_client` for multiple releases we make pulp_client class here do cleanup and then add to puppet-certs init.pp an `include certs::pulp_client` so that it performs any cleanup and owns the dependency allowing us to keep other modules clean (and drop dependencies).

ekohl:
We can include it in the main `certs` class with an `ensure => absent`. That's probably a good pattern.

PR #326, opened 2021-05-03
Add acceptance tests for certs::qpid

ehelms:


ehelms:
Looking for some help here, in the tests, I have been trying to do a standard test of the certificate's content by checking the Issuer and Subject lines to match. However, from system to system (CI vs. local, EL7 vs. EL8) the output of the certificate from the nssdb seems to vary and matching that is difficult.

Basically, my question is how to match a line that presents like below, but does not "line-break" in the same way all the time:

```
       +        Issuer: "CN=centos7-64.example.com,OU=SomeOrgUnit,O=Katello,L=Raleigh
       +            ,ST=North Carolina,C=US"
```

ekohl:
You currently expect `/Issuer: "CN=centos7-64.example.com"/` so that includes the closing quote. You could do a partial match instead.

If you want to match the full DN, perhaps you can capture it with `\s*` between the commas, assuming it's a multi line regex. I looked in the mapage of certutil and didn't see anything that disabled line wrapping.

ehelms:
Woot, tests are passing now.

spec/acceptance/qpid_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+    describe command("certutil -L -d #{nssdb_dir}") do
+      its(:exit_status) { should eq 0 }
+      its(:stdout) { should match(/^ca                                                           CT,C,c$/i) }
+      its(:stdout) { should match(/^broker                                                       u,u,u$/i) }
+    end
+
+    describe command("certutil -L -d #{nssdb_dir} -n ca") do
+      its(:exit_status) { should eq 0 }
+      its(:stdout) { should match(/        Subject: "CN=#{host_inventory['fqdn']},OU=SomeOrgUnit,O=Katello,L=Raleig
+            h,ST=North Carolina,C=US"/) }
```
ekohl, line 58: I think I'd prefer this if it works.
```suggestion
      its(:stdout) { should match(/\s*Subject: "CN=#{host_inventory['fqdn']},OU=SomeOrgUnit,O=Katello,L=Raleig\n            h,ST=North Carolina,C=US"/) }
```

ehelms: Done, thanks for that suggestion.

ekohl (approved):

PR #327, opened 2021-05-04
Fixes #32506: Add an nssdb_certificate type and provider 

ehelms:
This adds a type for managing NSS database certificates (and private keys)
within an NSS database. This new type reduces the number of intermediary
entities that have to be created, for example the pkcs12 export, and
reduces the overall Puppet code required. This type aims to provide
a clean interface to managing NSS database certificates and ensure
that the NSS database is in sync with the certificates being managed.

Requires: https://github.com/theforeman/puppet-certs/pull/326

### DIFF
--- a/lib/puppet/provider/nssdb_certificate/certutil.rb
+++ b/lib/puppet/provider/nssdb_certificate/certutil.rb
@@ -1,0 +1,117 @@
+Puppet::Type.type(:nssdb_certificate).provide(:certutil) do
+  commands :certutil => 'certutil'
+  commands :openssl => 'openssl'
+  commands :pk12util => 'pk12util'
+
+  def create
+    add_certificate
+    add_private_key if resource[:private_key]
+  end
+
+  def destroy
+    delete_certificate
+    delete_private_key if resource[:private_key]
+  end
+
+  def exists?
+    nssdb_content
+  end
+
+  def certificate
+    nssdb_fingerprint
+  end
+
+  def certificate=(value)
+    delete_certificate unless nssdb_content.nil?
+    add_certificate
+  end
+
+  def fingerprint(file)
+    return unless File.exist?(file)
+
+    openssl('x509', '-sha256', '-noout', '-fingerprint', '-in', file).strip.split('=')[1]
+  rescue Puppet::ExecutionFailure => e
+    Puppet.warn("Failed to read certificate #{file}: #{e}")
+    nil
+  end
+
+  private
+
+  def add_certificate
+    certutil(
+      '-A',
+      '-a',
+      '-d', resource[:nssdb],
+      '-n', resource[:cert_name],
+      '-t', resource[:trustargs],
+      '-i', resource[:certificate],
+      '-f', resource[:password_file]
+    )
+  end
+
+  def delete_certificate
+    certutil(
+      '-D',
+      '-d', resource[:nssdb],
+      '-n', resource[:cert_name]
+    )
+  end
+
+  def add_private_key
+    Tempfile.open('pkcs12') do |pkcs12|
+      openssl(
+        'pkcs12',
+        '-export',
+        '-in', resource[:certificate],
+        '-inkey', resource[:private_key],
+        '-out', pkcs12.path,
+        '-password', "file:#{resource[:password_file]}",
+        '-name', resource[:cert_name]
+      )
+
+      pk12util(
+        '-i', pkcs12.path,
+        '-d', resource[:nssdb],
+        '-w', resource[:password_file],
+        '-k', resource[:password_file]
+      )
+    end
+  end
+
+  def delete_private_key
+    certutil(
+      '-F',
+      '-d', resource[:nssdb],
+      '-n', resource[:cert_name]
+    )
+  end
+
+  def nssdb_content
+    return unless directory_readable?(resource[:nssdb])
+
+    certutil(
+      '-L',
+      '-a',
+      '-d', resource[:nssdb],
+      '-n', resource[:cert_name]
+    )
+  rescue Puppet::ExecutionFailure => e
+    Puppet.debug("Failed to read nssdb contents from #{resource[:nssdb]}: #{e}")
+    nil
+  end
+
+  def nssdb_fingerprint
+    cert_info = nssdb_content
+    return unless cert_info
+
+    Tempfile.open('cert') do |temp_cert|
+      temp_cert.write(cert_info)
+      temp_cert.rewind
+      fingerprint(temp_cert.path)
+    end
+  end
+
+  def directory_readable?(file)
+    File.directory?(file) && File.readable?(file)
+  end
+end
--- a/lib/puppet/type/nssdb_certificate.rb
+++ b/lib/puppet/type/nssdb_certificate.rb
@@ -1,0 +1,52 @@
+Puppet::Type.newtype(:nssdb_certificate) do
+  desc 'adds a certificate to an nssdb'
+
+  ensurable
+
+  def self.title_patterns
+    [ [ /(.+):(.+)/m, [ [:nssdb], [:cert_name] ] ] ]
+  end
+
+  newparam(:cert_name, :namevar => true) do
+    desc "The certificate name used to store inside the nssdb"
+  end
+
+  newparam(:nssdb, :namevar => true) do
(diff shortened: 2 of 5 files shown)


lib/puppet/provider/nssdb_certificate/certutil.rb
```diff
@@ -0,0 +1,133 @@
...
+  rescue Puppet::ExecutionFailure => e
+    Puppet.debug("Failed to read nssdb contents: #{e}")
+    nil
+  end
+
+  def nssdb_fingerprint
+    finger_print = nil
+
+    Tempfile.open('cert') do |temp_cert|
+      temp_cert.write(nssdb_content)
```
ekohl, line 126: This can return nil in which case there is no fingerprint.

lib/puppet/provider/nssdb_certificate/certutil.rb
```diff
@@ -0,0 +1,133 @@
...
+    add_certificate
+  end
+
+  def fingerprint(file)
+    return unless File.exist?(file)
+
+    openssl('x509', '-sha256', '-noout', '-fingerprint', '-in', file).strip.split('=')[1]
+  end
+
+  def directory_readable?(file)
```
ekohl, line 35: Can be private?

lib/puppet/type/nssdb_certificate.rb
```diff
@@ -0,0 +1,52 @@
...
+    isrequired
+  end
+
+  newparam(:password_file) do
+    desc "Path to file containing the nssdb password"
+    isrequired
+  end
+
+  autorequire(:file) do
+    [self[:password_file], File.dirname(self[:nssdb]), self[:certificate], self[:private_key]]
```
ekohl, line 50: Should this require the nssdb directory itself? That can ensure the directory is never world readable: first ensure the directory has correct permissions, then run nssdb commands.

ekohl (commented):
I like this. Just a first look and there's probably more.

lib/puppet/type/nssdb_certificate.rb
ehelms: Sorry, I do not follow the nuance of the question. `File.dirname(self[:nssdb])` causes a require on the directory.

ekohl: The parent directory, but not the directory itself. That can mean you end up running
* `File.dirname(self[:nssdb])`
* nssdb_certificate itself
* `self[:nssdb]`

What I'm asking is: should we always ensure:
* `self[:nssdb]`
* nssdb_certificate itself

And because the file type also requires parents, you get `File.dirname()` for free.

ehelms: Ahh yes, will make that change. A bit of a copy paste fail on my part adapting the truststore_certificate work.

ehelms:
Rebased now that the acceptance tests are in master

lib/puppet/provider/nssdb_certificate/certutil.rb
```diff
@@ -0,0 +1,136 @@
...
+  end
+
+  def nssdb_content
+    return unless directory_readable?(resource[:nssdb])
+
+    certutil(
+      '-L',
+      '-a',
+      '-d', resource[:nssdb],
+      '-n', resource[:cert_name]
```
ekohl, line 111: I also thought about this option because it can make some options easier to understand. Should we apply it everywhere?

lib/puppet/provider/nssdb_certificate/certutil.rb
```diff
@@ -0,0 +1,136 @@
...
+  def certificate
+    nssdb_fingerprint
+  end
+
+  def certificate=(value)
+    delete_certificate unless nssdb_content.nil?
+    add_certificate
+  end
+
+  def fingerprint(file)
```
ekohl, line 29: Should it have a rescue section? Not sure what we can expect if the file contains some invalid content should be an easy test. I'd expect it to log a warning and return nil.

lib/puppet/provider/nssdb_certificate/certutil.rb
```diff
@@ -0,0 +1,136 @@
...
+
+    unless cert_info.nil?
+      Tempfile.open('cert') do |temp_cert|
+        temp_cert.write(cert_info)
+        temp_cert.rewind
+        finger_print = fingerprint(temp_cert.path)
+      end
+    end
+
+    finger_print
```
ekohl, line 130: I think this is the same
```suggestion
    cert_info = nssdb_content
    return unless cert_info

    Tempfile.open('cert') do |temp_cert|
      temp_cert.write(cert_info)
      temp_cert.rewind
      fingerprint(temp_cert.path)
    end
```

ekohl (commented):
I don't see code to modify the trust args. Is that something that is supposed to be supported?

lib/puppet/provider/nssdb_certificate/certutil.rb
ehelms: The `-n` option? It is used in the other calls in the file.

ehelms:
> I don't see code to modify the trust args. Is that something that is supposed to be supported?

As in changing the trustargs without changing the certificate?

ehelms:
> I don't see code to modify the trust args. Is that something that is supposed to be supported?

As in changing the trustargs without changing the certificate?

lib/puppet/provider/nssdb_certificate/certutil.rb
ekohl: I meant the style of having two related options on a single line.

ehelms: Ahh yes, I also like this style and have applied it throughout the file.

lib/puppet/provider/nssdb_certificate/certutil.rb
```diff
@@ -0,0 +1,119 @@
...
+    delete_certificate unless nssdb_content.nil?
+    add_certificate
+  end
+
+  def fingerprint(file)
+    return unless File.exist?(file)
+
+    openssl('x509', '-sha256', '-noout', '-fingerprint', '-in', file).strip.split('=')[1]
+  rescue Puppet::ExecutionFailure => e
+    Puppet.warn("Failed to read certificate: #{e}")
```
ekohl, line 34:
```suggestion
    Puppet.warn("Failed to read certificate #{file}: #{e}")
```

lib/puppet/provider/nssdb_certificate/certutil.rb
```diff
@@ -0,0 +1,119 @@
...
+    end
+  end
+
+  def delete_private_key
+    certutil(
+      '-F',
+      '-d',
+      resource[:nssdb],
+      '-n',
+      resource[:cert_name]
```
ekohl, line 87:
```suggestion
      '-d', resource[:nssdb],
      '-n', resource[:cert_name]
```

lib/puppet/provider/nssdb_certificate/certutil.rb
```diff
@@ -0,0 +1,119 @@
...
+    return unless directory_readable?(resource[:nssdb])
+
+    certutil(
+      '-L',
+      '-a',
+      '-d', resource[:nssdb],
+      '-n', resource[:cert_name]
+    )
+  rescue Puppet::ExecutionFailure => e
+    Puppet.debug("Failed to read nssdb contents: #{e}")
```
ekohl, line 101:
```suggestion
    Puppet.debug("Failed to read nssdb contents from #{resource[:nssdb]}: #{e}")
```

ekohl (commented):
> As in changing the trustargs without changing the certificate?

Exactly. I don't even know if it's supported and we don't rely on it now. That means I'd be OK with noting it as a known limitation in the parameter description. I just want to prevent us from being surprised when it doesn't work if we do need it.

Some small inline suggestions, but pretty much :+1:

ekohl (approved):

ehelms:
I am going to hold merging this until the outcome of some design discussions happening here: https://github.com/theforeman/puppet-certs/pull/329#discussion_r629315815

This structurally has parallels to the truststore / keystore discussions and having a congruent design model has benefits.

ehelms:
Given with the keystore I have settled on a model that is managing the store and the certificates with dedicated resources rather than one catch all resource I am proceeding with the design here.

PR #332, opened 2021-05-12
Fixes #32585: Add function to extract Artemis client certificate subj…

ehelms:
…ect dynamically

### DIFF
--- a/lib/puppet/functions/certs/certificate_subject.rb
+++ b/lib/puppet/functions/certs/certificate_subject.rb
@@ -1,0 +1,23 @@
+# @summary
+#   Extracts a certificate's subject in RFC2253 form
+#
+# @example How to extract a certificate's subject
+#   $cert_subject = certs::certificate_subject($path_to_certificate)
+#
+Puppet::Functions.create_function(:'certs::certificate_subject') do
+  # @param certificate_path
+  dispatch :certificate_subject do
+    param 'String[1]', :certificate_path
+    return_type 'Optional[String]'
+  end
+
+  def certificate_subject(certificate_path)
+    begin
+      cert = OpenSSL::X509::Certificate.new(File.read(certificate_path))
+      cert.subject.to_s(OpenSSL::X509::Name::RFC2253)
+    rescue OpenSSL::X509::CertificateError, Errno::ENOENT => e
+      Puppet.debug("The file at #{certificate_path} could not be read or is not a valid x509 certificate: #{e}")
+      nil
+    end
+  end
+end
--- a/manifests/candlepin.pp
+++ b/manifests/candlepin.pp
@@ -26,7 +26,6 @@ class certs::candlepin (
 ) inherits certs {
   $java_client_cert_name = 'java-client'
   $artemis_alias = 'artemis-client'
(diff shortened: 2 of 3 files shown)


ehelms:
Incorporates comments from https://github.com/theforeman/puppet-katello/pull/405

manifests/candlepin.pp
```diff
@@ -76,6 +75,8 @@
   $alias = 'candlepin-ca'
 
   if $deploy {
+    $artemis_client_dn = certs::certificate_subject($client_cert)
```
ekohl, line 78: I still suspect we may need to use Deferred. Regular functions run on the Puppetserver while the certificate files are on the agent. While it may work in our installer where everything is on the same machine, I think it'll break in regular operation.

ehelms:
Requires https://github.com/theforeman/puppet-candlepin/pull/199

lib/puppet/functions/certs/certificate_subject.rb
```diff
@@ -0,0 +1,24 @@
...
+    param 'String[1]', :certificate_path
+    return_type 'Optional[String]'
+  end
+
+  def certificate_subject(certificate_path)
+    begin
+      cert = OpenSSL::X509::Certificate.new(File.read(certificate_path))
+      cert.subject.to_s(OpenSSL::X509::Name::RFC2253)
+    rescue OpenSSL::X509::CertificateError, Errno::ENOENT
+      nil
```
ekohl, line 21: Perhaps at least a debug line with the exception?

lib/puppet/functions/certs/certificate_subject.rb
```diff
@@ -0,0 +1,24 @@
+# @summary
+#   Extracts a certificate's subject in RFC2253 form
+#
+# @example How to extract a certificate's subject
+#   $cert_subject = certs::certificate_subject($path_to_certificate)
+#
+Puppet::Functions.create_function(:'certs::certificate_subject') do
+  # @param certificate_path
+  #
+  # @return Optional[String]
```
ekohl, line 10: If you use `return_type`, I think you don't need to specify this.

ekohl (approved):

lib/puppet/functions/certs/certificate_subject.rb
```diff
@@ -0,0 +1,23 @@
...
+    param 'String[1]', :certificate_path
+    return_type 'Optional[String]'
+  end
+
+  def certificate_subject(certificate_path)
+    begin
+      cert = OpenSSL::X509::Certificate.new(File.read(certificate_path))
+      cert.subject.to_s(OpenSSL::X509::Name::RFC2253)
+    rescue OpenSSL::X509::CertificateError, Errno::ENOENT
+      Puppet.debug("The file at #{certificate_path} could not be read or is not a valid x509 certificate.")
```
ekohl, line 19: Should we log why?
```suggestion
    rescue OpenSSL::X509::CertificateError, Errno::ENOENT => e
      Puppet.debug("The file at #{certificate_path} could not be read or is not a valid x509 certificate: #{e}.")
```

PR #334, opened 2021-05-18
Fixes #32506: Add keystore puppet provider type

ehelms:
This allows creating and managing a PKCS12 based keystore that
is initially empty. This allows managing the permissions and ownership
of the keystore separately from any certificates within it.

This is an alternative to what was discussed here -- https://github.com/theforeman/puppet-certs/pull/329#discussion_r629315815

The follow up would be then to introduce a type `keystore_certificate` that would roughly look like:

```
    keystore_certificate { "${keystore}:tomcat":
      ensure          => 'present',
      certificate     => $tomcat_cert,
      private_key  => $tomcat_key,
      ca_file           => $ca_cert,
      password_file => $keystore_password_path,
    }
```

Where keystore_certificate would have an autorequires on `Keystore[$keystore]` effectively. While I like the discussed all in one declaration:

```
keystore { $keystore:
  ensure        => present,
  password_file => $keystore_password_path,
  certificates  => [
    {
      alias => $alias,
      certificate => $ca_cert,
    },
    {
      alias => $artemis_alias,
      certificate => $client_cert,
      private_key => $client_key,
    },
  ],
}
```

My guy feeling is that having broken out, smaller dedicated resources will be easier to manage.

### DIFF
--- a/lib/puppet/type/keystore.rb
+++ b/lib/puppet/type/keystore.rb
@@ -1,0 +1,55 @@
+Puppet::Type.newtype(:keystore) do
+  desc 'generates an empty pkcs12 keystore'
+
+  ensurable
+
+  newparam(:keystore, :namevar => true) do
+    desc "Path to the keystore"
+    isrequired
+  end
+
+  newparam(:password_file) do
+    desc "Path to file containing the keystore password"
+    isrequired
+  end
+
+  newparam(:owner, parent: Puppet::Type::File::Owner) do
+    desc "Specifies the owner of the keystore. Valid options: a string containing a username or integer containing a uid."
+  end
+
+  newparam(:group, parent: Puppet::Type::File::Group) do
+    desc "Specifies a permissions group for the keystore. Valid options: a string containing a group name or integer containing a gid."
+  end
+
+  newparam(:mode, parent: Puppet::Type::File::Mode) do
+    desc "Specifies the permissions mode of the keystore. Valid options: a string containing a permission mode value in octal notation."
+  end
+
+  autorequire(:file) do
+    [self[:password_file]]
+  end
+
+  def generate
+    file_opts = {
+      ensure: (self[:ensure] == :absent) ? :absent : :file,
+      path: self[:keystore],
+    }
+
+    [:owner,
+     :group,
+     :mode].each do |param|
+      file_opts[param] = self[param] unless self[param].nil?
+    end
+
+    metaparams = Puppet::Type.metaparams
+    excluded_metaparams = [:before, :notify, :require, :subscribe, :tag]
+
+    metaparams.reject! { |param| excluded_metaparams.include? param }
+
+    metaparams.each do |metaparam|
+      file_opts[metaparam] = self[metaparam] unless self[metaparam].nil?
+    end
+
+    [Puppet::Type.type(:file).new(file_opts)]
+  end
+end
--- a/manifests/candlepin.pp
+++ b/manifests/candlepin.pp
@@ -110,12 +110,7 @@ class certs::candlepin (
       unless    => "keytool -list -keystore ${keystore} -storepass:file ${keystore_password_path} -alias tomcat | grep $(openssl x509 -noout -fingerprint -sha256 -in ${tomcat_cert} | cut -d '=' -f 2)",
       logoutput => 'on_failure',
       path      => ['/bin/', '/usr/bin'],
-    } ~>
-    file { $keystore:
-      ensure => file,
-      owner  => 'root',
-      group  => $group,
-      mode   => '0640',
+      require   => Keystore[$keystore],
     } ~>
     certs::keypair { 'candlepin':
       key_pair    => Cert[$java_client_cert_name],
@@ -129,6 +124,14 @@ class certs::candlepin (
       key_owner   => $user,
       key_group   => $client_keypair_group,
       key_mode    => '0440',
+    }
+
+    keystore { $keystore:
+      ensure        => present,
+      password_file => $keystore_password_path,
+      owner         => 'root',
+      group         => $group,
+      mode          => '0640',
     }
 
     file { $truststore_password_path:
--- a/lib/puppet/provider/keystore/keytool.rb
+++ b/lib/puppet/provider/keystore/keytool.rb
@@ -1,0 +1,51 @@
+Puppet::Type.type(:keystore).provide(:keytool) do
+  commands :keytool => 'keytool'
+
+  def create
+    generate_keystore
+  end
+
+  def destroy
+    delete_keystore
+  end
+
+  def exists?
+    File.exist?(resource[:keystore])
+  end
+
+  private
+
+  def generate_keystore
+    temp_alias = 'temporary-entry'
+
+    begin
+      keytool(
+        '-genkey',
+        '-storetype', 'pkcs12',
+        '-keystore', resource[:keystore],
+        '-storepass:file', resource[:password_file],
+        '-alias', temp_alias,
+        '-dname', "CN=#{temp_alias}"
+      )
+    rescue Puppet::ExecutionFailure => e
+      Puppet.error("Failed to generate new keystore with temporary entry: #{e}")
+      return nil
+    end
+
+    begin
+      keytool(
+        '-delete',
+        '-keystore', resource[:keystore],
+        '-storepass:file', resource[:password_file],
+        '-alias', temp_alias
+      )
+    rescue Puppet::ExecutionFailure => e
+      Puppet.error("Failed to delete temporary entry when generating empty keystore: #{e}")
+      return nil
+    end
+  end
+
+  def delete_keystore
+    File.rm(resource[:keystore])
+  end
+end


lib/puppet/type/keystore.rb
```diff
@@ -0,0 +1,55 @@
...
+    desc "Specifies a permissions group for the keystore. Valid options: a string containing a group name or integer containing a gid."
+  end
+
+  newparam(:mode, parent: Puppet::Type::File::Mode) do
+    desc "Specifies the permissions mode of the keystore. Valid options: a string containing a permission mode value in octal notation."
+  end
+
+  autorequire(:file) do
+    [self[:password_file]]
+  end
```
wbclark, line 30: Confirming I understood this correctly:

The Keystore will have a `password_file` (L6-9 above) so that the interface will include:

```puppet
keystore { $some_keystore:
  password_file => $some_password_file_path,
}
```

And this autorequire is silently similar to:

```puppet
keystore { $some_keystore:
  password_file => $some_password_file_path,
  require => File[$some_password_file_path],
}
```

?

ehelms: Correct.

wbclark (commented):
So this implements a keystore type and keytool provider for that type.

The provider implements the `create` method by first generating the keystore with a temporary entry and then immediately removing the temporary entry, so that the result of calling create is an empty keystore. `destroy` and `exists?` are implemented with standard `File` methods.

Owner, group owner, and mode are managed in the type as params with parents from the `File` type. I can see how these are then built into the `file_opts` hash within the `generate` method, which returns an array `[Puppet::Type.type(:file).new(file_opts)]`

I have a few questions about that:

1. Why the `[]`s here? I haven't found much about a type's `generate` method in the puppet type development documentation.

2. What is the relationship between `Puppet::Type::File` and `Puppet::Type.type(:file)` ?

3. Am I correct that the type's `generate` method is called at catalogue compile time, while the provider's `create` method is called when the resource is actually evaluated? So nothing in `generate` actually creates the file on disk because this is the type not the provider; instead generate creates a resource in the catalogue which has the same basic properties and behavior as a File resource?

lib/puppet/type/keystore.rb
```diff
@@ -0,0 +1,55 @@
+Puppet::Type.newtype(:keystore) do
+  desc 'generates an empty pkcs12 keystore'
```
wbclark, line 2: Should this be?

```suggestion
  @doc = 'generates an empty pkcs12 keystore'
```

I see it done this way in the puppet type development documentation (while the parameters still use `desc`). What is the difference?

lib/puppet/type/keystore.rb
wbclark: It turns out this is extremely similar to the example of autorequires in the type development documentation.

The one difference called out there: "An important difference between automatic and explicit relationships is that automatic relationships do not require the other resources to exist, while explicit relationships do."

ekohl: The pure Puppet equivalent is more like a [collector](https://puppet.com/docs/puppet/7/lang_collectors.html). Untested (so there may be typos) but roughly:
```puppet
keystore { $some_keystore:
  password_file => $some_password_file_path,
}
File <| title == $some_password_file_path |> -> Keystore[$some_keystore]
```

ekohl:
Just throwing it out there, since some meta discussion about types has started. There's also [resource_api](https://github.com/puppetlabs/puppet-resource_api) that's included in Puppet 6 (in earlier versions you could install the gem but I didn't want to deal with that). https://puppet.com/docs/puppet/7/create_types_and_providers_resource_api.html#writing_the_type_and_provider describes how to use it. Perhaps that's easier to use. If we're moving to Puppet 6 anyway, it may be easier to use.

There are [limitations](https://puppet.com/docs/puppet/7/about_the_resource_api.html#resource_api_limitations) that the low level API doesn't have. It's very likely we can't use `generate`.

>     1. Why the `[]`s here? I haven't found much about a type's `generate` method in the puppet type development documentation.

It is indeed a well hidden gem. You can generate a number of resources. Another way to think about it is to say that the return type of `generate` is `Array<Puppet::Type>`.

>     2\. What is the relationship between `Puppet::Type::File` and `Puppet::Type.type(:file)` ?

`Puppet::Type.type` provides a nice frontend. You don't have to know the exact resource as it's implemented, you just get the type name as you would also use in a Puppet manifest.

>     3\. Am I correct that the type's `generate` method is called at catalogue compile time, while the provider's `create` method is called when the resource is actually evaluated? So nothing in `generate` actually creates the file on disk because this is the type not the provider; instead generate creates a resource in the catalogue which has the same basic properties and behavior as a File resource?

I'm not sure where exactly it's called. I suspect it's the reason we see the number of resources go up during application (as you can see in the installer's output). We copied this trick from puppetlabs-concat so it's used quite a lot in the various modules we apply.

lib/puppet/type/keystore.rb
ehelms: Attempting to quickly look through the code, I think this is just a matter of using a method vs. setting an instance variable:

https://github.com/puppetlabs/puppet/blob/67ad21b64237eb4f808844bdaa3c2b0e4cdef95c/lib/puppet/util/docs.rb#L4-L6

Using the `desc` method seems more robust from an API stand-point. And all our other types do this as well so there is consistency in use.

ehelms:
> Just throwing it out there, since some meta discussion about types has started. There's also [resource_api](https://github.com/puppetlabs/puppet-resource_api) that's included in Puppet 6 (in earlier versions you could install the gem but I didn't want to deal with that). https://puppet.com/docs/puppet/7/create_types_and_providers_resource_api.html#writing_the_type_and_provider describes how to use it. Perhaps that's easier to use. If we're moving to Puppet 6 anyway, it may be easier to use.

I have considered using the newer, "easier" declaration methods for types. What I came back to was that for us a project it would be better to use consistent methods for building types and providers as that allows developers to connect the dots between providers and types we write and more easily understand what is happening. Thus I stuck with patterns already present in this code base and others we wrote (e.g. https://github.com/theforeman/puppet-foreman_proxy_content/tree/master/lib/puppet).

lib/puppet/type/keystore.rb
wbclark: OK, thanks for clearing that up. The method is there to be used, so let's use it then.

lib/puppet/type/keystore.rb
```diff
@@ -0,0 +1,55 @@
...
+    desc "Specifies the permissions mode of the keystore. Valid options: a string containing a permission mode value in octal notation."
+  end
+
+  autorequire(:file) do
+    [self[:password_file]]
+  end
+
+  def generate
+    file_opts = {
+      ensure: (self[:ensure] == :absent) ? :absent : :file,
```
wbclark, line 34: I thought that this opens a possibility where someone declares `ensure => 'purged',` (for example) and this ultimately (and incorrectly) gets interpreted as `:file`.

But, trying to reproduce the potential problem I see that I am prevented from making that mistake: `Error: Parameter ensure failed on Keystore[/etc/candlepin/certs/keystore]: Invalid value "purged". Valid values are present, absent. (file: /etc/puppetlabs/code/environments/production/modules/certs/manifests/candlepin.pp, line: 129)`

I then found https://github.com/puppetlabs/puppet/blob/67ad21b64237eb4f808844bdaa3c2b0e4cdef95c/lib/puppet/type.rb#L197 and https://github.com/puppetlabs/puppet/blob/67ad21b64237eb4f808844bdaa3c2b0e4cdef95c/lib/puppet/property/ensure.rb#L13-L30 so I am fully satisfied and understand the logic here.

manifests/candlepin.pp
```diff
@@ -131,6 +126,14 @@
       key_mode    => '0440',
     }
 
+    keystore { $keystore:
+      ensure        => present,
```
wbclark, line 130:
```suggestion
      ensure        => 'present',
```

NIT: I've previously received feedback that it's preferred to avoid naked strings

wbclark (approved):
This works great in my testing and is a significant step forward in flexibility.

In the future I'd like to see unit tests perhaps, but for the time being I'm satisfied that this should work well for our purposes given that the acceptance tests remain working. :+1:

ehelms:
I'm going to merge to unblock two other PRs with new types and in parallel work on adding some stand-alone tests for the types as that will be needed additionally for some of the already merged types and pay off if we move types to be used in other modules.

manifests/candlepin.pp
ekohl: There's some edge cases, like `present` and `installed` for `ensure`. I those are very common. I'm not sure we need to start converting those.

lib/puppet/type/keystore.rb
```diff
@@ -0,0 +1,55 @@
...
+    [:owner,
+     :group,
+     :mode].each do |param|
+      file_opts[param] = self[param] unless self[param].nil?
+    end
+
+    metaparams = Puppet::Type.metaparams
+    excluded_metaparams = [:before, :notify, :require, :subscribe, :tag]
+
+    metaparams.reject! { |param| excluded_metaparams.include? param }
```
ekohl, line 47: I think changing metaparams in place may be risky. I'd avoid it.

PR #335, opened 2021-05-20
Fixes #32631: Add keystore_certificate provider type

ehelms:
<del>Requires https://github.com/theforeman/puppet-certs/pull/334</del>

### DIFF
--- a/lib/puppet/provider/keystore_certificate/openssl.rb
+++ b/lib/puppet/provider/keystore_certificate/openssl.rb
@@ -1,0 +1,95 @@
+Puppet::Type.type(:keystore_certificate).provide(:openssl) do
+  commands :openssl => 'openssl'
+  commands :keytool => 'keytool'
+
+  def create
+    add_certificate
+  end
+
+  def destroy
+    delete_certificate
+  end
+
+  def exists?
+    keystore_content
+  end
+
+  def certificate
+    keystore_fingerprint
+  end
+
+  def certificate=(value)
+    delete_certificate unless keystore_content.nil?
+    add_certificate
+  end
+
+  def fingerprint(file)
+    return unless File.exist?(file)
+
+    openssl('x509', '-sha256', '-noout', '-fingerprint', '-in', file).strip.split('=')[1]
+  end
+
+  private
+
+  def add_certificate
+    Tempfile.open('temp_keystore') do |temp_store|
+      openssl(
+        'pkcs12',
+        '-export',
+        '-in', resource[:certificate],
+        '-inkey', resource[:private_key],
+        '-out', temp_store,
+        '-name', resource[:alias],
+        '-CAfile', resource[:ca],
+        '-password', "file:#{resource[:password_file]}"
+      )
+
+      keytool(
+        '-importkeystore',
+        '-noprompt',
+        '-srckeystore', temp_store,
+        '-srcstorepass:file', resource[:password_file],
+        '-destkeystore', resource[:keystore],
+        '-deststorepass:file', resource[:password_file],
+        '-srcalias', resource[:alias],
+        '-destalias', resource[:alias]
+      )
+    end
+  rescue Puppet::ExecutionFailure => e
+    Puppet.err("Failed to add certificate to keystore: #{e}")
+    nil
+  end
+
+  def delete_certificate
+    keytool(
+      '-delete',
+      '-noprompt',
+      '-keystore', resource[:keystore],
+      '-alias', resource[:alias],
+      '-storepass:file', resource[:password_file]
+    )
+  end
+
+  def keystore_content
+    return unless file_readable?(resource[:keystore])
+    return unless file_readable?(resource[:password_file])
+
+    keytool(
+      '-list',
+      '-keystore', resource[:keystore],
+      '-storepass:file', resource[:password_file],
+      '-alias', resource[:alias]
+    )
+  rescue Puppet::ExecutionFailure => e
+    Puppet.debug("Failed to read keystore contents: #{e}")
+    nil
+  end
+
+  def keystore_fingerprint
+    keystore_content&.scan(/^Certificate fingerprint \(SHA-256\): (.+)$/)&.first&.first
+  end
+
+  def file_readable?(file)
+    File.file?(file) && File.readable?(file)
+  end
+end
--- a/lib/puppet/type/keystore_certificate.rb
+++ b/lib/puppet/type/keystore_certificate.rb
@@ -1,0 +1,67 @@
+Puppet::Type.newtype(:keystore_certificate) do
+  desc 'adds a certificate and private key to a PKCS12 keystore'
+
+  ensurable
+
+  def self.title_patterns
+    [
+      [
+        /(.+):(.+)/,
+        [[:keystore], [:alias]]
+      ],
+      [
+        /(.+)/,
+        [[:alias]],
+      ],
+    ]
+  end
+
+  newparam(:alias, :namevar => true) do
+    desc "The certificate alias used to store inside the keystore"
+  end
+
+  newparam(:keystore, :namevar => true) do
+    desc "Path to the keystore to use when importing the certificate"
+    isrequired
+  end
+
+  newproperty(:certificate) do
+    desc "Path to the file containing the certificate to add to the keystore"
+
+    def fingerprint(file)
+      provider.fingerprint(file)
+    end
+
+    def should_to_s(newvalue)
+      self.class.format_value_for_display(fingerprint(newvalue))
+    end
+
+    def insync?(is)
+      is == fingerprint(should)
+    end
+  end
+
+  newparam(:private_key) do
+    desc "Path to file containing the private key"
+    isrequired
+  end
+
+  newparam(:ca) do
+    desc "Path to the file containing the CA certificate"
+  end
+
+  newparam(:password_file) do
+    desc "Path to file containing the keystore password"
+    isrequired
+  end
+
+  autorequire(:file) do
+    [
+      self[:password_file],
+      self[:certificate],
+      self[:private_key],
+      self[:ca],
+      self[:keystore]
+    ]
+  end
+end
--- a/manifests/candlepin.pp
+++ b/manifests/candlepin.pp
@@ -68,8 +68,8 @@ class certs::candlepin (
 
   $keystore_password = extlib::cache_data('foreman_cache_data', $keystore_password_file, extlib::random_password(32))
   $truststore_password = extlib::cache_data('foreman_cache_data', $truststore_password_file, extlib::random_password(32))
-  $keystore_password_path = "${pki_dir}/keystore_password-file"
-  $truststore_password_path = "${pki_dir}/truststore_password-file"
+  $keystore_password_path = "${pki_dir}/${keystore_password_file}"
+  $truststore_password_path = "${pki_dir}/${truststore_password_file}"
   $client_req = "${pki_dir}/java-client.req"
   $client_cert = "${pki_dir}/certs/${java_client_cert_name}.crt"
   $client_key = "${pki_dir}/private/${java_client_cert_name}.key"
@@ -91,12 +91,36 @@ class certs::candlepin (
       unprotect     => true,
       strip         => true,
       password_file => $ca_key_password_file,
-    } ~>
+    }
+
     certs::keypair { 'tomcat':
-      key_pair  => Cert[$tomcat_cert_name],
-      key_file  => $tomcat_key,
-      cert_file => $tomcat_cert,
-    } ~>
+      key_pair    => Cert[$tomcat_cert_name],
+      key_file    => $tomcat_key,
+      manage_key  => true,
+      key_owner   => 'root',
+      key_group   => $client_keypair_group,
+      key_mode    => '0440',
+      cert_file   => $tomcat_cert,
+      manage_cert => true,
+      cert_owner  => 'root',
+      cert_group  => $client_keypair_group,
+      cert_mode   => '0440',
+    }
+
+    certs::keypair { 'candlepin':
+      key_pair    => Cert[$java_client_cert_name],
+      key_file    => $client_key,
+      cert_file   => $client_cert,
+      manage_cert => true,
+      cert_owner  => 'root',
+      cert_group  => $client_keypair_group,
+      cert_mode   => '0440',
+      manage_key  => true,
+      key_owner   => 'root',
+      key_group   => $client_keypair_group,
+      key_mode    => '0440',
+    }
+
     file { $keystore_password_path:
       ensure    => file,
       content   => $keystore_password,
@@ -104,26 +128,6 @@ class certs::candlepin (
       group     => $group,
       mode      => '0440',
       show_diff => false,
-    } ~>
-    exec { 'candlepin-generate-ssl-keystore':
-      command   => "openssl pkcs12 -export -in ${tomcat_cert} -inkey ${tomcat_key} -out ${keystore} -name tomcat -CAfile ${ca_cert} -caname root -password \"file:${keystore_password_path}\"",
-      unless    => "keytool -list -keystore ${keystore} -storepass:file ${keystore_password_path} -alias tomcat | grep $(openssl x509 -noout -fingerprint -sha256 -in ${tomcat_cert} | cut -d '=' -f 2)",
-      logoutput => 'on_failure',
-      path      => ['/bin/', '/usr/bin'],
-      require   => Keystore[$keystore],
-    } ~>
-    certs::keypair { 'candlepin':
-      key_pair    => Cert[$java_client_cert_name],
-      key_file    => $client_key,
-      cert_file   => $client_cert,
-      manage_cert => true,
-      cert_owner  => $user,
-      cert_group  => $client_keypair_group,
-      cert_mode   => '0440',
-      manage_key  => true,
-      key_owner   => $user,
-      key_group   => $client_keypair_group,
-      key_mode    => '0440',
     }
 
     keystore { $keystore:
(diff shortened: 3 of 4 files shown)


lib/puppet/provider/keystore/keytool.rb
```diff
@@ -0,0 +1,49 @@
...
+      keytool(
+        '-genkey',
+        '-storetype', 'pkcs12',
+        '-keystore', resource[:keystore],
+        '-storepass:file', resource[:password_file],
+        '-alias', temp_alias,
+        '-dname', "CN=#{temp_alias}"
+      )
+    rescue Puppet::ExecutionFailure => e
+      Puppet.error("Failed to generate new keystore with temporary entry: #{e}")
```
ekohl, line 31: Does error halt execution? It looks like it'll then proceed with attempting to delete the entry, even if it failed.

ehelms: Addressed in the origin PR -- https://github.com/theforeman/puppet-certs/pull/334/files#diff-ab1ff29ab535e60c2e133829981732e7bdc416e5eabf07703b5b32e0ecafda51R18

manifests/candlepin.pp
```diff
@@ -130,6 +121,30 @@
...
+
+    keystore { $keystore:
+      ensure        => present,
+      password_file => $keystore_password_path,
+      owner         => 'root',
+      group         => $group,
+      mode          => '0640',
+    }
+
+    keystore_certificate { "${keystore}:tomcat":
```
evgeni, line 140: I must admit I find the notion of `$keystore:$alias` for the name rather confusing. Is this a common notation in the Java/keytool universe?

I see you could also write
```
keystore_certificate { "funny_name":
  keystore => $keystore,
  alias     => 'tomcat'
  …
}
```

And I find this more readable. But wouldn't block this on my weird brain :)

manifests/candlepin.pp
```diff
@@ -130,6 +121,30 @@
       key_mode    => '0440',
     }
 
+    file { $keystore_password_path:
+      ensure  => file,
+      content => $keystore_password,
```
evgeni, line 126: unrelated to this PR (as this is old code), but does this leak the password to the logs?

lib/puppet/type/keystore_certificate.rb
```diff
@@ -0,0 +1,62 @@
...
+      self.class.format_value_for_display(fingerprint(newvalue))
+    end
+
+    def insync?(is)
+      is == fingerprint(should)
+    end
+  end
+
+  newparam(:private_key) do
+    desc "Path to file containing the keystore password"
```
evgeni, line 36:
```suggestion
    desc "Path to file containing the private key"
```

lib/puppet/type/keystore_certificate.rb
```diff
@@ -0,0 +1,62 @@
...
+    isrequired
+  end
+
+  newparam(:ca_file) do
+    desc "Path to the CA certificate"
```
evgeni, line 41:
```suggestion
    desc "Path to the file containing th CA certificate"
```

(to be consistent with the other param docs)

manifests/candlepin.pp
```diff
@@ -130,6 +121,30 @@
...
+      group         => $group,
+      mode          => '0640',
+    }
+
+    keystore_certificate { "${keystore}:tomcat":
+      ensure        => present,
+      password_file => $keystore_password_path,
+      certificate   => $tomcat_cert,
+      private_key   => $tomcat_key,
+      ca_file       => $ca_cert,
```
evgeni, line 145: should *all* these be called `*_file`, as they are all… files?

manifests/candlepin.pp
ekohl: It's more of a Puppet convention. I think that what you write will also work with the current code. The benefit of the current version is that you can write `tomcat` into multiple keystores without having duplicate resources, so this works:
```puppet
keystore_certificate { "${keystore1}:tomcat":
}
keystore_certificate { "${keystore2}:tomcat":
}
```

If you don't, you need to use things like `funny_name1` and `funny_name2`. There are arguments for both.

That said, it's probably good to write a test case for both syntaxes to verify it really does work for both versions.

evgeni: Yeah, I also read the code that both notations *should* work.

manifests/candlepin.pp
ehelms: Solid question -- https://github.com/theforeman/puppet-certs/pull/337

manifests/candlepin.pp
ehelms: I have seen in other modules related to certs not attaching `_file` so I lean towards going with that model and it feels tidy.

ehelms:
Rebased and updated based on comments

lib/puppet/provider/keystore_certificate/openssl.rb
```diff
@@ -0,0 +1,95 @@
...
+  def create
+    add_certificate
+  end
+
+  def destroy
+    delete_certificate
+  end
+
+  def exists?
+    keystore_content
```
ekohl, line 14: I don't think this is a valid thing. I'd expect that `keystore_content` always returns a string which evaluates to true. Or does it rely on it raising a `Puppet::ExecutionFailure`if the alias isn't present?

lib/puppet/type/keystore_certificate.rb
```diff
@@ -0,0 +1,62 @@
...
+      self[:certificate],
+      self[:private_key],
+      self[:ca],
+      self[:keystore]
+    ]
+  end
+
+  autonotify(:file) do
+    [self[:keystore]]
+  end
```
ekohl, line 61: Is this correct? Shouldn't the keystore be created before the certificate is added? Otherwise you risk incorrect permissions which can leave the keystore world readable. Probably not a huge deal since we still use password files but still.

manifests/candlepin.pp
```diff
@@ -91,27 +91,22 @@
...
+      key_file    => $tomcat_key,
+      manage_key  => true,
+      key_owner   => $user,
+      key_group   => $client_keypair_group,
+      key_mode    => '0440',
+      cert_file   => $tomcat_cert,
+      manage_cert => true,
+      cert_owner  => $user,
+      cert_group  => $client_keypair_group,
+      cert_mode   => '0440',
```
ekohl, line 107: Did we not manage these before? Did that leave the key world readable?

ehelms: The `privkey` provider ensures a sane default if we do not manage the key through a file resource:

https://github.com/theforeman/puppet-certs/blob/master/lib/puppet/provider/privkey/katello_ssl_tool.rb#L33

```
# ls -la /etc/pki/katello/private/katello-tomcat.key 
-r--------. 1 root root 3243 May 21 13:53 /etc/pki/katello/private/katello-tomcat.key
```

lib/puppet/provider/keystore_certificate/openssl.rb
ehelms: The latter. We have two existence conditions currently:

1) Does the specified alias exist in the specified keystore
2) Does the specified certificate fingerprint match what is in the keystore

lib/puppet/type/keystore_certificate.rb
ehelms: Yea, this was a leftover from the previous implementation that I didn't question until you raised it. I will drop it.

lib/puppet/provider/keystore_certificate/openssl.rb
```diff
@@ -0,0 +1,95 @@
...
+        '-destkeystore', resource[:keystore],
+        '-deststorepass:file', resource[:password_file],
+        '-srcalias', resource[:alias],
+        '-destalias', resource[:alias]
+      )
+    end
+  rescue Puppet::ExecutionFailure => e
+    Puppet.debug("Failed to add certificate to keystore: #{e}")
+    nil
+  end
```
wbclark, line 61:
```suggestion
  def add_certificate
    Tempfile.open('temp_keystore') do |temp_store|
      openssl(
        'pkcs12',
        '-export',
        '-in', resource[:certificate],
        '-inkey', resource[:private_key],
        '-out', temp_store,
        '-name', resource[:alias],
        '-CAfile', resource[:ca],
        '-password', "file:#{resource[:password_file]}"
      )
      rescue Puppet::ExecutionFailure => e
        Puppet.err("Failed to export certificate to PKCS12: #{e}")
        return
      end

      keytool(
        '-importkeystore',
        '-noprompt',
        '-srckeystore', temp_store,
        '-srcstorepass:file', resource[:password_file],
        '-destkeystore', resource[:keystore],
        '-deststorepass:file', resource[:password_file],
        '-srcalias', resource[:alias],
        '-destalias', resource[:alias]
      )
      rescue Puppet::ExecutionFailure => e
        Puppet.err("Failed to add certificate to keystore: #{e}")
        return
      end
    end
  end
```

This changes:

1. `def add_certificate` was missing `end`

2. Log issues here at error level

3. Move `rescue` blocks into method, and `return`

(I haven't yet tested this)

wbclark: I don't this is quite right. I see now the intended structure was

```
def add_certificate
  Tempfile.open('temp_keystore') do |temp_store|
    openssl(...)
    keytool(...)
  rescue
    ...
  end
end
```

ehelms: I can make the change to error level. The two operations are "atomic" in their execution given the Temp file that is being created. The code gets kinda of ugly attempting to rescue multiple places and ensure the Temp file is cleaned up in each and every case. The failure being caught for either invocation would trigger the rescue and printing useful information to the user. That is why I chose not to split it up.

ekohl (commented):
Small nits, otherwise :+1:

manifests/candlepin.pp
```diff
@@ -91,27 +91,22 @@
...
-      command   => "openssl pkcs12 -export -in ${tomcat_cert} -inkey ${tomcat_key} -out ${keystore} -name tomcat -CAfile ${ca_cert} -caname root -password \"file:${keystore_password_path}\"",
-      unless    => "keytool -list -keystore ${keystore} -storepass:file ${keystore_password_path} -alias tomcat | grep $(openssl x509 -noout -fingerprint -sha256 -in ${tomcat_cert} | cut -d '=' -f 2)",
-      logoutput => 'on_failure',
-      path      => ['/bin/', '/usr/bin'],
-      require   => Keystore[$keystore],
-    } ~>
+      key_pair    => Cert[$tomcat_cert_name],
+      key_file    => $tomcat_key,
+      manage_key  => true,
+      key_owner   => $user,
```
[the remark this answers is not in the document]
ekohl: Looks like you modified one that I didn't spot, but not this particular instance ;)

lib/puppet/provider/keystore_certificate/openssl.rb
```diff
@@ -0,0 +1,95 @@
...
+      '-storepass:file', resource[:password_file],
+      '-alias', resource[:alias]
+    )
+  rescue Puppet::ExecutionFailure => e
+    Puppet.debug("Failed to read keystore contents: #{e}")
+    nil
+  end
+
+  def keystore_fingerprint
+    keystore_content&.scan(/^Certificate fingerprint \(SHA-256\): (.+)$/)&.first&.first
```
wbclark, line 89: After a bit of testing, I understand that the second `&.first` is necessary because otherwise idempotence is broken:

```
Notice: /Stage[main]/Certs::Candlepin/Keystore_certificate[/etc/candlepin/certs/keystore:tomcat]/certificate: certificate changed ['FD:64:E2:53:2D:2D:30:AC:8D:C7:9A:D4:9C:D8:DD:D1:3D:09:98:B5:41:9D:F0:ED:66:9F:40:C2:7B:13:7C:7E'] to 'FD:64:E2:53:2D:2D:30:AC:8D:C7:9A:D4:9C:D8:DD:D1:3D:09:98:B5:41:9D:F0:ED:66:9F:40:C2:7B:13:7C:7E'
```

I had a concern about multiple certificates in the keystore, but we are specifying the `-alias` so we should get only the one we care about

manifests/candlepin.pp
wbclark: I tested this, but:

```
  Error: Evaluation Error: Error while evaluating a Resource Statement, No set of title patterns matched the title "funny_name". (file: /etc/puppetlabs/code/environments/production/modules/certs/manifests/candlepin.pp, line: 140, column: 5) on node centos7-64-1
```

My code was:

```puppet
    keystore_certificate { "funny_name":
      keystore      => $keystore,
      alias         => 'tomcat',
      ensure        => present,
      password_file => $keystore_password_path,
      certificate   => $tomcat_cert,
      private_key   => $tomcat_key,
      ca            => $ca_cert,
    }
```

lib/puppet/provider/keystore_certificate/openssl.rb
ehelms: That's the idea.

manifests/candlepin.pp
ehelms: I've updated this to support passing:

  1) either just `keystore` in which case the name is the `alias`
  2) `keystore:alias`
  3) `keystore` and `alias` parameters with a random name

My workflow is adding some stand alone provider tests in follow up PRs so I will handle testing some of that in there.

ehelms:
@wbclark @evgeni Been a wee bit, y'all reviewed this originally can you come back to it?

lib/puppet/type/keystore_certificate.rb
```diff
@@ -0,0 +1,67 @@
+Puppet::Type.newtype(:keystore_certificate) do
+  desc 'adds a certificate and private key to a PKCS12 keystore'
+
+  ensurable
+
+  def self.title_patterns
+    [
+      [
+        /(.+):(.+)/m,
```
wbclark, line 9: Do we expect the possibility of multi-line resource titles?

Further  below (L13) do do we intend to map multi-line resource titles to an alias only?

Some elaboration on the choice of `/.../m` would be helpful to review this

ehelms: I should know this but what is a multi-line resource title?

ehelms: I did it this way to support `/path/to/keystore:alias` and `alias` as titles.

wbclark: I'm reading at https://ruby-doc.org/core-2.4.1/Regexp.html#EXTENDED

"/pat/m - Treat a newline as a character matched by ."

How does it relate to `/path/to/keystore:alias` ? There is some problem with the `/` characters in the path?

ehelms: I didn't much think about it. I copied this pattern from the `truststore_certificate` we had built previously, and that originates from https://github.com/theforeman/puppet-certs/pull/312

manifests/candlepin.pp
ekohl: Marking as unresolved. While that PR was merged, this now drops the `show_diff` again.

manifests/candlepin.pp
ekohl: Can we mark this as resolved now?

wbclark (commented):
We have both a `$keystore_password_file` and `$keystore_password_path` in manifests/candlepin.pp

The `$keystore_password_file` is only used in `$keystore_password = extlib::cache_data('foreman_cache_data', $keystore_password_file, extlib::random_password(32))`

Is that intended as a transitionary design to move the location while keeping the same content, or is it simply an oversight?

ehelms:
> We have both a `$keystore_password_file` and `$keystore_password_path` in manifests/candlepin.pp
> 
> The `$keystore_password_file` is only used in `$keystore_password = extlib::cache_data('foreman_cache_data', $keystore_password_file, extlib::random_password(32))`
> 
> Is that intended as a transitionary design to move the location while keeping the same content, or is it simply an oversight?

When using `cache_data` you are almost doing a key-value scenario where you give it a filename (key) and the password (value) you want to store by that name in the caching location. So you end up with something like:

```
/opt/puppetlabs/puppet/cache/foreman_cache_data/keystore_password-file
```

If you were instead to pass it `$keystore_password_path`, you would end up with:

```
/opt/puppetlabs/puppet/cache/foreman_cache_data/etc/pki/katello/keystore_password-file
```

evgeni:
Doesn't that mean we can't have two keystores at `/etc/pki/katello/keystore` and `/etc/pki/candlepin/keystore` (stupid example, but two stores with the same `basemame`)?

ehelms:
> Doesn't that mean we can't have two keystores at `/etc/pki/katello/keystore` and `/etc/pki/candlepin/keystore` (stupid example, but two stores with the same `basemame`)?

Nothing prevents this. If this isn reference to the `cache_data`, you just have to ensure the cache key is different for each keystores password or that you change the namespace. For example, if `/etc/pki/candlepin/keystore` was instead defined in `puppet-candlepin` you might should opt to choose `candlepin_cache_data` as your namespace.

wbclark:
Got it, thanks.

Instead of:

```puppet
  $keystore_password = extlib::cache_data('foreman_cache_data', $keystore_password_file, extlib::random_password(32))
  $truststore_password = extlib::cache_data('foreman_cache_data', $truststore_password_file, extlib::random_password(32))    
  $keystore_password_path = "${pki_dir}/keystore_password-file"
  $truststore_password_path = "${pki_dir/truststore_password-file"
```

I believe this would be a cleaner alternative:

```puppet
  $keystore_password = extlib::cache_data('foreman_cache_data', $keystore_password_file, extlib::random_password(32))
  $truststore_password = extlib::cache_data('foreman_cache_data', $truststore_password_file, extlib::random_password(32))    
  $keystore_password_path = "${pki_dir}/${keystore_password_file}"
  $truststore_password_path = "${pki_dir/${truststore_password_file}"
```

So at least the path variable would be built from the file variable.

ehelms:
Oh sure I can close that previous gap in how things were defined. Updated.

ehelms:
I believe I have now addressed everything. Could I get another review?

ekohl (approved):
I'd also consider checking file paths in the type are absolute paths, but I won't block on that.

wbclark (approved):
I'd like to better understand the implementation choices in the `title_patterns` but since it was copied from an already working type and tests validate that it's working here as well, I think I can treat that as my own homework -- starting with adding test cases for the various title patterns. I don't see that as something that needs to block this work, as the primary case that we care about is already validated.

PR #336, opened 2021-05-20
Fixes #32637: Add truststore type and provider 

ehelms:
Builds on #334 re-factoring what is present there into a common module.

I did not want to create one giant PR with all of these types but rather a set of progressive PRs that build one on each other to make it easier to review, discuss and ultimately to merge in smaller logical pieces.

### DIFF
--- a/lib/puppet/provider/keystore/keytool.rb
+++ b/lib/puppet/provider/keystore/keytool.rb
@@ -1,51 +1,7 @@
+require 'puppet_x/certs/provider/keystore'
+
 Puppet::Type.type(:keystore).provide(:keytool) do
   commands :keytool => 'keytool'
 
-  def create
-    generate_keystore
-  end
-
-  def destroy
-    delete_keystore
-  end
-
-  def exists?
-    File.exist?(resource[:keystore])
-  end
-
-  private
-
-  def generate_keystore
-    temp_alias = 'temporary-entry'
-
-    begin
-      keytool(
-        '-genkey',
-        '-storetype', 'pkcs12',
-        '-keystore', resource[:keystore],
-        '-storepass:file', resource[:password_file],
-        '-alias', temp_alias,
-        '-dname', "CN=#{temp_alias}"
-      )
-    rescue Puppet::ExecutionFailure => e
-      Puppet.err("Failed to generate new keystore with temporary entry: #{e}")
-      return nil
-    end
-
-    begin
-      keytool(
-        '-delete',
-        '-keystore', resource[:keystore],
-        '-storepass:file', resource[:password_file],
-        '-alias', temp_alias
-      )
-    rescue Puppet::ExecutionFailure => e
-      Puppet.err("Failed to delete temporary entry when generating empty keystore: #{e}")
-      return nil
-    end
-  end
-
-  def delete_keystore
-    File.rm(resource[:keystore])
-  end
+  include Puppet_X::Certs::Provider::Keystore
 end
--- a/lib/puppet/provider/truststore/keytool.rb
+++ b/lib/puppet/provider/truststore/keytool.rb
@@ -1,0 +1,15 @@
+require 'puppet_x/certs/provider/keystore'
+
+Puppet::Type.type(:truststore).provide(:keytool) do
+  commands :keytool => 'keytool'
+
+  include Puppet_X::Certs::Provider::Keystore
+
+  def store
+    resource[:truststore]
+  end
+
+  def type
+    'truststore'
+  end
+end
--- a/lib/puppet/type/truststore.rb
+++ b/lib/puppet/type/truststore.rb
@@ -1,0 +1,54 @@
+Puppet::Type.newtype(:truststore) do
+  desc 'Generates an empty pkcs12 truststore'
+
+  ensurable
+
+  newparam(:truststore, :namevar => true) do
+    desc "Path to the truststore"
+    isrequired
+  end
+
+  newparam(:password_file) do
+    desc "Path to file containing the truststore password"
+    isrequired
+  end
+
+  newparam(:owner, parent: Puppet::Type::File::Owner) do
+    desc "Specifies the owner of the truststore. Valid options: a string containing a username or integer containing a uid."
+  end
+
+  newparam(:group, parent: Puppet::Type::File::Group) do
+    desc "Specifies a permissions group for the truststore. Valid options: a string containing a group name or integer containing a gid."
+  end
+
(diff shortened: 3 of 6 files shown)


lib/puppet/provider/truststore/keytool.rb
```diff
@@ -0,0 +1,13 @@
+Puppet::Type.type(:truststore).provide(:keytool) do
```
wbclark, line 1:
```suggestion
require 'puppet_x/certs/keystore'

Puppet::Type.type(:truststore).provide(:keytool) do
```

wbclark: I believe this was missed

wbclark (commented):
`lib/puppet/type/{key,trust}store.rb` are almost exactly identical

is an additional abstraction here possible, moving common code into a shared type, so that only the `newparam(:truststore, ...)` / `newparam(:keystore, ...)` and `path:` file_opts in the `generate` methods are declared differently for each respective type?

lib/puppet_x/certs/keystore.rb
```diff
@@ -0,0 +1,59 @@
+module Puppet_X
```
wbclark, line 1: I assume the `Puppet_X` name is a placeholder?

ehelms: This is a namespace to differentiate it from the Puppet namespace and prevent clashing. From all my research, that is the suggested way to provide your own common code.

lib/puppet/provider/truststore/keytool.rb
ehelms: Thanks, I will address after the underlying (the keystore) PR is merged.

ehelms:
> is an additional abstraction here possible, moving common code into a shared type, so that only the `newparam(:truststore, ...)` / `newparam(:keystore, ...)` and `path:` file_opts in the `generate` methods are declared differently for each respective type?

I think some of it can be, but also note a lot of the descriptions have truststore vs keystore in them and I don't want to write a bunch of substitution code that results in hard to read code. Given the type is an interface, I'd want to ensure it's as readable and understandable as possible.

ehelms:
Rebased with the merge of the keystore provider

lib/puppet/provider/keystore/keytool.rb
```diff
@@ -1,51 +1,7 @@
+require 'puppet_x/certs/keystore'
```
ekohl, line 1: It may be nice to differentiate this somehow so that it's clear that this is just for the provider.

lib/puppet/type/truststore.rb
```diff
@@ -0,0 +1,55 @@
...
+    end
+
+    metaparams = Puppet::Type.metaparams
+    excluded_metaparams = [:before, :notify, :require, :subscribe, :tag]
+
+    metaparams.reject! { |param| excluded_metaparams.include? param }
+
+    metaparams.each do |metaparam|
+      file_opts[metaparam] = self[metaparam] unless self[metaparam].nil?
+    end
```
ekohl, line 51: I think https://github.com/theforeman/puppet-certs/pull/341 should be applied to this as well.

lib/puppet_x/certs/keystore.rb
```diff
@@ -0,0 +1,59 @@
...
+          keytool(
+            '-genkey',
+            '-storetype', 'pkcs12',
+            '-keystore', store,
+            '-storepass:file', resource[:password_file],
+            '-alias', temp_alias,
+            '-dname', "CN=#{temp_alias}"
+          )
+        rescue Puppet::ExecutionFailure => e
+          Puppet.error("Failed to generate new #{type} with temporary entry: #{e}")
```
ekohl, line 37:
```suggestion
          Puppet.err("Failed to generate new #{type} with temporary entry: #{e}")
```

lib/puppet_x/certs/keystore.rb
```diff
@@ -0,0 +1,59 @@
...
+
+        begin
+          keytool(
+            '-delete',
+            '-keystore', store,
+            '-storepass:file', resource[:password_file],
+            '-alias', temp_alias
+          )
+        rescue Puppet::ExecutionFailure => e
+          Puppet.error("Failed to delete temporary entry when generating empty #{type}: #{e}")
```
ekohl, line 49:
```suggestion
          Puppet.err("Failed to delete temporary entry when generating empty #{type}: #{e}")
```

lib/puppet/provider/keystore/keytool.rb
ehelms: Done. I introduced another directory layer.

wbclark (commented):
`lib/puppet_x/certs/provider/keystore.rb` assumes that we have `commands :keytool => 'keytool'` (for example in the `generate_keystore` method)` yet that is only declared in the actual providers that `include Puppet_X::Certs::Provider::Keystore`.

The current design works and isn't breaking anything, but would it be better design to just move `commands :keytool => 'keytool'` into the shared module?

wbclark (approved):
Aside from the NIT (I wouldn't even call it that, more of a musing or question) I raised above, I don't have any further comments. LGTM

lib/puppet/type/truststore.rb
```diff
@@ -0,0 +1,54 @@
+Puppet::Type.newtype(:truststore) do
+  desc 'generates an empty pkcs12 truststore'
```
wbclark, line 2:
```suggestion
  desc 'Generates an empty pkcs12 truststore'
```

I suppose there is this nittiest of nits as well

ekohl: I wouldn't say that managing an empty keystore is really the point. It's more that it manages a pkcs12 keystore. It's ensurable so I'd assume it can also ensure it's really absent (not present on disk anymore). It can also manage (file) permissions on it.

ekohl (approved):
Other than the description, this looks good to me.

ehelms:
> `lib/puppet_x/certs/provider/keystore.rb` assumes that we have `commands :keytool => 'keytool'` (for example in the `generate_keystore` method) yet that is only declared in the actual providers that `include Puppet_X::Certs::Provider::Keystore`.
> 
> The current design works and isn't breaking anything, but would it be better design to just move `commands :keytool => 'keytool'` into the shared module?

The trick here is that `commands` is a method being called which does not work inside of a Ruby module. There are techniques if I recall for executing methods on inclusion but this clean to me right now.

PR #337, opened 2021-05-21
Fixes #32647: Do not show diff on password files

ehelms:


evgeni (approved):
I like how the diff of the "no diff" patch is super noisy thanks to whitespace.

ehelms:
> I like how the diff of the "no diff" patch is super noisy thanks to whitespace.

Here is your no noise diff -- https://github.com/theforeman/puppet-certs/pull/337/files?w=1

PR #341, opened 2021-05-25
Refs #32506 - Avoid in place modification of array

ekohl:
In 47db06c5fd01fba4e06af5325470c8c146074bba a new keystore provider was added. This modifies metaparams in place which can have side effects. This patch changes it to not need it.

Fixes: 47db06c5fd01fba4e06af5325470c8c146074bba

### DIFF
--- a/lib/puppet/type/keystore.rb
+++ b/lib/puppet/type/keystore.rb
@@ -41,13 +41,12 @@ Puppet::Type.newtype(:keystore) do
       file_opts[param] = self[param] unless self[param].nil?
     end
 
-    metaparams = Puppet::Type.metaparams
     excluded_metaparams = [:before, :notify, :require, :subscribe, :tag]
 
-    metaparams.reject! { |param| excluded_metaparams.include? param }
-
-    metaparams.each do |metaparam|
-      file_opts[metaparam] = self[metaparam] unless self[metaparam].nil?
+    Puppet::Type.metaparams.each do |metaparam|
+      unless self[metaparam].nil? || excluded_metaparams.include?(metaparam)
+        file_opts[metaparam] = self[metaparam]
+      end
     end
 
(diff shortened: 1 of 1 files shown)


wbclark:
> In [47db06c](https://github.com/theforeman/puppet-certs/commit/47db06c5fd01fba4e06af5325470c8c146074bba) a new keystore provider was added. This modifies metaparams in place which can have side effects. This patch changes it to not need it.
> 
> Fixes: [47db06c](https://github.com/theforeman/puppet-certs/commit/47db06c5fd01fba4e06af5325470c8c146074bba)

Thanks @ekohl . Could you elaborate a bit on the potential side effects of modifying metaparams in place?

ekohl:
What I would be worried about is that other code may also call `Puppet.type.metaparams`. Since it's modified in place, they get a different result than Puppet out of the box. That may mean notifications no longer happen and lead to very subtle bugs.

ehelms (approved):

ekohl:
> I think it could be a good community thing to file a bug against puppetlabs-concat so they can consider a similar change given that was the inspiration for this change (https://github.com/puppetlabs/puppetlabs-concat/blob/main/lib/puppet/type/concat_file.rb#L323).

I opened https://github.com/puppetlabs/puppetlabs-concat/pull/705

ehelms:
Awesome, thanks! I'm leaving this open for a day or so as @wbclark wanted to circle back and make sure he understood the why.

wbclark:
> What I would be worried about is that other code may also call `Puppet.type.metaparams`. Since it's modified in place, they get a different result than Puppet out of the box. That may mean notifications no longer happen and lead to very subtle bugs.

Thanks for the clear and concise explanation @ekohl . I now understand exactly what you are saying. Your foresight and experience brings great value to this project

wbclark (approved):

PR #343, opened 2021-05-26
Bump master to 13.0.0

ehelms:


wbclark (approved):
I checked commits between now and the 12.1.0 version bump, and I don't see any labeled as breaking - all are skip, bug, or enhancement. I deduce this is planned for future breaking changes that we know are coming and I recall that we discussed this.

The actual later release PR and changelogs would still allow us easily find breaking changes without worrying about the exact ordering relative to this particular commit.

ACK

ehelms:
Prior to merging this, updates to modules that depend on:

 * https://github.com/theforeman/puppet-katello/pull/415
 * https://github.com/theforeman/puppet-foreman_proxy_content/pull/360

ekohl:
As noted earlier, we may be able postpone it a bit by using `deploy => false` when calling it. If that's sufficient, we can do it without a breaking change. If we've transitioned every module, we can do a huge "deploy removal"-PR.

ehelms:
Since we intend to drop Puppet 5 (https://github.com/theforeman/puppet-certs/pull/345) we ought to accept this bump to the major version.

Which means first merging:

https://github.com/theforeman/puppet-katello/pull/415
https://github.com/theforeman/puppet-foreman_proxy_content/pull/360

PR #344, opened 2021-05-27
Fixes #32511 - Add a puppet type and provider to manage an nssdb

wbclark:


wbclark:
Currently facing an issue of circular dependencies.

If I understood correctly, the nssdb dir requires the nssdb resource, which requires the nssdb password file, which requires the nssdb dir

wbclark:
Fixed a few issues, still facing a few more. Going to get some rest and pick it back up tomorrow.

ehelms:
I think we have a few design choices we could make:

 1) nssdb type accepts a password and we create and manage a file to hold said password in the nssdb directory
 2) the password file is stored outside of the nssdb directory
 3) manage directory, password file and nssdb separately

I would prefer 1 or 2 I think based on the fact that the directory itself tends to be a "part" of the whole NSS database concept and you need to manage the directory the same as the files.

lib/puppet/provider/nssdb/certutil.rb
```diff
@@ -0,0 +1,44 @@
...
+    delete_nssdb
+  end
+
+  def exists?
+    files = resource[:nssdb_files]
+    files = [files] unless files.is_a?(Array)
+
+    files.each do |file|
+      return False unless File.exist?(file)
+    end
```
ehelms, line 18:
```suggestion
    files.all? { |file| File.exist?(file) }
```

You can use the `all?` check to see if every file matches some condition. If any one of them fails then it returns false.

manifests/ssltools/nssdb.pp
```diff
@@ -33,18 +33,13 @@
...
-    require => Package['nss-tools'],
-  } ->
-  file { $nssdb_files:
-    owner => 'root',
-    group => $group,
-    mode  => '0640',
+  }
+
+  nssdb { $nss_db_dir:
+    nssdb_files   => $nssdb_files,
```
ehelms, line 39: This should not be needed as this is an detail of the NSS database implementation.

lib/puppet/type/nssdb.rb
```diff
@@ -0,0 +1,62 @@
+Puppet::Type.newtype(:nssdb) do
+  desc "Generates an empty NSSDB"
```
ehelms, line 2:
```suggestion
  desc "Generates an empty NSS database"
```

lib/puppet/type/nssdb.rb
```diff
@@ -0,0 +1,62 @@
...
+
+  ensurable
+
+  newparam(:nssdb_dir, :namevar => true) do
+    desc "Path to NSS DB directory"
```
ehelms, line 7:
```suggestion
    desc "Path to NSS database directory"
```

lib/puppet/provider/nssdb/certutil.rb
```diff
@@ -0,0 +1,44 @@
...
+      certutil(
+        '-N',
+        '-d', resource[:nssdb_dir],
+        '-f', resource[:password_file]
+      )
+    rescue Puppet::ExecutionFailure => e
+      Puppet.err("Failed to generate new NSS DB at #{resource[:nssdb_dir]} with password file #{resource[:password_file]}: #{e}")
+      return nil
+    end
+  end
```
ehelms, line 34: In Ruby, if the entire function is being rescued you can avoid a begin and do:

```
def generate_nssdb
  .....
  ....
end
```

lib/puppet/provider/nssdb/certutil.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  def create
+    generate_nssdb
+  end
+
+  def destroy
+    delete_nssdb
+  end
+
+  def exists?
```
ehelms, line 12: Instead of checking that certain files exist (cause that has constraints as we see in todays code having ti know the full set of files per platform) this could use native methods in `certutil` to verify a working database exists:

```
# certutil -L -d /etc/pki/katello/nssdb/

Certificate Nickname                                         Trust Attributes
                                                             SSL,S/MIME,JAR/XPI

ca                                                           CT,C,c
broker                                                       u,u,u

# certutil -L -d /etc/pki/katello/
certutil: function failed: SEC_ERROR_LEGACY_DATABASE: The certificate/key database is in an old, unsupported format.
```

manifests/ssltools/nssdb.pp
wbclark: The `delete_nssdb` method in the provider also relies on this parameter currently.

lib/puppet/provider/nssdb/certutil.rb
wbclark: Ok, so I have the implementation

```ruby
  def exists?
    certutil(
      '-L',
      '-d', resource[:nssdb_dir]
    )
  end
```

I think it's not necessary to `rescue Puppet::ExecutionFailure` assuming the `commands` method handles exit codes the way I think it should (but I couldn't find that documented).

manifests/ssltools/nssdb.pp
ehelms: We can and should manage the owner/group/mode of the files inside the nssb directory but a user of the provider should not have to declare what those files are -- they are internal details, part of how nssdb works,  that a user should not have to care about. They are defined in the current implementation because there needed to be some way to make the creation idempotent. With the provider abstraction, you can do this more natively. For example, like https://github.com/theforeman/puppet-certs/pull/344#discussion_r640539099

lib/puppet/provider/nssdb/certutil.rb
ehelms: I would think you'd want the rescue to convert the error into a debug message and move on. The error generated if the NSS database does not exist should be indicator we need to create it not a blocking failure. I think this is shown in the test output:

```
  Error: /Stage[main]/Certs::Ssltools::Nssdb/Nssdb[/etc/pki/katello/nssdb]: Could not evaluate: Execution of '/usr/bin/certutil -L -d /etc/pki/katello/nssdb' returned 255: certutil: function failed: SEC_ERROR_LEGACY_DATABASE: The certificate/key database is in an old, unsupported format.
```

lib/puppet/provider/nssdb/certutil.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  def destroy_nssdb
+    files = resource[:nssdb_files]
+    files = [files] unless files.is_a?(Array)
+
+    files.each do |file|
+      File.rm(file)
+    end
+  end
```
wbclark, line 42: @ehelms you have requested an implementation of this method which does not require the user to pass the list of `nssdb_files` to the provider.

I'm not sure how you envision this, but did you have in mind performing an `rm -rf $nssdb_dir` ? What if the password file (as in the previous implementation) or some other files extra to the base NSS DB files are present in this directory (for example, backups)?

ehelms: If we are managing the nssdb directory (which I think we should since that is how it works) then anything *in* the directory should be considered managed by this resource and thus can be created, update and destroyed with the lifecycle of this resource. So if this type puts the password file in that directory then it should also remove it.

lib/puppet/type/nssdb.rb
```diff
@@ -0,0 +1,69 @@
...
+    end
+
+    resources = []
+    ensure_condition = self[:ensure] == :absent
+
+    dir_opts = file_opts.dup
+    dir_opts.merge({ :path => resource[:nssdb_dir] })
+    dir_opts.merge({ :mode => resource[:mode] }) unless self[:mode].nil?
+    dir_opts.merge({ :ensure => ensure_condition ? :absent : :directory })
+    dir_opts.merge({ :recurse => true })
```
wbclark, line 58: This is what I landed on for now to manage attributes for all generated files, but I forsee a few problems:

1. If the password_file exists without the nssdb_dir, this should cause a duplicate resource declaration error

2. I'm unable to manage the directory mode ('0755' in the previous implementation) separately from the file mode (0640). In particular, how is it even possible to set execute permission on the directory without setting it on all files within, when using `recurse => true` ?

A possible solution to these problems could be to use: https://forge.puppet.com/modules/npwalker/recursive_file_permissions

Your feedback and suggestions would be appreciated @ehelms @ekohl

wbclark: Unsure at the time if this is what is causing `Error: /Stage[main]/Certs::Ssltools::Nssdb/Nssdb[/etc/pki/katello/nssdb]: Failed to generate additional resources using 'generate': Title or name must be provided`... will do some more testing

ehelms: Given that the "nss database" is thought of as a directory, perhaps we should not try to manage the directory and Files in the directory separately user/group/permission wise. We should set them all the same.

And to that end, perhaps we should not put the password file in there either and thereby treat it like we do the other resources? For example, the file tree might be:

```
/etc/qpid/certs
/etc/qpid/certs/nssdb-password_file
/etc/qpid/certs/nssdb/key3.db
/etc/qpid/certs/nssdb/cert9.db
```

lib/puppet/provider/nssdb/certutil.rb
```diff
@@ -0,0 +1,72 @@
...
+
+  def exists?
+    nssdb_exists?
+  end
+
+  private
+
+  def create_nssdb_dir
+    mkdir('-p', resource[:nssdb_dir])
```
ehelms, line 26: You can let Ruby handle this for you:

```
FileUtils.mkdir_p
```

lib/puppet/provider/nssdb/certutil.rb
```diff
@@ -0,0 +1,72 @@
...
+      '-d', resource[:nssdb_dir],
+      '-f', resource[:password_file]
+    )
+  rescue Puppet::ExecutionFailure => e
+    Puppet.err("Failed to generate new NSS database at #{resource[:nssdb_dir]} with password file #{resource[:password_file]}: #{e}")
+    return nil
+  end
+
+  def destroy_password_file
+    FileUtils.rm_f resource[:password_file]
```
ehelms, line 55:
```suggestion
    FileUtils.rm_f(resource[:password_file])
```

While what you have is correct, I think we should keep to a single style of calling functions

lib/puppet/provider/nssdb/certutil.rb
```diff
@@ -0,0 +1,72 @@
...
+    Puppet.err("Failed to create NSS database directory at #{resource[:nssdb_dir]}: #{e}")
+    return nil
+  end
+
+  def generate_password_file
+    openssl(
+      'rand',
+      '-base64', '24',
+      '-out', resource[:password_file]
+    )
```
ehelms, line 37: Could we do this with Ruby instead of openssl?

ehelms: Second thought, with every other resource we have we generate a password or password file and then hand it to the resource (look at keystore for example). I think nssdb could (and should) be the same to not overload the type. 

If you look at my thought here (https://github.com/theforeman/puppet-certs/pull/344/files#r654092223) that is a way we could unify it such that NSS database works the same as keystore or truststore.

lib/puppet/provider/nssdb/certutil.rb
```diff
@@ -0,0 +1,72 @@
...
+    Puppet.err("Failed to generate new NSS database at #{resource[:nssdb_dir]} with password file #{resource[:password_file]}: #{e}")
+    return nil
+  end
+
+  def destroy_password_file
+    FileUtils.rm_f resource[:password_file]
+  end
+
+  def destroy_nssdb
+    FileUtils.rm_rf resource[:nssdb_dir], secure: true
```
ehelms, line 59: See https://github.com/theforeman/puppet-certs/pull/344/files#r654092797

lib/puppet/provider/nssdb/certutil.rb
```diff
@@ -0,0 +1,52 @@
+require 'fileutils'
+
+Puppet::Type.type(:nssdb).provide(:certutil) do
+  commands :certutil => 'certutil'
+  commands :openssl => 'openssl'
+  commands :mkdir => 'mkdir'
```
ehelms, line 6: This can be dropped now

manifests/ssltools/nssdb.pp
```diff
@@ -34,17 +24,10 @@
...
-    group   => $group,
-    creates => $nssdb_files,
-    require => Package['nss-tools'],
-  } ->
-  file { $nssdb_files:
-    owner => 'root',
-    group => $group,
-    mode  => '0640',
+  nssdb { $nss_db_dir:
+    password_file => $nss_db_password_file,
```
ehelms, line 28: You'll want an `ensure => present` for the first line of the resource since its an `ensurable` type.

manifests/ssltools/nssdb.pp
```diff
@@ -1,25 +1,15 @@
 # Sets up nssdb
 class certs::ssltools::nssdb (
-  $nss_db_dir = $certs::nss_db_dir,
```
ehelms, line 3: If we drop this we break the interface. I'd opt to keep it and introduce the `pki_dir` parameter so it's an additive change.

lib/puppet/type/nssdb.rb
wbclark: OK, latest update again removes the password file management from the provider/type, and moves the location into the parent dir.

manifests/ssltools/nssdb.pp
wbclark: I think the broken interface shouldn't actually break anything (except possibly tests, which I didn't check).

Since in manifests/qpid.pp we `include certs::ssltools::nssdb`

ehelms: Technically, for our use case that is true, but this is not marked as a private class. So you would be breaking a public ally available interface.

manifests/ssltools/nssdb.pp
```diff
@@ -34,17 +24,11 @@
     group  => $group,
     mode   => '0640',
   } ->
```
ehelms, line 26: You can drop this line as the `nssdb` autorequires the password_file

manifests/ssltools/nssdb.pp
wbclark: I think within the context of the rather sweeping changes we are making in this module, labeling this as a breaking change and releasing it all under a new major version would be OK. What do you think?

wbclark:
Remaining test failures are

```
Failures:

  1) certs with default params File "/etc/pki/katello/nssdb" is expected to be mode 640
     Failure/Error: it { should be_mode 640 }
       expected `File "/etc/pki/katello/nssdb".mode?(640)` to be truthy, got false
       
     # ./spec/acceptance/qpid_spec.rb:48:in `block (4 levels) in <top (required)>'
     # /usr/share/gems/gems/bundler-2.1.4/libexec/bundle:46:in `block in <top (required)>'
     # /usr/share/gems/gems/bundler-2.1.4/libexec/bundle:34:in `<top (required)>'

  2) certs with default params File "/etc/pki/katello/nssdb" is expected to be grouped into "qpidd"
     Failure/Error: it { should be_grouped_into 'qpidd' }
       expected `File "/etc/pki/katello/nssdb".grouped_into?("qpidd")` to be truthy, got false
       
     # ./spec/acceptance/qpid_spec.rb:50:in `block (4 levels) in <top (required)>'
     # /usr/share/gems/gems/bundler-2.1.4/libexec/bundle:46:in `block in <top (required)>'
     # /usr/share/gems/gems/bundler-2.1.4/libexec/bundle:34:in `<top (required)>'

Finished in 58.77 seconds (files took 1 minute 30.23 seconds to load)
31 examples, 2 failures

Failed examples:

rspec ./spec/acceptance/qpid_spec.rb:48 # certs with default params File "/etc/pki/katello/nssdb" is expected to be mode 640
rspec ./spec/acceptance/qpid_spec.rb:50 # certs with default params File "/etc/pki/katello/nssdb" is expected to be grouped into "qpidd"
```

wbclark:
Managing of file group owner and mode is definitely not working:

```
[root@centos8-64-1 ~]# ls -la /etc/pki/katello/nssdb
total 76
drwxr-xr-x. 2 root root    55 Jun 18 14:31 .
drwxr-xr-x. 5 root root    75 Jun 18 14:31 ..
-rw-------. 1 root root 36864 Jun 18 14:31 cert9.db
-rw-------. 1 root root 36864 Jun 18 14:31 key4.db
-rw-------. 1 root root   429 Jun 18 14:31 pkcs11.txt
```

Also, even if it were working as currently intended, that still wouldn't work because: we are trying to create the directory and all files with mode 0640, but the directory requires execute permissions.

lib/puppet/type/nssdb.rb
```diff
@@ -0,0 +1,53 @@
...
+
+    Puppet::Type.metaparams.each do |metaparam|
+      unless self[metaparam].nil? || excluded_metaparams.include?(metaparam)
+        file_opts[metaparam] = self[metaparam]
+      end
+    end
+
+    file_opts.merge({ :path => self[:nssdb_dir] })
+    file_opts.merge({ :ensure => self[:ensure] == :absent ? :absent : :directory })
+    file_opts.merge({ :recurse => true })
```
ehelms, line 49: `merge` is not an inplace update to the hash. Ruby tends to have functions with `!` at the end, such as `merge!` to differentiate "destructive" actions. That being said, I think you could just define these at the top like we do in other resources:

https://github.com/theforeman/puppet-certs/blob/master/lib/puppet/type/keystore.rb#L33-L36

wbclark: Thanks @ehelms !

Originally I was doing something like

```ruby
resources = []
resources.append(Puppet::Type.type(:file).new(common_file_opts.merge({ :new_opt_key => :new_opt_value })))
resources.append(Puppet::Type.type(:file).new(common_file_opts.merge({ :new_opt_key => :new_opt_different_value })))
```

And I didn't change the pattern when I moved to multiple new option keys (at this point it would have been broken without `merge!`), and then moved away from managing multiple resources entirely.

Now I moved all of these static attributes to the top to avoid using `merge` or `merge!` entirely, since it's no longer needed.

wbclark:
```
[root@centos8-64-1 ~]# ls -la /etc/pki/katello/nssdb
total 76
drwxr-x---. 2 root qpidd    55 Jun 18 15:18 .
drwxr-xr-x. 5 root root     75 Jun 18 15:17 ..
-rw-r-----. 1 root qpidd 36864 Jun 18 15:18 cert9.db
-rw-r-----. 1 root qpidd 36864 Jun 18 15:17 key4.db
-rw-r-----. 1 root qpidd   429 Jun 18 15:17 pkcs11.txt
```

Does puppet automatically add execute permissions to the directory when I use `mode => 0640, recurse => true` ?! Pretty neat if that's actually the case, and saves me a major headache.

manifests/ssltools/nssdb.pp
ehelms: I think we should aim to be least invasive as possible. You'll have to additionally go clean up https://github.com/theforeman/puppet-certs/blob/master/manifests/init.pp#L104 as you are now defining this in two locations.

From an interface definition standpoint, I think declaring what the nssdb path is makes more sense than declaring a parent folder and then always naming it nssdb. But I am getting into broader design thinking versus how we use this today and how we want to use this tomorrow (e.g. if we move to the planned model of puppet-qpid declaring the nssdb we would likely want to put it in /etc/qpid/certs for example)

wbclark:
:green_apple: :green_apple: :green_apple: !

wbclark:
Updated the interface for certs::ssltools::nssdb to keep the nss_db_dir paramter and add a new nss_db_password_file parameter. So this can now be labeled as an enhancement rather than a breaking change.

manifests/ssltools/nssdb.pp
```diff
@@ -1,25 +1,13 @@
...
-
   ensure_packages(['openssl', 'nss-tools'])
 
-  file { $nss_db_dir:
-    ensure => directory,
-    owner  => 'root',
-    group  => $group,
-    mode   => '0755',
-  } ->
   exec { 'generate-nss-password':
```
ehelms, line 11: Capturing an IRC discussion point, can we move this to use the `cache_data` model we use everywhere else?

lib/puppet/provider/nssdb/certutil.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  def nssdb_exists?
+    certutil(
+      '-L',
+      '-d', resource[:nssdb_dir],
+      '-f', resource[:password_file]
+    )
+  rescue Puppet::ExecutionFailure
+    Puppet.debug("Did not find existing NSS database at #{resource[:nssdb_dir]}")
```
ekohl, line 47: Is an execution failure always this? Can't there be some other error? Like an invalid password in the password file.

wbclark: Great point.

I'll make the change to

```ruby
    Puppet.debug("Unable to verify NSS database at #{resource[:nssdb_dir]} with password file #{resource[:password_file]}: #{e}")
```

Open to alternative wording if you have strong feelings about it.

BTW, I think we are OK with treating incorrect password file contents / location as -> NSS database does not exist and should be (re)-created. That should subsequently trigger a refresh of the `nssdb_certificate` resources, which should still add the same certificate content as before to the new NSS db.

manifests/ssltools/nssdb.pp
wbclark: Done! See https://github.com/theforeman/puppet-certs/pull/344/commits/6d8af5e6ffe97174f058ad61e3cb6ee7a568f438

manifests/ssltools/nssdb.pp
```diff
@@ -1,50 +1,27 @@
 # Sets up nssdb
 class certs::ssltools::nssdb (
   $nss_db_dir = $certs::nss_db_dir,
+  $nss_db_password_path = $certs::nss_db_password_path,
+  $nss_db_password = $certs::nss_db_password,
   $group = 'qpidd',
 )  {
   Exec { logoutput => 'on_failure' }
 
-  $nss_db_password_file   = "${nss_db_dir}/nss_db_password-file"
```
ehelms, line 8: Unfortunately ,changing this to be the file itself and not the path will break things:

https://github.com/theforeman/puppet-katello/blob/master/manifests/qpid.pp#L23-L25

manifests/init.pp
```diff
@@ -102,6 +102,9 @@
   }
 
   $nss_db_dir   = "${pki_dir}/nssdb"
+  $nss_db_password_file = 'nss_db_password-file'
+  $nss_db_password_path = "${pki_dir}/${nss_db_password_file}"
+  $nss_db_password = extlib::cache_data('foreman_cache_data', $nss_db_password_file, extlib::random_password(32))
```
ehelms, line 107: We could use `basename` (https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/REFERENCE.md#basename) here passing in the current `$nss_db_password_file`

manifests/ssltools/nssdb.pp
wbclark: As we undertake this work of building better tools to manage certificates for a more flexible, modular, and maintainable design, I'd like to standardize as much as possible the naming conventions for variables between the various providers and manifests here.

So with `keystore_certificate` we have an established convention that `name` refers to the file basename and `path` refers to fully qualified path.

I've opened https://github.com/theforeman/puppet-katello/pull/419 to explore what it looks like to go forward with the change.

wbclark: As an aside, should the change be considered "breaking" if it only renames a variable that is not in the declared interface for the class? Is this a case where we should perhaps move that variable into the interface, or is this actually pretty standard practice whenever the class isn't explicitly declared "private" ?

manifests/init.pp
wbclark: @ehelms do you feel strongly that this is the better alternative vs. making the simple change in puppet-katello I suggested on the other comment?

manifests/ssltools/nssdb.pp
ehelms: I'm glad you asked, cause I also dont know that I am clear on the answer between variables declared as parameters vs. variables found in the class itself. Help us @ekohl you're our only hope

ehelms: > As we undertake this work of building better tools to manage certificates for a more flexible, modular, and maintainable design, I'd like to standardize as much as possible the naming conventions for variables between the various providers and manifests here.
> 
> So with `keystore_certificate` we have an established convention that `name` refers to the file basename and `path` refers to fully qualified path.

I think the care that should be taken is not mixing the two concerns: creating a new provider and renaming variables. Both are valid, and can be done independently of one another. So if you think we should rename, lets do that as its own PR and change and then build on that. Or tackle it as a follow up. This makes the changes easier to understand, review and also coordinate the fixes of across modules.

If we aim to push the management of the keystore, truststore and nssdb into the modules that use them (e.g. puppet-candlepin, puppet-qpid) I am not too worried about naming conventions in this module as we will be doing a larger change and can ensure we follow convention there.

manifests/init.pp
ehelms: This captures my thoughts -- https://github.com/theforeman/puppet-certs/pull/344#discussion_r655618132

If you want to make this naming change, I would prefer if you split it out to its own issue, we merge that change and the puppet-katello change then layer this on top. Or vice versa.

manifests/ssltools/nssdb.pp
wbclark: > I'm glad you asked, cause I also dont know that I am clear on the answer between variables declared as parameters vs. variables found in the class itself. Help us @ekohl you're our only hope

Thinking about it a bit more, I guess the answer is obvious: declared parameters are inputs to the class which can be (or are required to be) assigned a value. Variables which are not declared as parameters could still be used elsewhere, but you don't have the ability to directly assign different values to them.

manifests/ssltools/nssdb.pp
```diff
@@ -1,50 +1,27 @@
...
-    ensure => directory,
-    owner  => 'root',
-    group  => $group,
-    mode   => '0755',
-  } ->
-  exec { 'generate-nss-password':
-    command => "openssl rand -base64 24 > ${nss_db_password_file}",
-    path    => '/usr/bin',
-    umask   => '0027',
+  file { $nss_db_password_path:
```
ehelms, line 12: Learning from a mistake I made, we'll want `show_diff => false` on this resource.

manifests/ssltools/nssdb.pp
```diff
@@ -1,50 +1,27 @@
...
-  file { $nssdb_files:
-    owner => 'root',
-    group => $group,
-    mode  => '0640',
+    mode    => '0640',
+  }
+
+  nssdb { $nss_db_dir:
+    ensure        => present,
+    password_file => $nss_db_password_path,
```
ehelms, line 22: 'Tis a bit ironic that we moved to using `_path` terminology as the variable but are passing it to a parameter using `_file` terminology. This gets into some bikeshedding of the naming conventions but I think it somewhat draws to my point here https://github.com/theforeman/puppet-certs/pull/344/files#r655487885

I think we should strive for consistency in our use of words.

lib/puppet/provider/nssdb/certutil.rb
```diff
@@ -0,0 +1,50 @@
+require 'fileutils'
+
+Puppet::Type.type(:nssdb).provide(:certutil) do
+  commands :certutil => 'certutil'
```
ekohl, line 4: I think this is the same
```suggestion
  commands :certutil
```

manifests/init.pp
ekohl: The `foreman_cache_data` cache namespace feels odd to me. If you read https://github.com/voxpupuli/puppet-extlib/blob/master/lib/puppet/functions/extlib/cache_data.rb you'll see it ends up in `$vardir/foreman_cache_data/nss_db_password-file.yaml`. On the other hand, we don't want to pollute `$vardir` with many directories. However, I think this would at least be clearer:
```suggestion
  $nss_db_password = extlib::cache_data('foreman_cache_data', 'certs-nss-db-password', extlib::random_password(32))
```

ekohl (commented):
If you're moving the password file, should we ensure the file is absent just to not leave cruft behind?

manifests/ssltools/nssdb.pp
ekohl: Many tools actually have some option that refers to file. For example, `-storepass:file` or `--password file:/path/to/file`. Given nss and certutil both refer to a file and not a path, I think the original password_file naming is not something we need to change.

wbclark:
Pushed an update based on some of the easier feedback to address.

Several other points of feedback really involve the same core area, which is the `_file` vs. `_path` terminology as well as how we name the cache data. I'm holding on addressing that until I am able to discus with @ehelms a bit later this afternoon, in order to settle on a design and tackle these issue simultaneously.

manifests/init.pp
wbclark: Thanks, @ekohl . That's a good point.

Looking around a bit, in pulpcore we use the `pulpcore_cache_data` cache namespace. But in puppet-candlepin, it's also `foreman_cache_data` which is a bit awkward: https://github.com/theforeman/puppet-candlepin/blob/b3b234376b48d00db22f8cb91913aa3e12bc964d/manifests/params.pp#L6

As it's currently written, changing the name of the password file which would necessarily mean changing the password itself. With your suggestion we could rename the file independently of changing the password, although I'm not sure if there is ever a real reason to do so -- I think it would still cause a refresh of the entire resource either way.

I would say, if we want to make that change, then let's remember to address it here as well: https://github.com/theforeman/puppet-certs/blob/fe831c38c8488be70ca28bb41dfb24854d94a8a6/manifests/candlepin.pp#L71-L72 .

manifests/ssltools/nssdb.pp
wbclark: I am really just looking for consistency, because we already do use this convention at https://github.com/theforeman/puppet-certs/blob/fe831c38c8488be70ca28bb41dfb24854d94a8a6/manifests/candlepin.pp#L73-L74

I think @ehelms shares the same goal and I suspect you are on the same page also, so I think it comes down to a matter of what should we address in this PR vs. what is well beyond the scope and needs to be addressed separately.

wbclark:
> If you're moving the password file, should we ensure the file is absent just to not leave cruft behind?

Thanks for bringing this up @ekohl . I've come up with what I believe is a very clever solution to this cleanup. Please see the new commit https://github.com/theforeman/puppet-certs/pull/344/commits/143c8ee47873cf6b861ac5919f6e15eb17224b5c and description in the commit message.

We should be able to verify this after merging by checking the nightly upgrade pipeline.

lib/puppet/provider/nssdb/certutil.rb
wbclark: You're right -- I tested it and confirmed it works.

wbclark: Actually I ran the wrong tests. It leads to `undefined method each` error

manifests/init.pp
```diff
@@ -102,6 +102,9 @@
   }
 
   $nss_db_dir   = "${pki_dir}/nssdb"
+  $nss_db_password_file = 'nss_db_password-file'
```
ehelms, line 105: I believe you do not need this anymore. You could keep it around for compatibility, but I am OK with merging this PR and https://github.com/theforeman/puppet-katello/pull/419 together given all the changes we have going in across the modules.

lib/puppet/provider/nssdb/certutil.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  def generate_nssdb
+    certutil(
+      '-N',
+      '-d', resource[:nssdb_dir],
+      '-f', resource[:password_file]
+    )
+  rescue Puppet::ExecutionFailure => e
+    Puppet.err("Failed to generate new NSS database at #{resource[:nssdb_dir]} with password file #{resource[:password_file]}: #{e}")
+    return nil
```
ekohl, line 34: Shouldn't this actually raise an error?
```suggestion
    raise Puppet::Error.new("Failed to generate new NSS database at #{resource[:nssdb_dir]} with password file #{resource[:password_file]}: #{e}", e)
```

manifests/ssltools/nssdb.pp
ekohl: I think variables declared inside a class is a bit of an odd thing in versioning. AFAIK it's not really described in documentation and as such not really an API (as opposed to class parameters). However, Puppet doesn't have a concept of private variables so other classes can use them (and we do - it can be very useful).

Maybe one solution is to always consider them private unless otherwise declared. That declaration could be done in puppet-strings using `@example` blocks. Really doing it well would probably mean introducing explicit `@variable` tags in puppet-strings.

That said, we don't make a statement about the API of this class. Perhaps this whole file should be considered private and add `@api private`.

wbclark:
Updated with the following changes:

1. Abandoned the `_file` -> `_path` rename as this was originally pursued for consistency with another bit of code. Now that @ekohl has pointed out that I shouldn't use a variable for the name of the cached extlib data, it no longer makes sense to align with that other design which I was copying, and consistency with all the other components named `_file` in closer proximity made far more sense.

For this reason I've also closed https://github.com/theforeman/puppet-katello/pull/419 . Thanks both of you for all the discussion on this topic.

2. The generate method should correctly raise a Puppet error.

3. Re-ordered and squashed some commits where it made sense to do so.

ekohl (changes requested):

manifests/ssltools/nssdb.pp
```diff
@@ -1,50 +1,28 @@
...
   Exec { logoutput => 'on_failure' }
 
-  $nss_db_password_file   = "${nss_db_dir}/nss_db_password-file"
-
-  $nssdb_files = $facts['os']['release']['major'] ? {
-    '7' => ["${nss_db_dir}/cert8.db", "${nss_db_dir}/key3.db", "${nss_db_dir}/secmod.db"],
-    default => ["${nss_db_dir}/cert9.db", "${nss_db_dir}/key4.db", "${nss_db_dir}/pkcs11.txt"]
-  }
-
   ensure_packages(['openssl', 'nss-tools'])
```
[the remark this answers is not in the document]
wbclark: Nice catch! TY

manifests/init.pp
```diff
@@ -102,6 +102,8 @@
   }
 
   $nss_db_dir   = "${pki_dir}/nssdb"
+  $nss_db_password_file = "${pki_dir}/nss_db_password-file"
+  $nss_db_password = extlib::cache_data('foreman_cache_data', 'certs-nss-db-password', extlib::random_password(32))
```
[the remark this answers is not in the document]
wbclark: Updated with the move to nss_db_password default to nssdp.pp as class default. I left the nss_db_password_file where it is.

I believe puppet-pulpcore uses an extlib::cache_data function in its class defaults for init.pp but I suppose that isn't parsed by Kafo

manifests/init.pp
```diff
@@ -102,6 +102,7 @@
   }
 
   $nss_db_dir   = "${pki_dir}/nssdb"
+  $nss_db_password_file = "${pki_dir}/nss_db_password-file"
```
ekohl, line 105: What is the benefit of having this in `init.pp`?

manifests/init.pp
```diff
@@ -102,6 +102,7 @@
   }
 
   $nss_db_dir   = "${pki_dir}/nssdb"
```
ekohl, line 104: For reference: f71e54a1c4c0981b951f6ad5118b837825b8148d and I'll quote:

> Technically we should remove it from init.pp too and only define it in nssdb.pp as a class parameter default, but puppet-katello uses it on the main class so we keep it in for now.

However, it looks like https://github.com/theforeman/puppet-katello/commit/6476f28f50e8256615403b07ce447d3315d70d85 fixed that. So now I'd prefer to remove this variable from init.pp and move it all to nssdb.pp.

manifests/init.pp
wbclark: I thought puppet-katello was pulling it from here, but on a closer reading it actually references `certs::qpid::nss_db_password_file` which in turn pulls it from `certs::ssltools::nssdb::nss_db_password_file` so I think we're good to move it to nssdb.pp

manifests/init.pp
wbclark: Updated with the move of $nss_db_dir and $nss_db_password_file to nssdb.pp, TY

lib/puppet/provider/nssdb/certutil.rb
```diff
@@ -0,0 +1,50 @@
...
+  def destroy_nssdb
+    FileUtils.rm_rf(resource[:nssdb_dir], secure: true)
+  end
+
+  def nssdb_exists?
+    certutil(
+      '-L',
+      '-d', resource[:nssdb_dir],
+      '-f', resource[:password_file]
+    )
```
ehelms, line 45: I led you astray here slightly, turns out that to view a NSS DB you do not need the password, and its not even an option for `-L`.

```
-L              List all certs, or print out a single named cert (or a subset)
   -h token-name     Name of token to search ("all" for all tokens)
   -n cert-name      Pretty print named cert (list all if unspecified)
   --email email-address 
                     Pretty print cert with email address (list all if unspecified)
   -d certdir        Cert database directory (default is ~/.netscape)
   -P dbprefix       Cert & Key database prefix
   -X                force the database to open R/W
   -r                For single cert, print binary DER encoding
   -a                For single cert, print ASCII encoding (RFC1113)
   --dump-ext-val OID 
                     For single cert, print binary DER encoding of extension OID
```

However, what does work is to try and list all private keys:

```
# certutil -K -d /etc/pki/katello/nssdb/ -f /etc/pki/katello/nssdb/nss_db_password-file 
certutil: Checking token "NSS Certificate DB" in slot "NSS User Private Key and Certificate Services"
< 0> rsa      f2e5cc6d95dcef5d1edfeb34bc0818c7fc1f2cc1   broker
```

And if bad password:

```
# certutil -K -d /etc/pki/katello/nssdb/ -f /etc/pki/katello/nss_db_password-file 
certutil: Checking token "NSS Certificate DB" in slot "NSS User Private Key and Certificate Services"
Incorrect password/PIN entered.
certutil: could not authenticate to token NSS Certificate DB.: SEC_ERROR_BAD_PASSWORD: The security password entered is incorrect.
```

I found this test running an upgrade pipeline to ensure this would work as we have theorized.
```
```

manifests/ssltools/nssdb.pp
```diff
@@ -1,50 +1,30 @@
...
-  } ->
-  exec { 'generate-nss-password':
-    command => "openssl rand -base64 24 > ${nss_db_password_file}",
-    path    => '/usr/bin',
-    umask   => '0027',
-    group   => $group,
-    creates => $nss_db_password_file,
-    require => Package['openssl'],
-  } ->
   file { $nss_db_password_file:
```
ekohl, line 14: You should clean up the old password file. Having unused password files around is confusing:
```suggestion
  file { "${nss_db_dir}/nss_db_password-file":
    ensure => absent,
  }

  file { $nss_db_password_file:
```

manifests/ssltools/nssdb.pp
```diff
@@ -1,50 +1,30 @@
 # Sets up nssdb
 class certs::ssltools::nssdb (
-  $nss_db_dir = $certs::nss_db_dir,
+  $pki_dir = $certs::pki_dir,
```
ekohl, line 3: I'd consider changing this:
```suggestion
  $nss_db_dir = "${certs::pki_dir}/nssdb",
  $nss_db_password_file = "${certs::pki_dir}/nss_db_password-file",
```

That also makes me wonder why we prefix eveything with `nss_db` if the class is already called `nssdb`. It ends up being `$certs::ssltools::nssdb::nss_db_*`

Perhaps we should make the class parameters unprefixed. We can keep compatibility variables in place until we update the code that uses them.

manifests/ssltools/nssdb.pp
```diff
@@ -1,50 +1,30 @@
...
+  $nss_db_password = extlib::cache_data('foreman_cache_data', 'certs-nss-db-password', extlib::random_password(32)),
   $group = 'qpidd',
 )  {
   Exec { logoutput => 'on_failure' }
```
ekohl, line 7: There is no more exec
```suggestion
```

ekohl (changes requested):

manifests/ssltools/nssdb.pp
ehelms: The old file gets cleaned up automatically, since the password changes, the nssdb is re-created and since it manages the directory it purges any files in there.

manifests/ssltools/nssdb.pp
wbclark: Nice catch, ty

manifests/ssltools/nssdb.pp
wbclark: You're right, this interface makes a lot more sense. I was sort of thinking about what is the minimum information I need to pass to make this work, and lost sight of, what are the core things I should be able to specify when creating an NSS DB.

wbclark: > That also makes me wonder why we prefix eveything with `nss_db` if the class is already called `nssdb`. It ends up being `$certs::ssltools::nssdb::nss_db_*`
> 
> Perhaps we should make the class parameters unprefixed. We can keep compatibility variables in place until we update the code that uses them.

I think this is a good point also. Probably best suited for a separate PR.

lib/puppet/provider/nssdb/certutil.rb
wbclark: Updated to use `-K`

If you run the pipeline again, please let me know where I could observe. TY!

manifests/ssltools/nssdb.pp
wbclark: It's subtle, but: https://github.com/theforeman/puppet-certs/pull/344/commits/a872939829a786b44db6aa1723ed25f8b5ed5060

Since the old password file exists in the NSS DB directory, it will get cleaned up when the NSS DB gets re-created.

wbclark: Technically the new password file does get created before the `nssdb` resource is evaluated (due to the auto-requires).

But the change to extlib cache/generate means the password itself should also change, and it is for this reason that the existence check will always return false the first time, as its performed with a new password.

There is no edge case where a user had previously overridden the password parameter, because it wasn't previously available as a parameter.

ehelms (approved):
This needs a squash either before or on merge.

This passed on a upgrade pipeline locally so I think it's good to go. Leaving unmerged to give @ekohl a chance to check given his reviews.

manifests/ssltools/nssdb.pp
```diff
@@ -1,50 +1,26 @@
 # Sets up nssdb
 class certs::ssltools::nssdb (
-  $nss_db_dir = $certs::nss_db_dir,
+  $nss_db_dir = "${certs::pki_dir}/nssdb",
+  $nss_db_password_file = "${certs::pki_dir}/nss_db_password-file",
+  $nss_db_password = extlib::cache_data('foreman_cache_data', 'certs-nss-db-password', extlib::random_password(32)),
   $group = 'qpidd',
```
ekohl, line 6: Just a last thing: can we use data types?
```suggestion
  Stdlib::Absolutepath $nss_db_dir = "${certs::pki_dir}/nssdb",
  Stdlib::Absolutepath $nss_db_password_file = "${certs::pki_dir}/nss_db_password-file",
  String[10] $nss_db_password = extlib::cache_data('foreman_cache_data', 'certs-nss-db-password', extlib::random_password(32)),
  String[1] $group = 'qpidd',
```

ekohl (approved):

manifests/ssltools/nssdb.pp
ehelms: That's a good call out.

wbclark:
Added data types to nssdb.pp & squashed commits.

Letting tests run with later changes and then will merge.

PR #346, opened 2021-06-01
Declare the build directory for all certificate creation

ehelms:
The build directory for certificates is declared as a parameter
but has no direct link to what the underlying tool would use
as the build directory. This change links the two up so that
there is no discrepancy and enables relying on the build directory
and paths within it.

### DIFF
--- a/lib/puppet_x/certs/common.rb
+++ b/lib/puppet_x/certs/common.rb
@@ -42,6 +42,18 @@ module PuppetX
 
         newparam(:password_file)
 
+        newparam(:build_dir) do
+          defaultto('/root/ssl-build')
+
+          validate do |value|
+            if value.empty?
+              raise ArgumentError, "build_dir cannot be empty"
+            else
+              super(value)
+            end
+          end
+        end
+
         newparam(:ca) do
           validate do |value|
             ca_resource = resource.catalog.resource(value.to_s)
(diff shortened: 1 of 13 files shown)


lib/puppet_x/certs/common.rb
```diff
@@ -42,6 +42,8 @@ module Common
 
         newparam(:password_file)
 
+        newparam(:build_dir)
```
ekohl, line 45: Should this have a default value? We should also verify that it's not empty, right? I don't think the tool underneath it would handle that well.

ekohl (commented):
:+1: to being explicitly passing it to the tool.

There is this this ugly trick in puppet-jenkins that looks at a class default:
https://github.com/voxpupuli/puppet-jenkins/blob/41d8090fc0da538f3dd6ffbc6b5f5e86d7755e46/lib/puppet/x/jenkins/config.rb#L50-L57
That's then used here:
https://github.com/voxpupuli/puppet-jenkins/blob/41d8090fc0da538f3dd6ffbc6b5f5e86d7755e46/lib/puppet/x/jenkins/provider/cli.rb#L164

Perhaps that's also a way to deal with it here? I'm not sure I like the magic way, but I think you should be aware of it.

lib/puppet_x/certs/common.rb
ehelms: I added both just to make it easier. The Jenkins trick does look a bit magical and I'd rather just be explicit.

ekohl (approved):

PR #348, opened 2021-06-01
Deploy managed certificate and key from build directory

ehelms:
This change no longer installs the RPMs that are generated when creating
certificates. Instead, the certificates that are deployed are pulled
from their locations generated on disk by the katello-ssl-tool. This
removes a layer of indirection and simplifies the creation and management
of the certificates.

With the RPMs present, a given key-pair would be created in the
ssl_build_dir, then deployed to /etc/pki/katello-certs-tools and then
copied and managed somewhere else on disk, often /etc/pki/katello.

### DIFF
--- a/lib/puppet/provider/certificate_file/certificate_file.rb
+++ b/lib/puppet/provider/certificate_file/certificate_file.rb
@@ -1,0 +1,3 @@
+Puppet::Type.type(:certificate_file).provide(:openssl) do
+  commands :openssl => 'openssl'
+end
--- a/lib/puppet/provider/certificate_file/certificate_file.rb
+++ b/lib/puppet/provider/certificate_file/certificate_file.rb
@@ -1,3 +1,24 @@
 Puppet::Type.type(:certificate_file).provide(:openssl) do
   commands :openssl => 'openssl'
+
+  def source_cert
+    if valid?(resource[:source_cert])
+      resource[:source_cert]
+    else
+      raise ArgumentError, "Source certificate is not a valid x509 certificate"
+    end
+  end
+
+  def valid?(certificate)
+    File.exist?(certificate)
+    pem_format?(certificate)
+  end
+
+  def pem_format?(certificate)
+    openssl(
+      'x509',
+      '-inform', 'PEM',
+      '-in', certificate
+    )
+  end
 end
--- a/lib/puppet/type/certificate_file.rb
+++ b/lib/puppet/type/certificate_file.rb
@@ -1,0 +1,48 @@
+Puppet::Type.newtype(:certificate_file) do
+  desc 'Manages an x509 certificate file from a source certificate'
+
+  ensurable
+
+  newparam(:path, :namevar => true) do
+    desc "Path the certificate will be copied to."
+  end
+
+  newparam(:owner, parent: Puppet::Type::File::Owner) do
+    desc "Specifies the owner of the truststore. Valid options: a string containing a username or integer containing a uid."
+  end
+
+  newparam(:group, parent: Puppet::Type::File::Group) do
+    desc "Specifies a permissions group for the truststore. Valid options: a string containing a group name or integer containing a gid."
+  end
+
+  newparam(:mode, parent: Puppet::Type::File::Mode) do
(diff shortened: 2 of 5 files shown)


ehelms:
This does not require but should depend on https://github.com/theforeman/puppet-certs/pull/346 which makes using the build directory more robust.

I have only converted one resource to show what this would look like and get feedback on this design change before I proceed further to the other certificate resources. Additionally, I could try to:

 a) modify the keypair resource to this model
 b) add providers for each public key and private key based on this model

This additionally will help with allowing this to work on non-RPM based systems as well.

ekohl (commented):
I think `certs::keypair` could actually do this. It already provides the abstraction needed.

However, perhaps `deploy` should accepts parameters where to deploy the files to and which permissions to use?

It may break down with something like an nss keystore, but the trivial case of a keypair it could simplify things.

ehelms:
Adding a note, this change will require this change https://github.com/theforeman/puppet-certs/pull/352 so that certificates coming from a tarball are available in the build directory for deployment. The tests added in that PR should help confirm the behavior continues to work through this re-factor.

ehelms:
I added a commit (https://github.com/theforeman/puppet-certs/pull/348/commits/6e61a88b180351648e1ecebdeafab130f348b708) to show what using a defines for the key pair would look like. I introduced a `key_pair` that is new to differentiate it from `keypair` given the amount of changes required in the new model. I think when looking at it some of the questions to ask ourselves is:

 a) does key_pair provide a meaningful abstraction compared to having two file declarations?
 b) should key_pair be a provider instead?
 c) should key_pair aim to be usable outside of puppet-certs, this is towards a more generic, re-usable type by for example puppet-candlepin?

manifests/key_pair.pp
```diff
@@ -0,0 +1,32 @@
...
+  $source_dir,
+  $key_file,
+  $cert_file,
+  $key_owner   = undef,
+  $key_group   = undef,
+  $key_mode    = undef,
+  $cert_owner  = undef,
+  $cert_group  = undef,
+  $cert_mode   = undef,
+  $require     = undef,
```
ekohl, line 11: I don't think you need to do this. That should be built into Puppet itself already.

manifests/key_pair.pp
```diff
@@ -0,0 +1,29 @@
...
+
+  file { $key_file:
+    ensure  => file,
+    source  => "${source_dir}/${title}.key",
+    owner   => $key_owner,
+    group   => $key_group,
+    mode    => $key_mode,
+  }
+
+  certificate_file { $cert_file:
```
ehelms, line 21: I added this new type to show how we could go with a provider for certificate file (and then one for private key) to provide and abstraction and additional validations. We should ask ourselves if this abstraction adds enough value on top of the File resource to be worth the additional layer.

lib/puppet/provider/certificate_file/certificate_file.rb
```diff
@@ -0,0 +1,24 @@
...
+      resource[:source_cert]
+    else
+      raise ArgumentError, "Source certificate is not a valid x509 certificate"
+    end
+  end
+
+  def valid?(certificate)
+    File.exist?(certificate)
+    pem_format?(certificate)
+  end
```
ehelms, line 15: Other things we could attempt to do is pass in the private key to this resource and validate that they have the same modulo to ensure they will work together. If we passed in the CA certificate that is supposed to be supplied with the certificate we could additionally verify they work together.

lib/puppet/type/certificate_file.rb
```diff
@@ -0,0 +1,57 @@
...
+  end
+
+  newproperty(:source_cert) do
+    desc "Specifies the source certificate that will be copied to the declared location."
+  end
+
+  def generate
+    file_opts = {
+      ensure: (self[:ensure] == :absent) ? :absent : :file,
+      source: self[:source_cert],
```
ehelms, line 33: The source_cert has to be implemented as a property and validated
inside the provider code. The validate block of a type executes during
catalog compilation and does not give access to the certificate on disk. 
Thus we pass it to the source attribute on the file to allow the File to 
handle the source.

PR #349, opened 2021-06-02
Mark compatible with puppet/trusted_ca 4.x

ekohl:
This major version only dropped some support so there's no API compatibility changes.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -10,7 +10,7 @@
   "dependencies": [
     {
       "name": "puppet-trusted_ca",
-      "version_requirement": ">= 2.0.0 < 4.0.0"
+      "version_requirement": ">= 2.0.0 < 5.0.0"
     },
     {
       "name": "puppetlabs-stdlib",


evgeni (approved):

ehelms:
Hrmm, do we still use this dependency? I recall it was used for the bootstrap RPM which we moved over here -- https://github.com/theforeman/puppet-foreman_proxy_content/blob/master/manifests/bootstrap_rpm.pp#L23-L27

ekohl:
Hmm, good point. Then it should be moved. Still, it is very weird that the bootstrap rpm is responsible for adding the CA to the trusted host. AFAIK Katello relies on it right now since connections to Pulp aren't explicitly configured with a CA and some tool may also expect it. If we drop the bootstrap RPM (as is the plan), this will still be needed (until we remove the requirement altogether).

I'll drop it in this module for now and send a PR to the other module to make the dependency explicit.

PR #350, opened 2021-06-02
Remove dependency on trusted_ca

ekohl:
953261ffd2eb52b7176ab365fb0c4e9245435d99 removed the last use of trusted_ca so this module no longer depends on it.

Marking as an enhancement since this should result in a minor version bump. It was noted that theforeman/foreman_proxy_content didn't add the dependency explicitly. We do bump patch versions in stable releases and that can break.

Fixes: 953261ffd2eb52b7176ab365fb0c4e9245435d99


ekohl:
https://github.com/theforeman/puppet-foreman_proxy_content/pull/362 should probably be merged first.

ehelms (approved):
The change to explicitly add this to puppet-foreman_proxy_content has been merged.

PR #351, opened 2021-06-02
Add acceptance tests for tar_create and tar_extract

ehelms:


spec/acceptance/certs_tar_extract_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+      class { 'certs':
+        tar_file  => '/root/foreman-proxy.example.com.tar.gz',
+        generate  => false,
+        node_fqdn => 'foreman-proxy.example.com',
+      }
+
+      include certs::apache
+    PUPPET
+
+    apply_manifest(install_certs, catch_failures: true)
```
ekohl, line 32: Can we also check for idempotency?
```suggestion
    apply_manifest(install_certs, catch_failures: true)
    apply_manifest(install_certs, catch_changes: true)
```

ehelms: I can confirm that this fails idempotency which is why I have it written the way I do. These tests will allow us to try and fix that as we'll have something testing our baseline. The reason for failure is due to the fact that is always extracts the tarball or always creates it without any care if it already exists or not.

wbclark: @ehelms with the understanding that this PR only adds testing, and the broken idempotence is already present in the module, what are the consequences for the foreman-installer if the module always creates or extracts the tarball?

wbclark:
Pulled this down and ran both tests on EL8. The tests successfully checked for `File "/root/foreman-proxy.example.com.tar.gz"`, x509 certificate, private key, and `foreman-proxy.example.com-apache-1.0-1.noarch` package.

spec/acceptance/certs_tar_extract_spec.rb
ehelms: I think that is the expected behavior in the installer right now as it recreates it fresh every time it runs. I think making this idempotent will require some thought as to how that applies to the installer given the current behavior.

ekohl: Perhaps an explicit comment that it isn't idempotent makes sense?

ehelms: Comment added

ekohl (approved):

PR #352, opened 2021-06-02
Include certs directory tarball

ehelms:
Requires: https://github.com/theforeman/puppet-certs/pull/351

Adding a bit more than what is in the commit message. This is working towards an approach that we deploy certificates from the build directory to their managed locations and eventually drop the use of RPMs. This will cut down on the amount of locations certificates exist, the amount of processing by the puppet module and open this up for use on non-RPM systems.

In order to do this I am trying an approach of re-factoring underneath that does additive changes that allow the end state to remain unchanged while dismantling all of the RPM mechanisms. So in this case, include the certificates in the tarball, this will allow them to exist in `/root/ssl-build`. From there, we can begin to make changes such as in (https://github.com/theforeman/puppet-certs/pull/348) to deploy from the build directory. Once we are only deploying from the build directory across servers and content proxies the underlying RPM mechanisms can be disabled or removed entirely. Then we can begin to explore direct management of certificates in the puppet modules for the services (e.g. puppet-candlepin) by having split the generation of certificates from the deployment of certificates.

### DIFF
--- a/manifests/tar_create.pp
+++ b/manifests/tar_create.pp
@@ -12,9 +12,17 @@ define certs::tar_create(
   $path               = $title,
   $foreman_proxy_fqdn = $certs::foreman_proxy_content::foreman_proxy_fqdn,
 ) {
+
+  $ca_rpms = 'ssl-build/*.noarch.rpm'
+  $ca_certificates = 'ssl-build/*.crt'
+
+  $foreman_proxy_certificate_rpms = "ssl-build/${foreman_proxy_fqdn}/*.noarch.rpm"
+  $foreman_proxy_certificates = "ssl-build/${foreman_proxy_fqdn}/*.crt"
+  $foreman_proxy_keys = "ssl-build/${foreman_proxy_fqdn}/*.key"
+
   exec { "generate ${path}":
     cwd     => '/root',
     path    => ['/usr/bin', '/bin'],
-    command => "tar -czf ${path} ssl-build/*.noarch.rpm ssl-build/${foreman_proxy_fqdn}/*.noarch.rpm",
+    command => "tar -czf ${path} ${ca_rpms} ${ca_certificates} ${foreman_proxy_certificate_rpms} ${foreman_proxy_certificates} ${foreman_proxy_keys}",
   }
 }
--- a/spec/acceptance/certs_tar_extract_spec.rb
+++ b/spec/acceptance/certs_tar_extract_spec.rb
@@ -52,7 +52,34 @@ describe 'certs with tar archive' do
     it { should have_matching_certificate('/etc/pki/katello/certs/katello-apache.crt') }
   end
 
+  describe x509_certificate('/root/ssl-build/foreman-proxy.example.com/foreman-proxy.example.com-apache.crt') do
+    it { should be_certificate }
+    it { should be_valid }
+    it { should have_purpose 'server' }
+    include_examples 'certificate issuer', "C = US, ST = North Carolina, L = Raleigh, O = Katello, OU = SomeOrgUnit, CN = #{fact('fqdn')}"
+    include_examples 'certificate subject', "C = US, ST = North Carolina, O = Katello, OU = SomeOrgUnit, CN = foreman-proxy.example.com"
+    its(:keylength) { should be >= 4096 }
+  end
+
+  describe x509_private_key('/root/ssl-build/foreman-proxy.example.com/foreman-proxy.example.com-apache.key') do
+    it { should_not be_encrypted }
+    it { should be_valid }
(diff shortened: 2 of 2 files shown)


ehelms:
Rebased as master now contains #351

ekohl (commented):
In the tar create test, you can add a `command("tar -taf /path/to/file.tar.gz")` and check stdout contains the right files.

ehelms:
> In the tar create test, you can add a `command("tar -taf /path/to/file.tar.gz")` and check stdout contains the right files.

The test already extracts the tarball and allows me to check the end product which is what I ultimately care about so I think I am covered.

manifests/tar_create.pp
```diff
@@ -15,6 +15,6 @@
   exec { "generate ${path}":
     cwd     => '/root',
     path    => ['/usr/bin', '/bin'],
-    command => "tar -czf ${path} ssl-build/*.noarch.rpm ssl-build/${foreman_proxy_fqdn}/*.noarch.rpm",
+    command => "tar -czf ${path} ssl-build/*.noarch.rpm ssl-build/*.crt ssl-build/${foreman_proxy_fqdn}/*",
```
wbclark, line 18: Previously we are generating a tarball with `ssl-build/*.noarch.rpm` and `ssl-build/${foreman_proxy_fqdn}/*.noarch.rpm`

Now we still get `ssl-build/*.noarch.rpm`, and additionally `ssl-build/*.crt`.

And for the proxy, change to the wildcard path `ssl-build/${foreman_proxy_fqdn}/*`

It's clear from reading the description that this is intended to be an intermediate design that allows to gradually move towards the intended design -- is this why you still collect `ssl-build/*.noarch.rpm` if it exists, and additionally `ssl-build/*.crt`?

Why collect those separately, but for the proxy fqdn, a wildcard without any trailing characters? Wouldn't it be better to be much more specific about what we intend to find there?

wbclark: Also perhaps this is overkill, but in trying to review this with more limited knowledge of the current workings, something like this might be nice:

```puppet
# collect the existing RPM for backwards compatibility, to be removed in the future.
$rpm_rel_path = "ssl-build/*.noarch.rpm"

# also collect the .crt file. added with puppet-certs #352
$crt_file_rel_path = "ssl-build/*.crt"

# now all files in the $foreman_proxy_fqdn subdir , not only RPMs
$proxy_fqdn_wildcard_path = "ssl-build/${foreman_proxy_fqdn}/*"
```

Then build the command with:

```puppet
command => "tar -czf ${path} ${rpm_rel_path} ${crt_file_rel_path} ${proxy_fqdn_wildcard_path}",
```

Obviously this would work the same, but we (myself in particular) might appreciate this in the future when we revisit to make the next phase of changes.

ehelms: > What is less clear to me: why collect the above separately and somewhat specifically, but for the proxy fqdn, a wildcard without any trailing characters? Wouldn't it be better to be much more specific about what we intend to find there?

We do not want to collect all of `ssl-build/` because there is openssl configuration, CA key, serial etc. in that directory. For the directories that are for the proxy at, e.g. `ssl-build/${hostname}` we want the certificate and keys and the RPMs, so it just seemed easier to grab everything and not accidentally miss something since we effectively need it all right now.

ehelms: I have updated with these ideas in mind.

wbclark: Thanks! It's a lot clearer to me now and I'm sure I'll appreciate it later when we have to revisit this again for the next steps.

spec/acceptance/certs_tar_extract_spec.rb
```diff
@@ -52,7 +52,34 @@ class { 'certs':
     it { should have_matching_certificate('/etc/pki/katello/certs/katello-apache.crt') }
   end
 
+  describe x509_certificate('/root/ssl-build/foreman-proxy.example.com/foreman-proxy.example.com-apache.crt') do
+    it { should be_certificate }
+    it { should be_valid }
+    it { should have_purpose 'server' }
+    include_examples 'certificate issuer', "C = US, ST = North Carolina, L = Raleigh, O = Katello, OU = SomeOrgUnit, CN = #{fact('fqdn')}"
+    include_examples 'certificate subject', "C = US, ST = North Carolina, O = Katello, OU = SomeOrgUnit, CN = foreman-proxy.example.com"
+    its(:keylength) { should be >= 2048 }
```
wbclark, line 61:
```suggestion
    its(:keylength) { should be >= 4096 }
```

wbclark (commented):
Please see my comment above regarding the keylength. Once that is addressed, I think this should be merged.

wbclark (approved):

manifests/tar_create.pp
ekohl: Indeed, this naming makes it quite clear. Thanks!

PR #353, opened 2021-06-03
POC: show simple-deploy

ekohl:
The goal of this to show how we can skip the intermediate RPM install phase and directly deploy.

It chooses to skip the cert resource if certificates are passed in.

This is an alternative to https://github.com/theforeman/puppet-certs/pull/348.

### DIFF
--- a/manifests/apache.pp
+++ b/manifests/apache.pp
@@ -1,55 +1,27 @@
 # Certs configurations for Apache
 class certs::apache (
-  $hostname             = $certs::node_fqdn,
-  $cname                = $certs::cname,
-  $generate             = $certs::generate,
-  $regenerate           = $certs::regenerate,
   Boolean $deploy = $certs::deploy,
   Stdlib::Absolutepath $pki_dir = $certs::pki_dir,
-  Optional[Stdlib::Absolutepath] $server_cert          = $certs::server_cert,
-  Optional[Stdlib::Absolutepath] $server_key           = $certs::server_key,
-  $country              = $certs::country,
-  $state                = $certs::state,
-  $city                 = $certs::city,
-  $org                  = $certs::org,
-  $org_unit             = $certs::org_unit,
-  $expiration           = $certs::expiration,
-  $default_ca           = $certs::default_ca,
-  $ca_key_password_file = $certs::ca_key_password_file,
-  $group                = 'root',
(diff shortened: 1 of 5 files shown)


manifests/apache.pp
```diff
@@ -52,21 +42,24 @@
       ca            => $default_ca,
       generate      => $generate,
       regenerate    => $regenerate,
-      deploy        => $deploy,
+      deploy        => false,
```
ekohl, line 45: I also had a thought of extending `cert` to actually deploy the files directly (without the intermediate RPM step).

manifests/apache.pp
```diff
@@ -17,7 +16,7 @@
   $expiration           = $certs::expiration,
   $default_ca           = $certs::default_ca,
   $ca_key_password_file = $certs::ca_key_password_file,
-  $group                = $certs::group,
+  $group                = 'root',
```
ekohl, line 19: Apache actually reads certificates and keys as the root user so this is the tightest permission that will still work.

ehelms:
> The goal of this to show how we can skip the intermediate RPM install phase and directly deploy.
> 
> It chooses to skip the cert resource if certificates are passed in.

Thanks for showing this. This is on my radar for #348 however, I had not touched it yet because there are both design questions and other pieces that have to be "unraveled" first to enable doing this:

 1) https://github.com/theforeman/puppet-certs/pull/348#issuecomment-853261407
 2) https://github.com/theforeman/puppet-certs/pull/352

I have a branch sitting with the not creating RPMs if deploy is false (https://github.com/ehelms/puppet-certs/tree/do-not-create-rpms) but this is more of a last step given the changes it entails and some of the aforementioned unraveling we first need to do to handle both a server and proxy.

ehelms:
Ah ha! Now I get what you were driving at... not a common interface on top of an interface but rather heading towards that model of a single certificate per hostname being generated that can be then copied-and-managed by the services that need to manage the certs. I'll incorporate this into the design roadmap I am building as that will be a later stage change.

ekohl:
Exactly. I'm not entirely sure yet if this won't bite us later on, but I feel like few people need the flexibility of the many-certs approach.

ekohl:
I feel like I've shown what I wanted to show :)

PR #354, opened 2021-06-03
Design: Document for discussion and planning of certificate handling redesign

ehelms:
I thought I would try this method for communicating the design and technical Roadmap of certificate redesign work I am undertaking. I am trying to achieve larger design goals through small, incremental changes and realize that the steps I envision in my head may not be obvious to those reviewing the incoming work.

Thus, I am trying out a pull request with design document that will remain in draft status. I do not intend to merge this ever. I intend this to serve as a communication and reference as well as a place for discussions of key points through the ability to comment on specific lines.

I am sure I missed something that will come out through discussion. I have learned a lot about how all the current code works so if there is an area that is not super clear, or you'd like a deeper exposition of how a part of it works today I'll add that.

### DIFF
--- a/REDESIGN.md
+++ b/REDESIGN.md
@@ -1,0 +1,137 @@
+# Redesign of Certificate Handling
+
+This document is not intended to be merged to the repository in the end.
+This document aims to be an outline of the design strategy envisioned for this repository, and the roadmap of changes needed to help get it there.
+This document is intended to help with discussions and information sharing to make the large set of smaller PRs coming in easier to understand with respect to the overall strategy.
+This document will get updated with PRs as they aim at each Roadmap item as well as any Redmine issues that get created to track the progress.
+
+## Design Targets
+
+There are three major areas that are targeted for redesign within this module.
+
+ * Split certificate generation from certificate deployment and management
+ * Deploy and Manage certificates within service modules
+ * Generate one certificate for a hostname
+
+### Split certificate generation from certificate deployment and management
+
+#### Problem Statement
+
+In theory, this module aims to both generate certificates and to deploy them to their final, managed locations per the needs of each service. This is laid out in the (README)[https://github.com/theforeman/puppet-certs#phases] but not strictly true as there is tighter cohesion than expected. There is unnecessary work that happens within the code to achieve a split generate and deploy model that involves certificates living in three different locations and RPMs being built and installed.
+
+Let's recap the current locations that certificates live on disk and the "stages" each pertains to:
+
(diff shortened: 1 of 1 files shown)


REDESIGN.md
```diff
@@ -0,0 +1,143 @@
...
+    * Set `deploy => false` on `cert` provider to avoid RPM install
+  * Remove RPM deployment and generation
+    * Add a change that stops generating RPMs when `deploy => false` is set on the `cert` provider
+    * Clean out RPM handling code from `cert` provider
+
+### Deploy and Manage certificates within service modules
+
+#### Problem Statement
+
+Services almost always need to set some level of ownership on certificates in order to consume them at runtime. This can be as simple as user and group. The certificates used by a service are also often co-located within a directory owned by that service. This creates dependency cycles and burdens when trying to manage the certificates from within puppet-certs. For example, puppet-candlepin needs to install Candlepinand ensure the tomcat user and group exists, as well as the `/etc/candlepin/certs` directory in order for puppet-certs to set the owner and group on the certificates. And, puppet-candlepin needs puppet-certs to have created and deployed the certificates with those correct ownerships before the service can be started.
```
ekohl, line 108:
```suggestion
Services almost always need to set some level of ownership on certificates in order to consume them at runtime. This can be as simple as user and group. The certificates used by a service are also often co-located within a directory owned by that service. This creates dependency cycles and burdens when trying to manage the certificates from within puppet-certs. For example, puppet-candlepin needs to install Candlepin and ensure the tomcat user and group exists, as well as the `/etc/candlepin/certs` directory in order for puppet-certs to set the owner and group on the certificates. And, puppet-candlepin needs puppet-certs to have created and deployed the certificates with those correct ownerships before the service can be started.
```

ekohl (approved):
I'm not entirely use we need separate types for `certificate` and `private_key`. We can consider using https://github.com/voxpupuli/puppet-openssl as well but I think we'll find out along the way what works best.

Overall I agree with the plan. If we keep `ssl-build` we should also consider moving it to `/var/lib` because `/root` is a bad place for it.

ehelms:
> I'm not entirely use we need separate types for `certificate` and `private_key`. We can consider using https://github.com/voxpupuli/puppet-openssl as well but I think we'll find out along the way what works best.

Take a look at what I added to https://github.com/theforeman/puppet-certs/pull/348 and my comments at the end. I think this all comes down to how much we want to create an abstraction and how much we want to try to add some 'verifications'.

ehelms:
This ended up being effectively the playbook I followed to re-work certificates and is no longer needed.

PR #355, opened 2021-06-04
Add acceptance tests for certs::foreman

ehelms:


manifests/foreman.pp
```diff
@@ -15,7 +15,7 @@
   $default_ca            = $certs::default_ca,
   $ca_key_password_file  = $certs::ca_key_password_file,
   $server_ca             = $certs::server_ca,
-) inherits certs::params {
+) inherits certs {
```
wbclark, line 18: It's not immediately obvious to me what this has to adding the foreman_spec.rb . Is this so that the test can `include certs::foreman` without having to specify params, and should this PR be labeled as 'enhancement' then?

wbclark (commented):
I noticed you're testing owner/group/mode for `/etc/foreman/client_cert.pem` and `/etc/foreman/client_key.pem` but not `/etc/foreman/proxy_ca.pem` -- any particular reason why?

spec/acceptance/foreman_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+  context 'with default parameters' do
+    it_behaves_like 'an idempotent resource' do
+      let(:manifest) { 'include certs::foreman' }
+    end
+
+    describe x509_certificate('/etc/foreman/client_cert.pem') do
+      it { should be_certificate }
+      it { should be_valid }
+      it { should have_purpose 'client' }
+      include_examples 'certificate issuer', "C = US, ST = North Carolina, L = Raleigh, O = Katello, OU = SomeOrgUnit, CN = #{FQDN}"
```
wbclark, line 29: I'm seeing this multiple times. That says to me there is an opportunity for de-duplication.

wbclark (commented):
I pulled this down and validated it runs on my EL8 image.

Test output matches my expectation based on reading `spec/acceptance/foreman_spec.rb`.

manifests/foreman.pp
ehelms: Without this, these tests will fail because it does not load `certs` class which there are parameters that it depends upon. So you could think of this as a bug fix.

ekohl: I also read this as a bugfix. It's mostly just chance we never ran into it before due to the way we called the code (if you first include certs, it does work).

spec/acceptance/foreman_spec.rb
ekohl: The thing is that they sometimes do differ a bit. However, we're talking about completely redesigning this so it doesn't really make sense to change it now.

ekohl (approved):

manifests/foreman.pp
wbclark: OK. Rather than creating a separate PR to get that to show up in changelogs, how about we create and link an issue?

This one can still be labeled to skip the changelog, and only the issue would show up.

ekohl: You can also make this just a bugfix PR and consider the acceptance tests as a regression test.

ehelms: Y'all just tell me what you want to see happen. For me this is about adding acceptance tests, and uncovered a chance bug fix that is required in order for this testing to work since it calls this class directly (which is not something we do in reality anywhere).

ehelms:
> I noticed you're testing owner/group/mode for `/etc/foreman/client_cert.pem` and `/etc/foreman/client_key.pem` but not `/etc/foreman/proxy_ca.pem` -- any particular reason why?

Since we are not directly managing the `proxy_ca.pem` file via a File resource I opted not to test its properties directly.

manifests/foreman.pp
ekohl: I consider this a bugfix with a regression test, which to me is more important than only adding a regression test.

ehelms: I have split this single change out to it's own commit to fix the bug and then layered the new tests on top.

ehelms:
@ekohl @wbclark is this OK now?

wbclark:
This would show up in changelogs as a bugfix with the title: "Add acceptance tests for certs::foreman"

IMO that is pretty unclear, but I also agree that is the "correct" title for the PR, since that reflects the true intent and the bugfix was really incidental.

For that reason, I created and linked the issue https://github.com/theforeman/puppet-certs/issues/362 - "certs::foreman should inherit certs, but only inherits certs::params" , and I labeled this PR as 'skip changelog'

Now when we generate changelogs for the next release, the issue will show up in the changelog and it will make sense to the reader what is the actual bugfix. For details of the implementation, they will be able to follow the link from the issue to this PR.

wbclark (approved):
LGTM

PR #356, opened 2021-06-04
Add acceptance tests qpid router

ehelms:


wbclark (commented):
How do you to determine when it's necessary to include a before block like

```ruby
before(:all) do
  on default, 'rm -rf /root/ssl-build'
end
```

Was this necessary with the certs-tar acceptance test only because of the broken idempotence?

Clearly the tests here are working and there is no conflict, and I can see there is no overlap in the files/certs being tested, but isn't it best practice to blow away ssl-build at the start of each test and trust that they are all independent without having to worry about the order which they are run?

ehelms:
Nah, it was just a timing thing. I added the before(:all) to clear out ssl-build in another PR because it was necessary and this was done in parallel so it lacked it. Thus it would have more depended upon ordering. This PR now has that same setup in it.

wbclark (commented):
For `/etc/pki/katello/qpid_router_client.crt` we test that the keylength is >= 2048 , but for `/root/ssl-build/#{FQDN}/#{FQDN}-qpid-router-client.crt' we test that it is >= 4096 . Is this an oversight?

My understanding is that /root/ssl-build is somewhat like a staging area used for creating the certs before they are copied into the locations expected by the application, so I was expecting these to be the same.

ehelms:
> For `/etc/pki/katello/qpid_router_client.crt` we test that the keylength is >= 2048 , but for `/root/ssl-build/#{FQDN}/#{FQDN}-qpid-router-client.crt' we test that it is >= 4096 . Is this an oversight?

Probably some copy paste failures. I have updated them. I opened a separate PR to fix this everywhere: https://github.com/theforeman/puppet-certs/pull/359

ekohl (approved):

PR #359, opened 2021-06-15
Update keylength tests to check for >= 4096

ehelms:


ekohl:
Failure related:
```
   1) certs::apache with server cert X509 certificate "/etc/pki/katello/certs/katello-apache.crt" keylength is expected to be >= 4096
     Failure/Error: its(:keylength) { should be >= 4096 }
       expected: >= 4096
            got:    2048
```

ehelms:
Fixed the failure, was the case of custom certificates being supplied

wbclark (approved):

PR #360, opened 2021-06-18
Introduce nssdb and nssdb_certificate types

ekohl:
These are more expressive of what to match.

Very incomplete, just as a way to describe the layout and design.

### DIFF
--- a/spec/support/acceptance/nssdb.rb
+++ b/spec/support/acceptance/nssdb.rb
@@ -1,0 +1,57 @@
+begin
+  require 'serverspec'
+rescue LoadError
+  # Not using acceptance tests
+else
+  module Serverspec
+    module Type
+      class Nssdb < Command
+        def certificates
+          # TODO: parse the output
+          stdout
+        end
+
+        private
+
+        def command
+          "certutil -L -d #{@name}"
+        end
+      end
+
+      class NssdbCertificate < Command
+        def subject
+          # TODO: multi line support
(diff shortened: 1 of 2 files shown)


spec/support/acceptance/nssdb.rb
```diff
@@ -0,0 +1,65 @@
...
+
+        private
+
+        def command
+          "certutil -L -d #{@name}"
+        end
+
+        def command_result
+          @command_result ||= @runner.run_command(command)
+        end
```
wbclark, line 22: Does this need a `command_result` method? What is unique to `Nssdb` that can't be inherited from `Command`?

ekohl: Because the default implementation calls `@name`. I've submitted https://github.com/mizzy/serverspec/pull/616 which would make this easier.

spec/support/acceptance/nssdb.rb
```diff
@@ -0,0 +1,65 @@
...
+  require 'serverspec'
+rescue LoadError
+  # Not using acceptance tests
+else
+  module Serverspec
+    module Type
+      class Nssdb < Command
+        def certificates
+          # TODO: parse the output
+          command_result
```
ekohl, line 11: Note to self: this should probably be
```suggestion
          stdout
```

ehelms:
Given these will be removed when Qpid support is dropped, I'm opting to close.

PR #363, opened 2021-06-23
Test that foreman client_cert.pem has keylength >= 4096

wbclark:
This was a minor oversight in 0a362f38a95b701e9dbd3b528a462c6212d5751a


ehelms (approved):
Good find, thanks!

PR #365, opened 2021-06-25
Support ensuring certs::keypair cert and key can be absent

ehelms:
This includes https://github.com/theforeman/puppet-certs/pull/313/commits/862fc57d27aece10ab182236f5bd2fbb91390217 with a small fix to allow tests to pass (thanks @ekohl !).

### DIFF
--- a/lib/puppet/provider/katello_ssl_tool.rb
+++ b/lib/puppet/provider/katello_ssl_tool.rb
@@ -18,6 +18,17 @@ module Puppet::Provider::KatelloSslTool
       deploy!   if deploy?
     end
 
+    def destroy
+      files_to_deploy.each do |file|
+        File.delete(file) if File.exist?(file)
+      end
+
+      output = execute([:rpm, '-q', rpmfile_base_name], failonfail: false)
+      if output.exitstatus == 0
+        rpm('-e', rpmfile_base_name)
+      end
+    end
+
     def self.details(cert_name)
       details = { :pubkey  => pubkey(cert_name),
                   :privkey   => privkey(cert_name) }
@@ -173,11 +184,16 @@ module Puppet::Provider::KatelloSslTool
 
     def exists?
       return false unless File.exists?(resource[:path])
+      return false unless File.exists?(source_path)
       expected_content_processed == current_content
     end
 
     def create
       File.open(resource[:path], "w", mode) { |f| f << expected_content_processed }
+    end
+
+    def destroy
+      File.delete(resource[:path]) if File.exist?(resource[:path])
     end
 
     protected
--- a/lib/puppet/provider/key_bundle/katello_ssl_tool.rb
+++ b/lib/puppet/provider/key_bundle/katello_ssl_tool.rb
@@ -2,6 +2,13 @@ require File.expand_path('../../katello_ssl_tool', __FILE__)
 
 Puppet::Type.type(:key_bundle).provide(:katello_ssl_tool, :parent => Puppet::Provider::KatelloSslTool::CertFile) do
 
+  def exists?
+    return false unless File.exists?(resource[:path])
+    return false unless File.exists?(privkey_source_path)
+    return false unless File.exists?(pubkey_source_path)
+    expected_content_processed == current_content
+  end
+
   protected
 
   def expected_content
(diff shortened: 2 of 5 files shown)


ehelms:
To see this change in action check out https://github.com/theforeman/puppet-certs/pull/366

lib/puppet_x/certs/common.rb
```diff
@@ -86,8 +86,15 @@ module Common
         newparam(:key_pair) do
           validate do |value|
             param_resource = resource.catalog.resource(value.to_s)
-            if param_resource && !['Puppet::Type::Ca', 'Puppet::Type::Cert'].include?(param_resource.class.to_s)
-              raise ArgumentError, "Expected Ca or Cert resource, got #{param_resource.class} #{param_resource.inspect}"
+
+            if param_resource.is_a?(Puppet::Resource)
+              param_resource_type = param_resource.resource_type
+            else
+              param_resource_type = param_resource.to_resource.resource_type
```
ekohl, line 93: This will fail if no `param_resource` was passed in. I think this whole block should be inside `if param_resource`

lib/puppet/provider/katello_ssl_tool.rb
```diff
@@ -18,6 +18,16 @@ def create
       deploy!   if deploy?
     end
 
+    def destroy
+      files_to_deploy.each do |file|
+        File.delete(file) if File.exist?(file)
+      end
+
+      if(system("rpm -q #{rpmfile_base_name} &>/dev/null"))
```
ekohl, line 26: Why is a call to `system()` needed? From reading the Puppet code I think this should work
```suggestion
      output = execute([:rpm, '-q', rpmfile_base_name], failonfail: false)
      if output.exitstatus == 0
```

ekohl (changes requested):

lib/puppet/provider/katello_ssl_tool.rb
ehelms: Thanks for that, to be honest I had largely copied the previous code from another spot in the file and the code in here is many, many years old.

ekohl (approved):
Reading the katello_ssl_tool provider makes me think we should generate a `package` resource but we probably don't want to invest too much time into that.

ehelms:
> Reading the katello_ssl_tool provider makes me think we should generate a `package` resource but we probably don't want to invest too much time into that.

Yep, solid idea if I wasn't trying to actively get rid of the packages all together.

lib/puppet_x/certs/common.rb
```diff
@@ -86,8 +86,17 @@ module Common
...
             param_resource = resource.catalog.resource(value.to_s)
-            if param_resource && !['Puppet::Type::Ca', 'Puppet::Type::Cert'].include?(param_resource.class.to_s)
-              raise ArgumentError, "Expected Ca or Cert resource, got #{param_resource.class} #{param_resource.inspect}"
+
+            if param_resource
+              if param_resource.is_a?(Puppet::Resource)
+                param_resource_type = param_resource.resource_type
+              else
+                param_resource_type = param_resource.to_resource.resource_type
+              end
```
ekohl, line 95:
```suggestion
              param_resource_type = if param_resource.is_a?(Puppet::Resource)
                                      param_resource.resource_type
                                    else
                                      param_resource.to_resource.resource_type
                                    end
```

lib/puppet/provider/key_bundle/katello_ssl_tool.rb
```diff
@@ -2,6 +2,13 @@
 
 Puppet::Type.type(:key_bundle).provide(:katello_ssl_tool, :parent => Puppet::Provider::KatelloSslTool::CertFile) do
 
+  def exists?
+    return false unless File.exists?(resource[:path])
+    return false unless File.exists?(privkey_source_path)
+    return false unless File.exists?(pubkey_source_path)
+    expected_content_processed == current_content
```
ekohl, line 9: Is this really part of `exists?` Isn't this more `insync?` behavior?

ehelms: I could see that, but this is taken from https://github.com/theforeman/puppet-certs/pull/365/files#diff-9f93cd98c4a64f890a47782b00ef580ea9b901d66db370cd21c399c63991a986R185 and so my intent is not to change this behavior at that level.

PR #367, opened 2021-06-29
Refs #32585: Allow producing certificate subject with spaces between …

ehelms:
…elements

Adds an option to produce a certificate subject with spaces between
elements. This is necessary if a certificate has been generated in
standard format rather than RFC2253 format. One specific use case is
when Artemis does strict checking of the certificate subject with
the subject in the /etc/tomcat/cert-users.properties file. They
need to be identical.


ekohl (commented):
I think that the analysis that it depends on the generated certificates is odd. It looked to me that RFC2253 is *always* without spaces. For example:
```ruby
OpenSSL::X509::Name.parse('DC=example, CN=nobody').to_s(OpenSSL::X509::Name::RFC2253)
=> "DC=example,CN=nobody"
```

RFC2253 doesn't appear to mention spaces.

There is also the `compat` format which appears to be what you're looking for:
```
OpenSSL::X509::Name.parse('DC=example,CN=nobody').to_s(OpenSSL::X509::Name::COMPAT)
=> "DC=example, CN=nobody"
```

spec/functions/certs_certificate_subject_spec.rb
```diff
@@ -1,27 +1,12 @@
...
-      7WCxGGtzmw5s5SxBkyY/buGDr+kx52yULl6ZrnJD6PfR30X+8G3ltvmaCQllYadX
-      eprUs5H2WDnTUUE78+cf1JK29Zs9it/l4t2uLc5Z94oXosFLkTKw6ZSB3X9J
-      -----END CERTIFICATE-----
-    CERT
+  let(:standard_cert) do
+    File.read('./fixtures/certificate_subject/certs/standard_format.crt')
+  end
+
+  let(:rfc2253_cert) do
+    File.read('./fixtures/certificate_subject/certs/standard_format.crt')
```
wbclark, line 9:
```suggestion
    File.read('./fixtures/certificate_subject/certs/rfc2253_format.crt')
```

spec/functions/certs_certificate_subject_spec.rb
```diff
@@ -1,27 +1,12 @@
...
-      90t5AgMBAAGjLzAtMAsGA1UdDwQEAwIFIDAeBgNVHREEFzAVghNmb3JlbWFuLmV4
-      YW1wbGUuY29tMA0GCSqGSIb3DQEBCwUAA4IBAQBlIEA2CfZIa8LtUYlDwa5v+5Wf
-      1ktmYTRtgEI+922T/eTB8uH1//VxpfK5ynljao7SNVcX+74Q+YH/4Ci4OfZvE5vA
-      1IJXog5bfE4mVc1qXhH7TokBQx1L6vtUh9OaTGpBAVnS3J5jLw6+Tdi9FOeZdKHZ
-      FvMnyZ7MQ6VjbLZsTy49o87Nstqkle48ivwSFrDU1cDN+6S/DUdHQnh8XtPB1PMh
-      7WCxGGtzmw5s5SxBkyY/buGDr+kx52yULl6ZrnJD6PfR30X+8G3ltvmaCQllYadX
-      eprUs5H2WDnTUUE78+cf1JK29Zs9it/l4t2uLc5Z94oXosFLkTKw6ZSB3X9J
-      -----END CERTIFICATE-----
-    CERT
+  let(:standard_cert) do
```
ekohl, line 4: You can also consider a different layout:
```ruby
require 'spec_helper'

describe 'certs::certificate_subject' do
  it 'should exist' { ... }

  context 'with certificate' do
    before do
      file = class_double("File")
      allow(File).to receive(:read).and_call_original
      allow(File).to receive(:read).with("/tmp/client_cert.crt").and_return(certificate)
    end

    context 'in a standard format' do
      let(:certificate) { File.read('...') }

      it { ... }
    end

    context 'in RFC2253 format' do
      let(:certificate) { File.read('...') }

      it { is_expected.to run.with_params('/tmp/client_cert.crt', true).and_return("O=Test Org, C=US, CN=foreman.example.com") }
    end

    context 'in an invalid format' do
      let(:certificate) { 'aab32433adfad' }

      it { ... }
    end
  end
end
```

It's briefer just to keep it readable. Mostly look at the indenting.

ehelms:
This one is hard to explain through tests and code because I can't *exactly* represent the situation that happens. I will do my best to try to explain though.

Artemis requires the defined user, when using certificate auth, to be present in the `cert-users.properties` file in RFC2253 format (https://activemq.apache.org/components/artemis/documentation/latest/security.html -- CertificateLoginModule section):

> Each username is mapped to a subject DN, encoded as a string (where the string encoding is specified by RFC 2253). For example, the system username is mapped to the CN=system,O=Progress,C=US subject DN. When performing authentication, the plug-in extracts the subject DN from the received certificate, converts it to the standard string format, and compares it with the subject DNs in the users.properties file by testing for string equality. Consequently, you must be careful to ensure that the subject DNs appearing in the users.properties file are an exact match for the subject DNs extracted from the user certificates.

What happens in our case is, our certificate are generated in standard format and not in RFC 2253 format. To have generated certificates in RFC 2253 format, we would need to be passing `-nameopt RFC2253` during certificate creation. This results in our certificate subject looking like:

```
        Subject: C = US, ST = North Carolina, O = candlepin, OU = SomeOrgUnit, CN = localhost
```

Note that there are spaces between each element. Therefore, if we were to specify the user in Artemis without those spaces, then it fails. So, we need to provide it with those spaces, if you look at a current install you see we have:

```
# cat /etc/tomcat/cert-users.properties 
katelloUser=C=US, ST=North Carolina, O=candlepin, OU=SomeOrgUnit, CN=localhost
```

This is my understanding of why we see the current failure, and from playing around on a real system. I agree this is kind of ugly and strange and largely feels like a bug inside Artemis for not normalizing the incoming certificate's subject with what is supplied but this appears to be the reality of our current situation.

ekohl:
> ` Subject: C = US, ST = North Carolina, O = candlepin, OU = SomeOrgUnit, CN = localhost`

This is just the OpenSSL representation of it, right? AFAIK on certificates the subject is ASN.1 encoded.

From the openssl man page:
> -nameopt option 
>    option which determines how the subject or issuer names are displayed. The option argument can be a single option or multiple options separated by commas. Alternatively the -nameopt switch may be used more than once to set multiple options. See the NAME OPTIONS section for more information.

Note that it says *displayed*.

As I read it now, it's really that Artemis expects the OpenSSL compat format and *NOT* RFC2253.

I've opened https://github.com/theforeman/puppet-certs/pull/368 as an example of what I mean.

ehelms:
> > ` Subject: C = US, ST = North Carolina, O = candlepin, OU = SomeOrgUnit, CN = localhost`
> 
> This is just the OpenSSL representation of it, right? AFAIK on certificates the subject is ASN.1 encoded.
> 
> From the openssl man page:
> 
> > -nameopt option
> > option which determines how the subject or issuer names are displayed. The option argument can be a single option or multiple options separated by commas. Alternatively the -nameopt switch may be used more than once to set multiple options. See the NAME OPTIONS section for more information.
> 
> Note that it says _displayed_.
> 
> As I read it now, it's really that Artemis expects the OpenSSL compat format and _NOT_ RFC2253.
> 
> I've opened #368 as an example of what I mean.

This is not correct. Artemis does not work if using `COMPAT` format. Using COMPAT we get:

```
# cat /etc/tomcat/cert-users.properties 
katelloUser=C=US, ST=North Carolina, O=candlepin, OU=SomeOrgUnit, CN=localhost
```
Using RFC2253, we get:

```
# cat /etc/tomcat/cert-users.properties 
katelloUser=CN=localhost,OU=SomeOrgUnit,O=candlepin,ST=North Carolina,C=US
```

Note that not only is the ordering different, reverse of each other, but there are no spaces between elements. However, even this fails to work. What I know works is:

```
# cat /etc/tomcat/cert-users.properties 
katelloUser=CN=localhost, OU=SomeOrgUnit, O=candlepin, ST=North Carolina, C=US
```

This seems equivalent to:

```
# openssl x509 -noout -subject -in /etc/pki/katello/certs/java-client.crt -nameopt RFC2253,sep_comma_plus_space
subject=CN=localhost, OU=SomeOrgUnit, O=candlepin, ST=North Carolina, C=US
```

ekohl:
That is interesting. It is why I chose the approach in https://github.com/theforeman/puppet-certs/pull/368 to add 2 commits: 1 that adds the tests and expected output in the current setup and then a commit that changes the output format. That makes it easier to see the impact.

ehelms:
I simplified the test cases down to the core of the new functionality and added a comment in the function that notes what the include_spaces is mimicing since it's not directly supported by Ruby's OpenSSL library.

lib/puppet/functions/certs/certificate_subject.rb
```diff
@@ -6,15 +6,25 @@
...
-    param 'String[1]', :certificate_path
+    required_param 'String[1]', :certificate_path
+    optional_param 'Optional[Boolean]', :include_spaces
     return_type 'Optional[String]'
   end
 
-  def certificate_subject(certificate_path)
+  def certificate_subject(certificate_path, include_spaces = false)
+    unless File.exist?(certificate_path)
+      Puppet.err("The file at #{certificate_path} does not exist, the certificate subject could not be calculated.")
```
ehelms, line 20: I added this check and block. I think this is better than a silent race condition that can happen.

ehelms:
There is also a race condition that would need solved here -- https://github.com/theforeman/puppet-katello/pull/424

ekohl:
Given we merged a revert commit, should this now be closed and addressed when we do re-introduce the functionality?

PR #369, opened 2021-07-02
Revert "Fixes #32585: Add function to extract Artemis client certific…

ehelms:
…ate subject dynamically"

This reverts commit 9f3f5bf4a3b9062833a279c02ac186b475d63505.

After some testing, and trying, I can't find a way to control the ordering with Deferred object here to ensure that it gets calculated after the file has been created. Therefore I am suggesting we revert this change while I look for a better solution. I am thinking that solution is going to have to be a type and provider to properly control ordering and existence. I think that will have to be done similarly to https://github.com/theforeman/puppet-candlepin/pull/198 and managing the `cert-users` file.


ekohl (commented):
Another possible solution may be [datacat](https://github.com/richardc/puppet-datacat) or some concat pattern.

Essentially create a defined type:
```puppet
define candlepin::cert_user (
  String[1] $value,
  String[1] $name = $title,
) {
  # some type to add a single line to cert-users.properties
}
```

It could even do both: insert the certificate into the truststore and add the entry. Overall pulling it into the Candlepin module makes sense to me.

ekohl (approved):
I meant to approve this as well ... Let's revert it for now while we look for a better solution.

PR #370, opened 2021-07-13
Deploy Apache certificates from build directory

ehelms:
This introduces a new wrapper object key_pair for deploying and
managing a certificate and private key based on input files. This change
skips the creation and installation of RPMs, instead opting to deploy
the key and certificate directly from the build directory. This skips
a lot of overhead that comes with the RPMs.


ehelms:
This is a simpler version of https://github.com/theforeman/puppet-certs/pull/348 without using types and relying on `file` type only. I've opened this one to contrast the code vs. simplicity. 

One thing I think I need to take for a test drive is applying this pattern to some of the services that have tended to be a bit more fickle.

manifests/key_pair.pp
```diff
@@ -0,0 +1,31 @@
+define certs::key_pair (
```
ekohl, line 1: I think the very similar name is confusing but I don't know what a better name would be.

Can we also declare this as a private API?
```suggestion
# @summary Deploy a certificate and its matching key
# @api private
define certs::key_pair (
```

manifests/key_pair.pp
```diff
@@ -0,0 +1,31 @@
...
+  $key_file,
+  $cert_file,
+  $key_ensure  = 'present',
+  $key_owner   = undef,
+  $key_group   = undef,
+  $key_mode    = undef,
+  $cert_ensure = 'present',
+  $cert_owner  = undef,
+  $cert_group  = undef,
+  $cert_mode   = undef,
```
ekohl, line 12: Can we get into the habit of always adding data types to new parameters?

I also think we should set a mode by default for both types. This reduces the risk of accidentally exposing private keys.

manifests/key_pair.pp
```diff
@@ -0,0 +1,31 @@
...
+) {
+
+  file { $key_file:
```
ekohl, line 15: Should this always have `show_diff => false`?

ekohl (commented):
I lost track of it, but was there ever a PoC where we change `deploy` on the cert resource to manage the files in the final location? Would that make sense?

On the other hand, I like the simplicity of this solution.

ehelms:
> I lost track of it, but was there ever a PoC where we change `deploy` on the cert resource to manage the files in the final location? Would that make sense?

I think that could work, but I like the clean separation of generation and deployment. I also like keeping deployment simple as its then easy to translate that into the service modules.

manifests/key_pair.pp
ehelms: Good call out, there is a request for this out there in bug land.

manifests/key_pair.pp
ehelms: - cert_key
- key_cert

If I migrate everything at once, I could drop the old `key_pair` and then it wouldn't clash. This is a bit of whether we want to try to do a piece at a time, or switch everything to this new method together.

ekohl: I think it makes sense to submit this as a small PR now, but we should really switch all at once IMHO. I think that's easier to deal with.

ehelms:
This now updates all classes to deploy from build directory. There are some changes to class interfaces that remove unused parameters. Each class has a test added to ensure what happens when deploy is false. In some cases, there are a certificates that no longer need to exist and those have been preserved via `key_pair` and set to be absent.

manifests/key_pair.pp
```diff
@@ -0,0 +1,34 @@
+# @summary Deploy a certificate and its matching key
+# @api private
+define certs::key_pair (
+  Stdlib::Absolutepath $source_dir,
+  Stdlib::Absolutepath $key_file,
+  Stdlib::Absolutepath $cert_file,
+  Enum['present', 'absent'] $key_ensure = 'present',
+  String $key_owner = undef,
+  String $key_group = undef,
+  Stdlib::Filemode $key_mode = '400',
```
wbclark, line 10: Here the `$key_mode` defaults to `400` while the `$cert_mode` defaults to `440`

However when this is used under `if $deploy { ... }` in apache.pp, both are set to `0440`. Is there a reason not to use that as the default?

I will check how this is done in the other commits in this PR to better understand.

ehelms: I think my idea had been to default to the most secure mode possible, restricting to read access by the user only. And then force anybody using this to override this default if they need something else. In a lot of our cases, we do `root:service_user` and the `service_user` (e.g. apache) needs to read the key thus making the permissions `440` rather than `400`.

wbclark:
In manifests/candlepin.pp, the `$deploy` parameter is still used inside `cert { $java_client_cert_name: ... }` unlike in other places where it is hardcoded to `false` -- unsure if that is intended or an oversight

I saw the same in manifests/foreman_proxy.pp and wanted to check that it's not an oversight.

Most other instances of `cert { ... }` class have the change to `deploy => false` throughout these commits.

ehelms:
> In manifests/candlepin.pp, the `$deploy` parameter is still used inside `cert { $java_client_cert_name: ... }` unlike in other places where it is hardcoded to `false` -- unsure if that is intended or an oversight
> 
> I saw the same in manifests/foreman_proxy.pp and wanted to check that it's not an oversight.
> 
> Most other instances of `cert { ... }` class have the change to `deploy => false` throughout these commits.

Both of those have `state => absent` so the deploy parameter does not matter. I can though switch it to false to make this clearer.

wbclark:
> Both of those have `state => absent` so the deploy parameter does not matter. I can though switch it to false to make this clearer.

Given https://github.com/theforeman/puppet-certs/blob/60581ed93e93f8a451db3442d77a97b9b92e4017/lib/puppet/provider/katello_ssl_tool.rb#L72-L73 , could you drop it entirely when you don't need it because `state => absent` ?

Another way of asking the same question: does `resource[:deploy]` evaluate to `nil` when the parameter is not passed, or does the type/provider really require it regardless?

ehelms:
I think the absent means it skips the exist and create calls that would call `deploy?`. However, to bring some clarity I will update.

ekohl (commented):
We talked about n-1 support but I don't see it here. Can we still generate a tarball with RPMs for n-1 with this?

ehelms:
This requires (and is currently rebased on) https://github.com/theforeman/puppet-certs/pull/380

ehelms:
> We talked about n-1 support but I don't see it here. Can we still generate a tarball with RPMs for n-1 with this?

I am thinking on this problem. As we have seen, its not just can we generate RPMs, but for N-1 support its what about certificates we drop version to version. I am first trying to think of a test case to generate to help ensure N-1(2) support otherwise trying to maintain with some modicum of sanity it will be nigh impossible.

ehelms:
Opened https://github.com/theforeman/puppet-certs/pull/381 to add some regression tests prior to this change going in. This set of changes does *not* drop the creation of RPMs inside the build directories, just stops the install of those RPMs onto the system. For N-1 support, that means the RPMs still exist and are included in any tar archives that are generated. The biggest risk then is any certificates that we drop support for. This change does not drop any certificates from getting built. 

Thus, we are safe against N-1 with this change. We can/should add regression testing via https://github.com/theforeman/puppet-certs/pull/381 to help ensure this now and in the future.

After this change goes in, I will look at changes to not generate RPMs unless requested by the code. That will end up with a cleaner use case in the end. Right now the RPM generation is overhead since they are not being used with this change (except for N-1 support) but the performance difference is minimal.

lib/puppet/type/cert_key_bundle.rb
```diff
@@ -0,0 +1,98 @@
...
+    desc "Path to the file that will contain the certificate and private key bundle"
+    isrequired
+  end
+
+  newparam(:certificate) do
+    desc "Path to certificate file to include in bundle"
+    isrequired
+
+    def insync?(is)
+      File.read(self[:path]).include?(File.read(should))
```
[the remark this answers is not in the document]
ehelms: I will apply these and the other comments over here on the stand-alone PR -- https://github.com/theforeman/puppet-certs/pull/380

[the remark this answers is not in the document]
ehelms: Correct. Since this is the private key and certificate in a single file I figure this `insync?` check should just check that the certificate is present.

ehelms:
This has been rebased to include the updated testing of the tarball, and the new cert_key_bundle with additional `foreman_proxy.pp` tests.

manifests/ca.pp
```diff
@@ -26,8 +27,8 @@
   file { $ca_key_password_file:
     ensure    => file,
     content   => $ca_key_password,
-    owner     => 'root',
-    group     => 'root',
+    owner     => $owner,
+    group     => $group,
```
ekohl, line 31: This doesn't look right. In the Katello scenario we set the group to foreman (https://github.com/theforeman/foreman-installer/blob/bef5bd620a5b1494d3b5779030bc6b29e861645c/config/katello-answers.yaml#L12-L13) and Foreman doesn't need to read this file.

manifests/ca.pp
```diff
@@ -41,79 +42,60 @@
...
+    }
+
+    file { $certs::katello_default_ca_cert:
       ensure => file,
-      owner  => 'root',
+      source => "${certs::ssl_build_dir}/${default_ca_name}.crt",
+      owner  => $owner,
       group  => $group,
-      mode   => '0644',
+      mode   => '0440',
```
ekohl, line 90: Why should this only be visible to the Foreman user? IMHO this should be root:root with mode 0644. Same below with the server CA cert.

manifests/foreman.pp
```diff
@@ -38,35 +38,32 @@
...
+      cert_owner => $owner,
+      cert_group => $group,
+      cert_mode  => '0440',
+      require    => Cert[$client_cert_name],
     }
 
     file { $ssl_ca_cert:
       ensure  => file,
+      source  => $server_ca,
       owner   => $owner,
```
ekohl, line 63: Unrelated, but `root` is probably a better owner here.

manifests/key_pair.pp
ekohl: The most secure mode is `0440` where the owner is `root` and only the group can read it. This disallows the service itself from overwriting it while still reading it. Remember that if you are the owner of a file, you can change the mode to anything else as well.

This also means you can default the `key_owner` and `cert_owner` to `root` if you always manage them anyway.

manifests/key_pair.pp
ekohl: Now that the PR removes the old one, can we rename it back to `cert::keypair`?

manifests/puppet.pp
```diff
@@ -34,35 +36,39 @@
...
     build_dir     => $certs::ssl_build_dir,
   }
 
   if $deploy {
     file { "${pki_dir}/puppet":
       ensure => directory,
-      owner  => 'puppet',
+      owner  => $owner,
+      group  => $group,
       mode   => '0700',
```
ekohl, line 49: This breaks things since Puppet is longer able to read these files:
```suggestion
      owner  => $owner,
      group  => $group,
      mode   => '0750',
```

manifests/init.pp
```diff
@@ -125,5 +125,4 @@
   Class['certs::ca']
 
   $default_ca = $certs::ca::default_ca
-  $server_ca = $certs::ca::server_ca
```
ekohl, line 128: Note this is used here:
https://github.com/theforeman/puppet-foreman_proxy_content/blob/37a6e9eaa058fefd1788625ad5def7243cb351e5/manifests/bootstrap_rpm.pp#L15

PR #371, opened 2021-07-13
Fixes #32382: Use certs::foreman client certificate to communicate wi…

ehelms:
…th Artemis

This hardcodes the Artemis client DN to the Foreman client certificate
and ensures that same certificate is in the truststore. This also
ensures the removal of the previous client certificates.

### DIFF
--- a/manifests/candlepin.pp
+++ b/manifests/candlepin.pp
@@ -24,12 +24,14 @@ class certs::candlepin (
   $group                    = 'tomcat',
   $client_keypair_group     = 'tomcat',
 ) inherits certs {
+  include certs::foreman
+
   $java_client_cert_name = 'java-client'
   $artemis_alias = 'artemis-client'
-  $artemis_client_dn = "CN=${hostname}, OU=${org_unit}, O=candlepin, ST=${state}, C=${country}"
+  $artemis_client_dn = $certs::foreman::client_dn
 
   cert { $java_client_cert_name:
-    ensure        => present,
+    ensure        => absent,
     hostname      => $hostname,
     cname         => $cname,
     country       => $country,
@@ -72,9 +74,8 @@ class certs::candlepin (
   $truststore_password = extlib::cache_data('foreman_cache_data', $truststore_password_file, extlib::random_password(32))
   $keystore_password_path = "${pki_dir}/${keystore_password_file}"
   $truststore_password_path = "${pki_dir}/${truststore_password_file}"
-  $client_req = "${pki_dir}/java-client.req"
-  $client_cert = "${pki_dir}/certs/${java_client_cert_name}.crt"
-  $client_key = "${pki_dir}/private/${java_client_cert_name}.key"
+  $client_key = $certs::foreman::client_key
+  $client_cert = $certs::foreman::client_cert
   $alias = 'candlepin-ca'
 
   if $deploy {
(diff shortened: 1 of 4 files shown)


manifests/candlepin.pp
```diff
@@ -171,10 +173,12 @@
       certificate   => $ca_cert,
     }
 
+    include certs::foreman
     truststore_certificate { "${truststore}:${artemis_alias}":
       ensure        => present,
       password_file => $truststore_password_path,
-      certificate   => $client_cert,
+      certificate   => $certs::foreman::client_cert,
+      require       => File[$certs::foreman::client_cert],
```
ekohl, line 181: Don't we have some autorequire for this?
https://github.com/theforeman/puppet-certs/blob/e8d0346188b317c5ba354b92fc25fdb3a9285adc/lib/puppet/type/truststore_certificate.rb#L40-L42
```suggestion
```

spec/acceptance/candlepin_spec.rb
```diff
@@ -1,11 +1,25 @@
 require 'spec_helper_acceptance'
 
 describe 'certs' do
+  FQDN = fact('fqdn')
```
ekohl, line 4: This is generating warnings: you can't redefine constants.

ekohl (changes requested):

spec/acceptance/candlepin_spec.rb
ehelms: Those warnings already exist because of this pattern in the other test files. So I'd prefer to keep this pattern and then fix it later across all the test files.

One a side note, it sucks this constant is not localized to the describe block as one might think.

ekohl: Agreed, but that's rspec for you.

ehelms: Proper fix -- https://github.com/theforeman/puppet-certs/pull/372

manifests/candlepin.pp
```diff
@@ -72,9 +74,8 @@
   $truststore_password = extlib::cache_data('foreman_cache_data', $truststore_password_file, extlib::random_password(32))
   $keystore_password_path = "${pki_dir}/${keystore_password_file}"
   $truststore_password_path = "${pki_dir}/${truststore_password_file}"
-  $client_req = "${pki_dir}/java-client.req"
-  $client_cert = "${pki_dir}/certs/${java_client_cert_name}.crt"
-  $client_key = "${pki_dir}/private/${java_client_cert_name}.key"
+  $client_key = $certs::foreman::client_cert
+  $client_cert = $certs::foreman::client_key
```
ehelms, line 78: The reason to redefine these is it allows us to keep the current wiring code intact -- https://github.com/theforeman/puppet-katello/blob/master/manifests/application.pp#L50-L51

manifests/foreman.pp
```diff
@@ -18,6 +18,7 @@
 ) inherits certs {
 
   $client_cert_name = "${hostname}-foreman-client"
+  $client_dn = "CN=${hostname}, OU=PUPPET, O=FOREMAN, ST=${state}, C=${country}"
```
ekohl, line 21: city is also a parameter. Is that `undef` by default so it doesn't show up? Do we have users who do pass in `$city`?

How about adding `$org` and `$org_unit` as class parameters and reusing them here?

ehelms: `$city` is defined, it just does not appear in the DN. I've never thought to go look why if its a parameter though. That should not hold this up but I could go investigate it separately. I've added the parameters for org and org_unit.

ekohl (approved):
I do think this is an intermediate solution, but it feels like a step in the right direction.

PR #375, opened 2021-07-21
Manage the certs::foreman ssl_ca_file

ehelms:
This manages the ssl_ca_file in certs::foreman ensuring the right
owner and group are set and allowing users of this class to customize
it if needed.

@ekohl Needed to implement https://github.com/ekohl/puppet-katello_devel/commit/8e2f945cd4e70d8e39737ad5a90c3e431442f22b otherwise the `proxy_ca.pem` file would end up owned by root:root.

### DIFF
--- a/manifests/foreman.pp
+++ b/manifests/foreman.pp
@@ -60,5 +60,13 @@ class certs::foreman (
     pubkey { $ssl_ca_cert:
       key_pair => $server_ca,
     }
+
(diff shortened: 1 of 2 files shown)


jturel (approved):
This worked in conjunction with the recent katello_devel fix

ekohl (approved):

PR #381, opened 2021-09-04
Add test to check tar archive contents

ehelms:
Aims to allow testing for regressions when certificates are added
or removed when aiming to support Katello N-1 support. Additionally
aims to detect if new items are added that should be checked for.

### DIFF
--- a/spec/support/acceptance/tar.rb
+++ b/spec/support/acceptance/tar.rb
@@ -1,0 +1,37 @@
+begin
+  require 'serverspec'
+rescue LoadError
+  # Not using acceptance tests
+else
+  module Serverspec
+    module Type
+      class Tar < Command
+        def contents
+          command_result.stdout.split("\n")
+        end
+
+        def exist?
+          @runner.check_file_exists(@name)
+        end
+
+        private
+
+        def command
+          "tar -tf #{@name}"
+        end
+
+        def command_result
+          @command_result ||= @runner.run_command(command)
+        end
+      end
+    end
+
+    module Helper
+      module Type
+        def tar(*args)
+          Serverspec::Type::Tar.new(*args)
+        end
+      end
+    end
+  end
+end
--- a/spec/acceptance/certs_tar_extract_spec.rb
+++ b/spec/acceptance/certs_tar_extract_spec.rb
@@ -37,6 +37,10 @@ describe 'certs with tar archive' do
     apply_manifest(install_certs, catch_failures: true)
   end
 
+  after(:context) do
+    on default, 'yum -y remove foreman-proxy.example.com*noarch*'
+  end
+
   describe x509_certificate('/etc/pki/katello/certs/katello-apache.crt') do
     it { should be_certificate }
     it { should be_valid }
--- a/spec/acceptance/foreman_proxy_content_spec.rb
+++ b/spec/acceptance/foreman_proxy_content_spec.rb
@@ -24,8 +24,42 @@ describe 'certs::foreman_proxy_content' do
       apply_manifest(pp, catch_failures: true)
     end
 
-    describe file('/root/foreman-proxy.example.com.tar.gz') do
+    let(:expected_files_in_tar) do
+      [
+        'ssl-build/katello-default-ca-1.0-1.noarch.rpm',
+        'ssl-build/katello-server-ca-1.0-1.noarch.rpm',
+        'ssl-build/katello-default-ca.crt',
+        'ssl-build/katello-server-ca.crt',
+        'ssl-build/foreman-proxy.example.com/foreman-proxy.example.com-apache-1.0-1.noarch.rpm',
+        'ssl-build/foreman-proxy.example.com/foreman-proxy.example.com-foreman-client-1.0-1.noarch.rpm',
+        'ssl-build/foreman-proxy.example.com/foreman-proxy.example.com-foreman-proxy-1.0-1.noarch.rpm',
+        'ssl-build/foreman-proxy.example.com/foreman-proxy.example.com-foreman-proxy-client-1.0-1.noarch.rpm',
+        'ssl-build/foreman-proxy.example.com/foreman-proxy.example.com-puppet-client-1.0-1.noarch.rpm',
+        'ssl-build/foreman-proxy.example.com/foreman-proxy.example.com-qpid-broker-1.0-1.noarch.rpm',
+        'ssl-build/foreman-proxy.example.com/foreman-proxy.example.com-qpid-router-client-1.0-1.noarch.rpm',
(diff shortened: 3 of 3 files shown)


spec/support/acceptance/tar.rb
```diff
@@ -0,0 +1,37 @@
...
+
+		module Helper
+			module Type
+				def tar(*args)
+					Serverspec::Type::Tar.new(*args)
+				end
+			end
+		end
+	end
+end
```
wbclark, line 37:
```suggestion
else
  module Serverspec
    module Type
      class Tar < Command
        def contents
          command_result.stdout
        end

        def contents_count
          command_result.stdout.split("\n").length
        end

        private

        def command
          "tar -tvf #{@name}"
        end

        def command_result
          @command_result ||= @runner.run_command(command)
        end
      end 
    end

    module Helper
      module Type
        def tar(*args)
          Serverspec::Type::Tar.new(*args)
        end
      end
    end
  end
end
```

New `~/.vimrc` ?

wbclark:
For some reason this is getting the apache cert with release `foreman-proxy.example.com-apache-1.0-2.noarch.rpm` when all the others have the release `1.0-1`

spec/support/acceptance/tar.rb
```diff
@@ -0,0 +1,37 @@
...
+
+        private
+
+        def command
+          "tar -tvf #{@name}"
+        end
+
+        def command_result
+          @command_result ||= @runner.run_command(command)
+        end
```
ekohl, line 25: Since https://github.com/mizzy/serverspec/commit/2e7e24a90522df8e435a689bd59d110ed56799d3 (included in serverspec 2.41.8) this is technically redundant. You may need to make `command` `protected` instead of `private`. However, we didn't change other ocurrances so I'm fine either way.

spec/support/acceptance/tar.rb
```diff
@@ -0,0 +1,37 @@
+begin
+  require 'serverspec'
+rescue LoadError
+  # Not using acceptance tests
+else
+  module Serverspec
+    module Type
+      class Tar < Command
+        def contents
+          command_result.stdout
```
ekohl, line 10: Why do you treat this as a string? Shouldn't you split it into multiple lines? Then you can use `its(:contents) { should include('ssl-build/foreman-proxy.example.com/foreman-proxy.example.com-qpid-router-server.key') }`.

ekohl (commented):
Overall :+1: for testing the tar contents

ehelms:
On Mon, Sep 6, 2021, 5:16 AM Ewoud Kohl van Wijngaarden <
***@***.***> wrote:

> ***@***.**** commented on this pull request.
>
> Overall 👍 for testing the tar contents
> ------------------------------
>
> In spec/support/acceptance/tar.rb
> <https://github.com/theforeman/puppet-certs/pull/381#discussion_r702740674>
> :
>
> > +        def command_result
>
> +          @command_result ||= @runner.run_command(command)
>
> +        end
>
>
> Since ***@***.***
> <https://github.com/mizzy/serverspec/commit/2e7e24a90522df8e435a689bd59d110ed56799d3>
> (included in serverspec 2.41.8) this is technically redundant. You may need
> to make command protected instead of private. However, we didn't change
> other ocurrances so I'm fine either way.
> ------------------------------
>
> In spec/support/acceptance/tar.rb
> <https://github.com/theforeman/puppet-certs/pull/381#discussion_r702741875>
> :
>
> > @@ -0,0 +1,37 @@
>
> +begin
>
> +  require 'serverspec'
>
> +rescue LoadError
>
> +  # Not using acceptance tests
>
> +else
>
> +  module Serverspec
>
> +    module Type
>
> +      class Tar < Command
>
> +        def contents
>
> +          command_result.stdout
>
>
> Why do you treat this as a string? Shouldn't you split it into multiple
> lines? Then you can use its(:contents) { should include('ssl-build/
> foreman-proxy.example.com/foreman-proxy.example.com-qpid-router-server.key')
> }.
>
Simplicity, mostly. I figured start with what it returns and avoid having
to parse the data, do matchers or splits. At the time, I did not figure I
gained much by doing an include vs a match.

ekohl:
You can actually do `include` on a string as well, so even today I'd say that `include` is better than match. It's a pattern that we use a lot, but IMHO is actually wrong. See https://relishapp.com/rspec/rspec-expectations/v/3-10/docs/built-in-matchers/include-matcher#string-usage as well.

spec/support/acceptance/tar.rb
ekohl: The reason I suggest it is that you can then use `match_array`. https://relishapp.com/rspec/rspec-expectations/v/3-10/docs/built-in-matchers. Something like:
```ruby
let(:expected_files_in_tar) do
  [
    ...
  ]
end

describe tar('/path') do
   its(:contents) { should match_array(expected_files_in_tar) }
end
```

That also removes the need for `contents_count` since it will check the length. It will also show you which additional entries are there. That is something you don't have today.

ehelms: Ahh ha, excellent way to solve my problem. Thanks. I have implemented this and added a direct exist check.

spec/support/acceptance/tar.rb
```diff
@@ -0,0 +1,39 @@
...
+  require 'serverspec'
+rescue LoadError
+  # Not using acceptance tests
+else
+  module Serverspec
+    module Type
+      class Tar < Command
+        def contents
+          command_result.stdout.split("\n").collect do |line|
+            line.split(' ')[5]
```
ekohl, line 11: If you use `tar -tf` (as opposed to `tar -tvf`) then it doesn't list the other columns, just the filenames. That means it also deals well with spaces in filenames.

ekohl (approved):
Thanks!

ekohl:
@wbclark want to take a look as well before merging?

wbclark:
This looks great to me. I had one remaining question:

Looking at an older `$hostname-certs.tar` archive from a lab system, it can contain potentially many RPM releases. Now that we include both RPMs as well as unpackaged CA certificates, certificates, and keys, what should be the behavior when using `regenerate => true` and do we need to test it?

Was having the complete release history of the RPMs ever actually necessary and if so, is it sufficient in the newer design to have the unversioned raw certificates, keys, etc?

ehelms:
So regenerate will force recreate everything -- certificates, keys and thus the RPMs. They all end up living in the same directory as nothing gets cleaned up. This translates to being put into the new tarball that gets generated as well. That being said, we could look into in the future some pruning or creating of that tarball smarter. For this change, and the change to deploy certs from the build directory rather than the RPMs I do not think any changes are necessary at present.

PR #386, opened 2021-12-21
Fixes #34189: Unencrypt CA key when deploying for Candlepin

ehelms:
Introduces a new type and provider for managing a private key file.
Has the ability to decrypt the private key file when deploying
which is needed for Candlepin deployments.

### DIFF
--- a/lib/puppet/provider/private_key/openssl.rb
+++ b/lib/puppet/provider/private_key/openssl.rb
@@ -1,0 +1,45 @@
+Puppet::Type.type(:private_key).provide(:openssl) do
+  commands :openssl => 'openssl'
+
+  def create
+    File.write(resource[:path], private_key_content)
+  end
+
+  def destroy
+    delete_private_key
+  end
+
+  def exists?
+    File.exist?(resource[:path]) && File.read(resource[:path]) == private_key_content
+  end
+
+  private
+
+  def private_key_content
+    key_path = full_path(resource[:source][0])
+
+    if resource[:decrypt]
+      decrypt(key_path, resource[:password_file])
+    else
+      File.read(key_path)
+    end
+  end
+
+  def delete_private_key
+    File.unlink(resource[:path])
+  end
+
+  def full_path(source)
+    Puppet::Util.uri_to_path(URI.parse(Puppet::Util.uri_encode(source)))
+  end
+
+  def decrypt(path, password_file)
+    openssl(
+      'rsa',
+      '-in', path,
+      '-passin', "file:#{password_file}",
+      '-text'
+    )
+  end
+
+end
--- a/lib/puppet/type/private_key.rb
+++ b/lib/puppet/type/private_key.rb
@@ -1,0 +1,66 @@
+Puppet::Type.newtype(:private_key) do
+  desc 'manage a private key'
+
+  ensurable
+
+  newparam(:path, :namevar => true) do
+    desc "Path to the private key file"
+    isrequired
+  end
+
+  newparam(:password_file) do
+    desc "Path to file containing the private key file password"
+  end
+
+  newparam(:decrypt, parent: Puppet::Parameter::Boolean) do
+    desc "If true, decrypts the private key before writing to disk. Requires password_file parameter to be set"
+  end
+
+  newparam(:owner, parent: Puppet::Type::File::Owner) do
+    desc "Specifies the owner of the private key. Valid options: a string containing a username or integer containing a uid."
+  end
+
+  newparam(:group, parent: Puppet::Type::File::Group) do
+    desc "Specifies a permissions group for the private key. Valid options: a string containing a group name or integer containing a gid."
+  end
+
+  newparam(:mode, parent: Puppet::Type::File::Mode) do
+    desc "Specifies the permissions mode of the private key. Valid options: a string containing a permission mode value in octal notation."
+  end
+
+  newparam(:source, parent: Puppet::Type::File::ParameterSource) do
+    desc "Specifies the source of the private key."
+    isrequired
+  end
+
+  validate do
+    self.fail _("You cannot specify to decrypt without including a password file") if self[:decrypt] && !self[:password_file]
+  end
+
+  autorequire(:file) do
+    [self[:password_file], self[:path], self[:source]]
+  end
+
+  def generate
+    file_opts = {
+      ensure: (self[:ensure] == :absent) ? :absent : :file,
+    }
+
+    [:path,
+     :owner,
+     :group,
+     :mode].each do |param|
+      file_opts[param] = self[param] unless self[param].nil?
+    end
+
+    excluded_metaparams = [:before, :notify, :require, :subscribe, :tag]
+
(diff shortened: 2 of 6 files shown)


ehelms:
See also https://github.com/theforeman/puppet-katello/pull/437

lib/puppet/type/private_key.rb
```diff
@@ -0,0 +1,70 @@
...
+
+    excluded_metaparams = [:before, :notify, :require, :subscribe, :tag]
+
+    Puppet::Type.metaparams.each do |metaparam|
+      unless self[metaparam].nil? || excluded_metaparams.include?(metaparam)
+        file_opts[metaparam] = self[metaparam]
+      end
+    end
+
+    [Puppet::Type.type(:file).new(file_opts)]
```
ekohl, line 67: Would we need a relation that the file itself is created before we write the content? For security reasons you want to first create the file with the right permissions and only then write the content. Technically it's a very short window where it could be world readable, but it's important to remember it's there.

lib/puppet/type/private_key.rb
```diff
@@ -0,0 +1,70 @@
...
+  desc 'manage a private key'
+
+  ensurable
+
+  newparam(:path, :namevar => true) do
+    desc "Path to the private key file"
+    isrequired
+  end
+
+  newparam(:password_file) do
```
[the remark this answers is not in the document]
ehelms: Does autorequire properly handle optional parameters? And that would enforce that the password file is a puppet managed entity right?

lib/puppet/type/private_key.rb
ehelms: How would we do that? Is that what this is doing? https://github.com/puppetlabs/puppetlabs-concat/blob/main/lib/puppet/type/concat_file.rb#L184-L188

lib/puppet/type/private_key.rb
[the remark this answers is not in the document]
ekohl: It should handle `nil` which is what happens when it's optional.

lib/puppet/type/private_key.rb
ekohl: That looks like a good way to do it.

lib/puppet/provider/private_key/openssl.rb
```diff
@@ -0,0 +1,47 @@
...
+      file.write(private_key_content)
+    end
+  end
+
+  def destroy
+    delete_private_key
+  end
+
+  def exists?
+    File.exist?(resource[:path]) && File.read(resource[:path]) == private_key_content
```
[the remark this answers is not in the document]
ehelms: I think that is only available on properties, in this case the `path` is the title param and thus not able to define that insync on.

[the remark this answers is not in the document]
ekohl: Isn't that technically more the `content`, which will depend on the `source` property?

lib/puppet/type/private_key.rb
```diff
@@ -0,0 +1,75 @@
...
+    Puppet::Type.metaparams.each do |metaparam|
+      unless self[metaparam].nil? || excluded_metaparams.include?(metaparam)
+        file_opts[metaparam] = self[metaparam]
+      end
+    end
+
+    [Puppet::Type.type(:file).new(file_opts)]
+  end
+end
+
```
ekohl, line 75:
```suggestion
```

lib/puppet/type/private_key.rb
```diff
@@ -0,0 +1,75 @@
...
+  def generate
+    file_opts = {
+      ensure: (self[:ensure] == :absent) ? :absent : :file,
+    }
+
+    [:path,
+     :owner,
+     :group,
+     :show_diff,
```
ekohl, line 59: Since the content isn't managed by the `file` resource, how useful is `show_diff`?

lib/puppet/provider/private_key/openssl.rb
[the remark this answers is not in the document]
ehelms: I just not sure why i would put or specify this. `source` is also a parameter

manifests/keypair.pp
```diff
@@ -12,15 +12,19 @@
...
-    show_diff => false,
+  private_key { $key_file:
+    ensure        => $key_ensure,
+    source        => "${source_dir}/${title}.key",
+    decrypt       => $key_decrypt,
+    password_file => $key_password_file,
+    owner         => $key_owner,
+    group         => $key_group,
+    mode          => $key_mode,
+    show_diff     => false,
```
evgeni, line 27:
```suggestion
```

```
error during compilation: no parameter named 'show_diff' (file: /home/runner/work/puppet-certs/puppet-certs/spec/fixtures/modules/certs/manifests/keypair.pp, line: 19) on Private_key[/etc/foreman/client_key.pem] (file: /home/runner/work/puppet-certs/puppet-certs/spec/fixtures/modules/certs/manifests/keypair.pp, line: 19) on node fv-az216-913.3le2vxowor4era4vmqnnplo0rf.xx.internal.cloudapp.net
```

evgeni:
So I somehow thought this is only needed for EL8 FIPS, and put that back.

But we actually need this for EL7 FIPS too :sob:

ekohl (commented):
Small note, otherwise this looks good.

lib/puppet/type/private_key.rb
```diff
@@ -0,0 +1,66 @@
+Puppet::Type.newtype(:private_key) do
+  desc 'manage a private key'
```
evgeni, line 2: one thing, that I find confusing as a user: we already have a `privkey` type:

https://github.com/theforeman/puppet-certs/blob/af1c91414f5f4c98cafb78180b0a340db7e3c52a/lib/puppet/type/privkey.rb#L3-L4

but it does something totally different, albeit having a similar name?

ekohl (approved):
Let's wait for the tests. Also, don't forget to squash the commits on merge.

ekohl:
The EL8 failures are probably due to modularity:
```
  Error: /Stage[main]/Certs::Install/Package[katello-certs-tools]/ensure: change from 'purged' to 'present' failed: Execution of '/usr/bin/dnf -d 0 -e 1 -y install katello-certs-tools' returned 1: Error: Unable to find a match: katello-certs-tools
```

evgeni:
Yeah, probably needs a similar treatment like 


 https://github.com/theforeman/puppet-foreman_proxy_content/pull/396

evgeni:
> Yeah, probably needs a similar treatment like
> 
> [theforeman/puppet-foreman_proxy_content#396](https://github.com/theforeman/puppet-foreman_proxy_content/pull/396)

Let's see if https://github.com/theforeman/puppet-certs/pull/390 helps.

ekohl:
Rebased it.

ekohl:
I'm struggling to determine if this is a bugfix, enhancement or breaking change. I think the latter because a change in Katello is needed. @evgeni?

PR #390, opened 2022-02-02
use foreman::repo to setup foreman repos during CI

evgeni:


spec/setup_acceptance_node.pp
```diff
@@ -1,12 +1,5 @@
-$major = $facts['os']['release']['major']
-
-# Defaults to staging, for release, use
-# $baseurl = "https://yum.theforeman.org/releases/nightly/el${major}/x86_64/"
-$baseurl = "http://koji.katello.org/releases/yum/foreman-nightly/RHEL/${major}/x86_64/"
```
evgeni, line 5: there is no good way to use koji with `foreman::repo` -- do we really need it?

.fixtures.yml
```diff
@@ -2,4 +2,6 @@ fixtures:
   repositories:
     concat:            "https://github.com/puppetlabs/puppetlabs-concat"
     extlib:            "https://github.com/voxpupuli/puppet-extlib"
+    foreman:           "https://github.com/theforeman/puppet-foreman"
+    redis:             "https://github.com/voxpupuli/puppet-redis"
```
ekohl, line 6: Do we need Redis? As long as we're not setting it up, we can leave the module out.

spec/setup_acceptance_node.pp
ekohl: Personally I think it's not that important and I'd be happy to only use releases. Previously we needed unreleased things because Katello was broken for too long but I think we should work harder to keep it green instead.

ekohl (approved):
It's green so I'm merging it so we can move forward. Thanks!

.fixtures.yml
ekohl: Ah right, that's a good reason.

PR #394, opened 2022-03-11
Fixes #34598: Disable fips for keytool 

ehelms:
Starting with EL 8, the Java stack has FIPS support built in. This
causes the keytool utility to break on FIPS enabled machines. To solve this,
and achieve EL7-like comptability for FIPS, FIPS is disabled during the keytool
runtime via -Dcom.redhat.fips=false for each invoke of keytool.


evgeni (approved):
can we get a redmine please? ;)

ehelms:
> can we get a redmine please? ;)

Sure, here's a whole instance https://projects.theforeman.org/ :smile: 

Dumb friday jokes aside, I'll add one

evgeni (approved):
I've tested this (together with https://github.com/Katello/katello-certs-tools/pull/45) and it allowed me to deploy Katello on an FIPS enabled EL8 machine.

PR #396, opened 2022-03-26
Deploy server CA certificate for smart-proxy rather than default

ehelms:
Let's start with the basics. The Katello installation scenario allows for supplying custom certificates (ones not created by the default CA) from the user so that the public facing sites present a certificate from the users CA provider. The primary location for these to get deployed is Apache, but we also deploy them to the smart-proxy.

For the entire lifecycle, at least from looking back at code, of the custom certificates feature (~7-8 years), we have been deploying the user's custom certificate and private key for the smart-proxy service to use but have been deploying the default CA instead of the CA that signed the server certificates. I think this is generally speaking an incorrect behavior - and things have managed to just work for years now.

As soon as we try to correct this, we will notice certain client based workflows will not work anymore. For example, if a Katello client signed by the default CA presents it's certificates to the smart-proxy for remote execution and REX uses the smart-proxy CA to verify them it will now fail. 

This plays a role in how we think about deploying something new like mosquitto (see https://github.com/theforeman/puppet-foreman_proxy/pull/726) where if we want to re-use the hostname certificates we also need a combined CA certificate so that clients connecting with Katello certificates can do so. 

This change opts to deploy both CA certificates as a chain file.

### DIFF
--- a/manifests/foreman_proxy.pp
+++ b/manifests/foreman_proxy.pp
@@ -107,13 +107,26 @@ class certs::foreman_proxy (
       require    => Cert[$proxy_cert_name],
     }
 
-    file { $proxy_ca_cert:
-      ensure  => file,
-      source  => $default_ca_cert,
+    concat { $proxy_ca_cert:
+      ensure  => present,
       owner   => $owner,
       group   => $group,
       mode    => '0440',
-      require => File[$default_ca_cert],
+      require => File[$server_ca_cert],
+    }
+
+    concat::fragment { 'default_ca_cert':
+      target => $proxy_ca_cert,
+      source => $default_ca_cert,
+      order  => '01',
+    }
+
+    if $server_cert {
+      concat::fragment { 'server_ca_cert':
+        target => $proxy_ca_cert,
+        source => $server_ca_cert,
+        order  => '02',
+      }
     }
 
     certs::keypair { $foreman_proxy_client_cert_name:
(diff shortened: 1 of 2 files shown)


manifests/foreman_proxy.pp
```diff
@@ -107,13 +107,26 @@
...
-    file { $proxy_ca_cert:
-      ensure  => file,
-      source  => $default_ca_cert,
+    concat { $proxy_ca_cert:
+      ensure  => present,
       owner   => $owner,
       group   => $group,
       mode    => '0440',
-      require => File[$default_ca_cert],
+      require => File[$server_ca_cert],
```
ekohl, line 115: This require is a bit odd. Wouldn't it be on the individual fragments? I'm not sure if it does that automatically (I'm guessing not) so be sure to check that.

ekohl (commented):
Doesn't this mean both certificates are allowed for client cert auth? I think we need so evaluate the implications of that.

evgeni:
Yeah, that feels wrong, only Katello CA (or equivalent) should be used for auth, IMHO.

evgeni:
I still don't exactly understand why Mosquito would need both?

ekohl:
Just a quick thought: it's possible to present a full chain in the certificate file and not use a separate CA file. We can also consider such an approach. I don't know how our tooling handles it today, but it's not uncommon.

PR #397, opened 2022-05-03
Ensure nssdb private key changes when certificate changes

ehelms:


evgeni:
The change looks reasonable to me. Two nitpicks:
- can we have a redmine (linked up with that one BZ)
- I think the fact that `delete_private_key` also deletes the certificate deserves a comment somewhere (or a different name, I don't care too much), as otherwise the introduced `if/else` look really weird.

evgeni (approved):

PR #400, opened 2022-06-09
Drop EL7 support

ehelms:


ekohl (approved):
Code wise :+1: though I wonder if we should drop this while we still have releases out there with EL7 support. Though we likely need to once nightly drops EL7 packages.

ehelms:
> Code wise +1 though I wonder if we should drop this while we still have releases out there with EL7 support. Though we likely need to once nightly drops EL7 packages.

Yup, once we drop repositories testing breaks. So the consensus was to drop EL7 support code in the Puppet modules, then drop our pipelines tests, then our actual repositories to prevent cascading failures.

PR #401, opened 2022-06-21
Update to voxpupuli-test 5

ekohl:


manifests/apache.pp
```diff
@@ -15,11 +15,10 @@
   String $org = $certs::org,
   String $org_unit = $certs::org_unit,
   String $expiration = $certs::expiration,
-  $default_ca = $certs::default_ca,
+  Type[Ca] $default_ca = $certs::default_ca,
```
evgeni, line 18: that refers to https://github.com/theforeman/puppet-certs/blob/master/lib/puppet/type/ca.rb? interesting notation…

evgeni (approved):

manifests/apache.pp
ekohl: Yes, but we're passing in `Ca[$ca]` so I think this is the correct data type for that. I'll have to admit I looked up `Type` though.

ehelms (approved):

PR #402, opened 2022-08-01
Release 16.0.0

ekohl:


ekohl:
@ehelms is this worth a release? It's only dropping things, not adding anything we need. Doing a whole dance of bumping versions in metadata feels like too much.

CHANGELOG.md
```diff
@@ -1,5 +1,18 @@
 # Changelog
 
+## [16.0.0](https://github.com/theforeman/puppet-certs/tree/16.0.0) (2022-08-01)
+
+[Full Changelog](https://github.com/theforeman/puppet-certs/compare/15.1.1...16.0.0)
+
+**Breaking changes:**
+
+- Drop EL7 support [\#400](https://github.com/theforeman/puppet-certs/pull/400) ([ehelms](https://github.com/ehelms))
+- Refs [\#35005](https://projects.theforeman.org/issues/35005): Drop certs::pulp\_client [\#399](https://github.com/theforeman/puppet-certs/pull/399) ([ehelms](https://github.com/ehelms))
```
ehelms, line 10: This is the bigger item to include the drop of, mostly to prevent confusion for those who go digging when debugging.

ekohl: Thinking more about this, I don't think it's worth it. A lot of code is already conditional. It's also not that we clean up the files it generated (`/etc/pki/certs/pulp-client.crt` and `/etc/pki/private/pulp-client.key` I think), which can also confuse people. Considering the churn a major release generates, I'm choosing not to release this now.

evgeni:
If we merge https://github.com/theforeman/puppet-certs/pull/403, this would be worth a release? ;-)

ekohl:
> If we merge #403, this would be worth a release? ;-)

Don't we need stable branches for that anyway?

evgeni:
> Don't we need stable branches for that anyway?

I guess that too, yeah.

evgeni (approved):

PR #405, opened 2023-05-01
Mark compatible with puppetlabs/concat 8.x

ekohl:


ekohl:
It should be dropped instead: https://github.com/theforeman/puppet-certs/pull/406

PR #406, opened 2023-05-03
Remove unused puppetlabs-concat dependency

ekohl:
Fixes: 2cd43627b06cc1ec648f29d6de613d547e84411d ("Removed unused rhsm_reconfigure_script")


evgeni:
Should this also be removed from fixtures then? We don't strictly test the deps from metadata.json

evgeni (approved):

PR #409, opened 2023-06-08
Allow overriding passwords generated by extlib

coreone:


manifests/candlepin.pp
```diff
@@ -68,8 +70,14 @@
     build_dir     => $certs::ssl_build_dir,
   }
 
-  $keystore_password = extlib::cache_data('foreman_cache_data', $keystore_password_file, extlib::random_password(32))
-  $truststore_password = extlib::cache_data('foreman_cache_data', $truststore_password_file, extlib::random_password(32))
+  # Generate and cache the password on the master once
+  # In multi-puppetmaster setups, the user should specify their own
+  $final_keystore_password = pick_default(
```
ekohl, line 75: Any particular reason to use `pick_default` over `pick`?

manifests/init.pp
```diff
@@ -73,6 +75,7 @@
   Boolean $generate = true,
   Boolean $regenerate = false,
   Boolean $deploy = true,
+  String[1] $ca_key_password = $certs::params::ca_key_password,
```
ekohl, line 78: What's the reason for moving this parameter to another line?

I also like the data type, but should we enforce at least X characters? for NSS we use 10 and I think that's a good number.

manifests/params.pp
```diff
@@ -23,6 +23,11 @@
   $qpid_client_cert = "${pulp_pki_dir}/qpid/client.crt"
   $qpid_client_ca_cert = "${pulp_pki_dir}/qpid/ca.crt"
 
+  # Generate and cache the password on the master once
+  # In multi-puppetmaster setups, the user should specify their own
+  $ca_key_password = extlib::cache_data('foreman_cache_data', 'ca_key_password', extlib::random_password(24))
+  $nss_db_password = extlib::cache_data('foreman_cache_data', 'certs-nss-db-password', extlib::random_password(32))
```
ekohl, line 29: What's the reason to move the NSS DB pass to `params.pp`? Looks like we can continue using it in `certs::ssltools:nssdb`. You do need to set it via Hiera, but you don't introduce it as a top level parameter anyway.

ekohl (commented):
Apologies for missing this completely while I was on vacation, but I do like this change.

manifests/init.pp
coreone: This wasn't a parameter before. The other line in the diff was where it was defined inside the class. I also increased the string size to 10 per your suggestion.

manifests/params.pp
coreone: This was just following a pattern. Most of the other modules I looked at had this kind of thing done in `params.pp`, so I just did it there as well. I can move it out to the class itself if that makes more sense.

coreone:
@ekohl anything else needed on this?

manifests/candlepin.pp
```diff
@@ -68,8 +70,14 @@
     build_dir     => $certs::ssl_build_dir,
   }
 
-  $keystore_password = extlib::cache_data('foreman_cache_data', $keystore_password_file, extlib::random_password(32))
-  $truststore_password = extlib::cache_data('foreman_cache_data', $truststore_password_file, extlib::random_password(32))
+  # Generate and cache the password on the master once
+  # In multi-puppetmaster setups, the user should specify their own
+  $final_keystore_password = pick(
+    $keystore_password, extlib::cache_data('foreman_cache_data', $keystore_password_file, extlib::random_password(32))
```
ekohl, line 76: This may be a slight nuance, but in this case I do think the data is generated and cached, even if unused. Would it make sense to use a more traditional approach with `if/else` here?

manifests/init.pp
```diff
@@ -73,6 +75,7 @@
   Boolean $generate = true,
   Boolean $regenerate = false,
   Boolean $deploy = true,
+  String[10] $ca_key_password = $certs::params::ca_key_password,
```
ekohl, line 78: Can you keep it on the line where it used to be? So between `ca_cert_stripped` and `ca_key_password_file`?

manifests/params.pp
ekohl: Ideally we would get rid of `params.pp`. In our installer we use Kafo, which can't deal with function calls in parameters (default parsing breaks). Now I do see that `ca_key_password` already was doing that in `init.pp` so now I wonder if that was broken all along.

ekohl: Ah, I missed something in my review: it's in the body now, not the actual class parameter. That's why it's not a problem today.

manifests/init.pp
coreone: This is new because `ca_key_password` was not in the parameters for the class previously.

manifests/candlepin.pp
coreone: Changed these 2 lines into `if/else` blocks

manifests/init.pp
ekohl: I'm debating having a default or not. Having it means it ends up in our installer's answer files. Not sure if that's a benefit or a downside. @ehelms?

I also wonder what happens if you change the password parameter. I fear it would break since there's no support to change the password.

ehelms: I'm a bit confused. I do not like having it as a default and showing up in the answers file. This currently already has a default coming from params.pp?

ekohl: The diff in GH isn't clear (I also failed at that at first). Right now it's in the body of the class. This PR moves it to a class parameter (with indirection via params since Kafo can't deal with calling functions as defaults).

ekohl:
I did some testing: if you change the password for either the keystore or the truststore it fails to open it:

```
Error 1: Puppet Truststore_certificate resource '/etc/candlepin/certs/truststore:artemis-client' failed. Logs:
  /Stage[main]/Certs::Candlepin/Truststore_certificate[/etc/candlepin/certs/truststore:artemis-client]
    Adding autorequire relationship with File[/etc/pki/katello/truststore_password-file]
    Adding autorequire relationship with File[/etc/foreman/client_cert.pem]
    Adding autonotify relationship with File[/etc/candlepin/certs/truststore]
    Starting to evaluate the resource (581 of 1413)
    Evaluated in 0.46 seconds
  /Stage[main]/Certs::Candlepin/Truststore_certificate[/etc/candlepin/certs/truststore:artemis-client]/ensure
    change from 'absent' to 'present' failed: Execution of '/bin/keytool -import -v -noprompt -storetype pkcs12 -keystore /etc/candlepin/certs/truststore -alias artemis-client -file /etc/foreman/client_cert.pem -storepass:file /etc/pki/katello/truststore_password-file -J-Dcom.redhat.fips=false' returned 1: keytool error: java.io.IOException: keystore password was incorrect
java.io.IOException: keystore password was incorrect
	at sun.security.pkcs12.PKCS12KeyStore.engineLoad(PKCS12KeyStore.java:2089)
	at java.security.KeyStore.load(KeyStore.java:1445)
	at sun.security.tools.keytool.Main.doCommands(Main.java:839)
	at sun.security.tools.keytool.Main.run(Main.java:380)
	at sun.security.tools.keytool.Main.main(Main.java:373)
Caused by: java.security.UnrecoverableKeyException: failed to decrypt safe contents entry: javax.crypto.BadPaddingException: Given final block not properly padded. Such issues can arise if a bad key is used during decryption.
	... 5 more
Error 2: Puppet Truststore_certificate resource '/etc/candlepin/certs/truststore:candlepin-ca' failed. Logs:
  /Stage[main]/Certs::Candlepin/Truststore_certificate[/etc/candlepin/certs/truststore:candlepin-ca]
    Adding autorequire relationship with File[/etc/pki/katello/truststore_password-file]
    Adding autorequire relationship with File[/etc/candlepin/certs/candlepin-ca.crt]
    Adding autonotify relationship with File[/etc/candlepin/certs/truststore]
    Starting to evaluate the resource (602 of 1413)
    Evaluated in 0.36 seconds
  /Stage[main]/Certs::Candlepin/Truststore_certificate[/etc/candlepin/certs/truststore:candlepin-ca]/ensure
    change from 'absent' to 'present' failed: Execution of '/bin/keytool -import -v -noprompt -storetype pkcs12 -keystore /etc/candlepin/certs/truststore -alias candlepin-ca -file /etc/candlepin/certs/candlepin-ca.crt -storepass:file /etc/pki/katello/truststore_password-file -J-Dcom.redhat.fips=false' returned 1: keytool error: java.io.IOException: keystore password was incorrect
java.io.IOException: keystore password was incorrect
	at sun.security.pkcs12.PKCS12KeyStore.engineLoad(PKCS12KeyStore.java:2089)
	at java.security.KeyStore.load(KeyStore.java:1445)
	at sun.security.tools.keytool.Main.doCommands(Main.java:839)
	at sun.security.tools.keytool.Main.run(Main.java:380)
	at sun.security.tools.keytool.Main.main(Main.java:373)
Caused by: java.security.UnrecoverableKeyException: failed to decrypt safe contents entry: javax.crypto.BadPaddingException: Given final block not properly padded. Such issues can arise if a bad key is used during decryption.
	... 5 more
```

You can delete the file and it will be recreated, but rotating passwords is not supported. I'm not sure what's the best solution to prevent users from this footgun.

@ehelms any thoughts on changing the keystore & truststore types to catch the password error and recreate the file?

ehelms:
> @ehelms any thoughts on changing the keystore & truststore types to catch the password error and recreate the file?

Possible to add I should think -- if password change is detected we need to trigger a delete of the truststore and then let it re-build itself.

ekohl:
I've opened https://github.com/theforeman/puppet-certs/pull/428 but it doesn't work yet. I'll continue on it tomorrow.

coreone:
@ekohl circling back around to this after I see that the #428 has been merged.

manifests/init.pp
ekohl: @ehelms can we revisit this? I still think this is a good direction.

I'm still leaning to `undef` and also support `Sensitive`. So `Variant[Undef, Sensitive[String[10]], String[10]]] $ca_key_password = undef` (or `Optional[Variant[Sensitive[String[10]], String[10]]]` which is equivalent).

ehelms:
I think this needs at least two acceptance tests added:

1. Providing a password on fresh run
2. Changing the password for an existing setup

manifests/init.pp
ehelms: If I understand correctly, for an existing setup, if a new ca_key_password is provided this will force a new key to be generated, which forces a new CA to be generated which cascades and forces all smart-proxy / Capsules and clients to need to be updated. 

Will the code properly handle that? I think we need a test for it. How do we help users not break their installations accidentally?

ekohl: One data point that just came up today: if puppet-agent is removed then the cache is wiped. If the ssl-build directory isn't wiped, then there's a mismatch. If we have the value as an explicit parameter, the value is stored in the answer file and there wouldn't be an issue. So an argument against `undef` as default and in favor of the current code.

ekohl:
@coreone Any chance you could take a look at the acceptance tests @ehelms asked about?

PR #411, opened 2023-08-04
puppetlabs/stdlib: Require 9.x

bastelfreak:


ehelms (approved):

ekohl:
I was holding off on merging this because many modules require < 9, just like our installer: https://github.com/theforeman/foreman-installer/blob/0d2fb671ea579bf1f44b93dded089c0db159b536/Puppetfile#L3-L5

ehelms:
Does this break something? Do I need to revert? The tests were passing so I assumed it was good to go.

ekohl:
Yes, I should have left comments that it should be a concentrated effort. Perhaps it's a good excuse to move forward with it on all modules since it needs to happen anyway.

PR #413, opened 2023-09-20
Fixes #36755 - Set up the correct CA cert for foreman-proxy

ekohl:
The Foreman Proxy certificate is signed by the server_ca_cert, but it previously set up the non-matching default_ca_cert.


ehelms:
Mentioned on IRC:

Foreman talks to the smart-proxy using client certificates derived from the default CA, so the smart-proxy needs the CA for those clients to verify them

PR #414, opened 2023-09-20
Remove all qpid related certificate handling

ehelms:
When all references to Qpid are dropped, this can be merged. As Katello 4.10 should be handling that, we ought to be able to remove the code references in puppet-katello and puppet-foreman_proxy_content and then merge this.


ekohl:
With a downstream hat on, won't we  still need to ensure things are absent in 4.11?

evgeni (changes requested):
The README still refers to qpid certs, otherwise LGTM

evgeni (approved):

PR #416, opened 2023-09-22
Deploy the CA password file to ssl build directory

ehelms:
As the pki_dir is intended for deployment rather than generation, and the CA password file is needed primarily for certificate generation it should reside in the build directory.

### DIFF
--- a/manifests/ca.pp
+++ b/manifests/ca.pp
@@ -23,6 +23,10 @@ class certs::ca (
   String $ca_key_password = $certs::ca_key_password,
   Stdlib::Absolutepath $ca_key_password_file = $certs::ca_key_password_file,
 ) {
+  file { "${certs::pki_dir}/private/${default_ca_name}.pwd":
+    ensure => absent,
+  }
+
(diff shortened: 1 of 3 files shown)


ekohl (commented):
Should we ensure the old path is absent as well, just to clean things up?

ekohl:
Something appears to be very broken in the acceptance tests.

ehelms:
I think we should ensure we are managing the build directory first -- https://github.com/theforeman/puppet-certs/pull/419

manifests/ca.pp
```diff
@@ -23,6 +23,10 @@
   String $ca_key_password = $certs::ca_key_password,
   Stdlib::Absolutepath $ca_key_password_file = $certs::ca_key_password_file,
 ) {
+  file { "${certs::pki_dir}/private/${default_ca_name}.pwd":
+    ensure => absent,
+  }
```
ehelms, line 28: Added the cleanup

ekohl (approved):

PR #417, opened 2023-09-22
Create pki_dir structure only if deploy

ehelms:
Contains / requires -- https://github.com/theforeman/puppet-certs/pull/416

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -11,24 +11,26 @@ class certs::config (
     mode   => '0700',
   }
 
-  file { $pki_dir:
-    ensure => directory,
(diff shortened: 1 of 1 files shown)


ehelms:
@ekohl could you review?

ekohl:
It took me a while to get to this, but I think the design could be slightly different. The biggest issue that I see is that every concrete class (qpid, puppet, foreman, etc) can deploy. In that case `$certs::deploy` may be `false` while `$certs::foreman::deploy` may be `true`.

I've opened https://github.com/theforeman/puppet-certs/pull/436 but I'm not 100% sure it is correct either.

PR #420, opened 2023-09-23
Copy the server CA certificate with file resource

ehelms:
This simplifies the code required to make the server CA certificate which is either just a copy of the default CA or the supplied custom CA certificate from a user.

The current implementation requires the server CA certificate in the `build_dir` due to how we generate the tarball for Capsules.

Stay tuned for further enhancements, for now this is a small step simplification.

### DIFF
--- a/manifests/apache.pp
+++ b/manifests/apache.pp
@@ -54,7 +54,6 @@ class certs::apache (
   Stdlib::Absolutepath $pki_dir = $certs::pki_dir,
   Optional[Stdlib::Absolutepath] $server_cert = $certs::server_cert,
   Optional[Stdlib::Absolutepath] $server_key = $certs::server_key,
-  Optional[Stdlib::Absolutepath] $server_cert_req = $certs::server_cert_req,
   String[2,2] $country = $certs::country,
   String $state = $certs::state,
   String $city = $certs::city,
@@ -70,19 +69,25 @@ class certs::apache (
   # This variable is unused but considered public API
   $apache_ca_cert = $certs::katello_server_ca_cert
 
+  $apache_cert_path = "${certs::ssl_build_dir}/${hostname}/${apache_cert_name}"
+
   if $server_cert {
-    cert { $apache_cert_name:
-      ensure         => present,
-      hostname       => $hostname,
-      cname          => $cname,
-      generate       => $generate,
-      deploy         => false,
-      regenerate     => $regenerate,
-      custom_pubkey  => $server_cert,
-      custom_privkey => $server_key,
-      custom_req     => $server_cert_req,
-      build_dir      => $certs::ssl_build_dir,
+    file { "${apache_cert_path}.crt":
+      ensure => file,
+      source => $server_cert,
+      owner  => 'root',
+      group  => 'root',
+      mode   => '0440',
     }
+    file { "${apache_cert_path}.key":
+      ensure => file,
+      source => $server_key,
+      owner  => 'root',
+      group  => 'root',
+      mode   => '0440',
+    }
+
+    $require_cert = File["${apache_cert_path}.crt"]
   } else {
     cert { $apache_cert_name:
       ensure        => present,
@@ -101,6 +106,8 @@ class certs::apache (
       password_file => $ca_key_password_file,
       build_dir     => $certs::ssl_build_dir,
     }
+
+    $require_cert = Cert[$apache_cert_name]
   }
 
   if $deploy {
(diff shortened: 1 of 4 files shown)


manifests/apache.pp
```diff
@@ -23,20 +22,23 @@
...
-      hostname       => $hostname,
-      cname          => $cname,
-      generate       => $generate,
-      deploy         => false,
-      regenerate     => $regenerate,
-      custom_pubkey  => $server_cert,
-      custom_privkey => $server_key,
-      custom_req     => $server_cert_req,
-      build_dir      => $certs::ssl_build_dir,
+    file { "${apache_cert_path}.crt":
```
ekohl, line 29: This looks the wrong way around: it copies from `/etc/pki/certs/katello-apache.crt` to the build directory. But what's the point of that? Then later we have `certs::keypair` which deploys from the build directory to this location. Isn't that redundant?

ehelms: This creates `/root/ssl-build/example.com/example.com-apache.crt` using the custom certificate passed in by the user to the `server_cert` variable.

ekohl: Can we skip that step? Like enhancing the deploy step to do the copy directly.

ehelms: I contemplated that option. There are two considerations:

 1. The tar create function expects this to live in the build directory and I was trying not to break that functionality yet but rather re-factor and simplify the code.
 2. I've considered dropping this copying, and instead having the installer copy user input certificates to the build directory rather than the Puppet layer. This would attempt to solve the issue where users give certificates as input to the installer, and then by some process remove the custom certificates and things break. This would also allow backup / restore to be more self-contained I believe.

ehelms: @ekohl thoughts?

ehelms: Rebased -- @ekohl I am still waiting on your thoughts to my reply.

manifests/apache.pp
```diff
@@ -54,7 +54,6 @@
   Stdlib::Absolutepath $pki_dir = $certs::pki_dir,
   Optional[Stdlib::Absolutepath] $server_cert = $certs::server_cert,
   Optional[Stdlib::Absolutepath] $server_key = $certs::server_key,
-  Optional[Stdlib::Absolutepath] $server_cert_req = $certs::server_cert_req,
```
evgeni, line 57: Why was that ever needed?  (curious-Evgeni is afraid  of the answer)

manifests/apache.pp
```diff
@@ -100,6 +103,12 @@
       deploy        => false,
       password_file => $ca_key_password_file,
       build_dir     => $certs::ssl_build_dir,
+    } ->
+    file { "${apache_cert_path}.crt":
+      ensure => file,
+      owner  => 'root',
+      group  => 'root',
+      mode   => '0440',
```
evgeni, line 111: What is this for? To satisfy the `require    => File["${apache_cert_path}.crt"],` below? I don't think this will correctly notify when the cert changed.

manifests/apache.pp
```diff
@@ -114,7 +123,7 @@
       cert_owner => 'root',
       cert_group => $group,
       cert_mode  => '0440',
-      require    => Cert[$apache_cert_name],
+      require    => File["${apache_cert_path}.crt"],
```
evgeni, line 126: should this also subscribe to the `.key`? Granted, I can't really think of a case where the key changes but not the cert, but better safe than sorry while debugging weird cert issues in the future?

manifests/apache.pp
ehelms: I am not sure -- https://github.com/theforeman/puppet-certs/blob/39713ae873a7865d367ecbbc39a981e3c45d46a0/lib/puppet/provider/cert/katello_ssl_tool.rb#L20 seems symbolic

manifests/apache.pp
ehelms: Correct. This was put here to provide something to normalize the difference between when certificates are generated and when they are provided by the user. I felt like the deploy step needed to have some way to ensure that the thing actually existed. 

I could try and enhance https://github.com/theforeman/puppet-certs/blob/master/lib/puppet/type/cert.rb to auto create a file resource, I was trying to avoid too much investment in this provider given it's complexity. My goal being to reduce down the simplicity of how all the cert stuff works step by step.

evgeni: can't we do smth like

```puppet
if $server_cert {
$require_cert = File["${apache_cert_path}.crt"]
} else {
$require_cert = Cert[$apache_cert_name]
}
…
if $deploy {
    certs::keypair { $apache_cert_name:
      require => $require_cert
   }
}
```

ehelms: Good suggestion - added

ehelms:
I think this is simpler and more straight forward now.

evgeni (approved):
I thought I acked that one already, but seems not

PR #421, opened 2023-09-24
Drop RPMs from being included in tarball

ehelms:
These were kept in for N-1 support when we first switched away from using RPMs for deployment. Enough release have passed that we can drop this.


ehelms:
@ekohl could you take a look?

evgeni:
The RPMs are still built and everything, right? Just not distributed in the certs-tar?

(README needs an update, it says the deployment still uses RPM, but that's unrelated to the changes in this PR)

evgeni (approved):
pending my understanding of the change ("build but don't tar/distribute the rpms") is correct, ack

ehelms:
> The RPMs are still built and everything, right? Just not distributed in the certs-tar?

Correct -- https://github.com/theforeman/puppet-certs/pull/426 is designed to remove all RPM generation and simplify the code quite a bit.

PR #426, opened 2023-09-26
Remove deploy via provider

ehelms:
This includes commits that are in existing PRs, and is working towards this change with small incremental updates.

### DIFF
--- a/lib/puppet/provider/ca/katello_ssl_tool.rb
+++ b/lib/puppet/provider/ca/katello_ssl_tool.rb
@@ -6,55 +6,19 @@ Puppet::Type.type(:ca).provide(:katello_ssl_tool, :parent => Puppet::Provider::K
   protected
 
   def generate!
-    if existing_pubkey
-      FileUtils.mkdir_p(build_path)
-      FileUtils.cp(existing_pubkey, build_path(File.basename(pubkey)))
-      katello_ssl_tool('--gen-ca',
-                       '--dir', resource[:build_dir],
-                       '--ca-cert-dir', target_path('certs'),
-                       '--ca-cert', File.basename(pubkey),
-                       '--ca-cert-rpm', rpmfile_base_name,
-                       '--rpm-only')
-    else
-      katello_ssl_tool('--gen-ca',
-                       '--dir', resource[:build_dir],
-                       '-p', "file:#{resource[:password_file]}",
-                       '--force',
-                       '--ca-cert-dir', target_path('certs'),
-                       '--set-common-name', resource[:common_name],
-                       '--ca-cert', File.basename(pubkey),
-                       '--ca-key', File.basename(privkey),
-                       '--ca-cert-rpm', rpmfile_base_name,
-                       *common_args)
+    katello_ssl_tool(
+      '--gen-ca',
+      '--dir', resource[:build_dir],
+      '--password', "file:#{resource[:password_file]}",
+      '--force',
+      '--ca-cert-dir', resource[:build_dir],
+      '--set-common-name', resource[:common_name],
+      '--ca-cert', File.basename(pubkey),
+      '--ca-key', File.basename(privkey),
+      '--no-rpm',
+      *common_args
+    )
 
-    end
     super
   end
-
-  def existing_pubkey
-    if resource[:ca]
-      ca_details[:pubkey]
-    elsif resource[:custom_pubkey]
-      resource[:custom_pubkey]
-    end
-  end
-
-  def deploy!
-    if File.exist?(rpmfile)
-      # the rpm is available locally on the file system
-      rpm('-Uvh', '--force', rpmfile)
-    else
-      # we search the rpm in yum repo
-      yum("install", "-y", rpmfile_base_name)
-    end
-  end
-
-  def files_to_deploy
-    [pubkey]
-  end
-
-  def self.privkey(name)
-    build_path("#{name}.key")
-  end
-
 end
--- a/lib/puppet/provider/katello_ssl_tool.rb
+++ b/lib/puppet/provider/katello_ssl_tool.rb
@@ -5,43 +5,24 @@ module Puppet::Provider::KatelloSslTool
 
     initvars
 
-    commands :rpm => 'rpm'
-    commands :yum => 'yum'
     commands :katello_ssl_tool_command => 'katello-ssl-tool'
 
     def exists?
-      ! generate? && ! deploy?
+      !generate?
     end
 
     def create
       generate! if generate?
-      deploy!   if deploy?
     end
 
     def destroy
-      files_to_deploy.each do |file|
-        FileUtils.rm_f(file)
-      end
-
-      output = execute([:rpm, '-q', rpmfile_base_name], failonfail: false)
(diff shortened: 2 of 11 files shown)


manifests/apache.pp
```diff
@@ -68,20 +67,23 @@
   $apache_cert = "${pki_dir}/certs/katello-apache.crt"
   $apache_key  = "${pki_dir}/private/katello-apache.key"
   # This variable is unused but considered public API
-  $apache_ca_cert = $certs::katello_server_ca_cert
```
ekohl, line 71: We use this, which is why there was an explicit comment above it to keep it in.
https://github.com/theforeman/puppet-katello_devel/blob/596745918fa758f209dbffb81d673c6c011186c9/manifests/apache.pp#L9
https://github.com/theforeman/puppet-foreman_proxy_content/blob/1c787b10b0d2f3bb4dbe1a7b7095feab03786a07/manifests/reverse_proxy.pp#L57

ekohl (commented):
Not a complete review, but one thing that jumped out.

manifests/apache.pp
ehelms: I fixed it over here -- https://github.com/theforeman/puppet-certs/pull/420 which has that particular commit broken out.

ehelms:
This includes https://github.com/theforeman/puppet-certs/pull/420 which I would like to have in first.

ehelms:
Well, this is annoying, the acceptance tests come out differently on EL9 due to spacing:

```
      Failure/Error: its(:issuer) { should eq("C = US, ST = North Carolina, L = Raleigh, O = Katello, OU = SomeOrgUnit, CN = #{fqdn}") }
        
        expected: "C = US, ST = North Carolina, L = Raleigh, O = Katello, OU = SomeOrgUnit, CN = centos9-64.example.com"
             got: "C=US, ST=North Carolina, L=Raleigh, O=Katello, OU=SomeOrgUnit, CN=centos9-64.example.com"
```

evgeni:
I could swear we have this issue somewhere else too and somehow made it work.

Katello certs tools?

ehelms:
> I could swear we have this issue somewhere else too and somehow made it work.
> 
> Katello certs tools?

I couldn't find anything. One option is to update serverspec project to use a standardized output when printing the subject or issuer with `-nameopt oneline`. Means we will have to wait and hope for a release.

Another option is to create our own matcher instead of `eq` to match and not care about the output style and rather care about the content.

Third option is to use command and have contorl over it:

```
describe command('openssl x509 -noout -text -in <cert> -nameopt oneline') do
  its(:stdout) {}
end
```

A fourth option is to carry a patch here in this repo to these two functions https://github.com/mizzy/serverspec/blob/master/lib/serverspec/type/x509_certificate.rb#L9-L15

evgeni:
Would `its(:issuer).strip(' ').to eq("blah")` work?

ehelms:
> Would `its(:issuer).gsub(' ', '').to eq("blah")` work?

This is the winner https://github.com/theforeman/puppet-certs/blob/master/spec/support/acceptance/matchers/match_without_whitespace.rb

ehelms:
This is rebased and tests updated. This does two things effectively:

- Removes the ability to deploy certificates through the provider that generates them
- Removes creation of the RPMs from katello-certs-tool that are no longer used for any part of the process

I think further re-factoring can come after this. My goal is to first remove the RPM aspect as cleanly as I can.

lib/puppet/provider/katello_ssl_tool.rb
```diff
@@ -5,43 +5,25 @@ class Cert < Puppet::Provider
 
     initvars
 
-    commands :rpm => 'rpm'
     commands :yum => 'yum'
```
evgeni, line 8:
```suggestion
```

it's also unused now

lib/puppet/provider/ca/katello_ssl_tool.rb
```diff
@@ -6,55 +6,19 @@
...
-                       '--ca-cert-dir', target_path('certs'),
-                       '--set-common-name', resource[:common_name],
-                       '--ca-cert', File.basename(pubkey),
-                       '--ca-key', File.basename(privkey),
-                       '--ca-cert-rpm', rpmfile_base_name,
-                       *common_args)
+    katello_ssl_tool(
+      '--gen-ca',
+      '--dir', resource[:build_dir],
+      '-p', "file:#{resource[:password_file]}",
```
evgeni, line 12: totally unrelated, but can this become
```suggestion
      '--password', "file:#{resource[:password_file]}",
```

evgeni: (I think there are several places that could use that treatment, feel free to tell me to do it myself in a followup)

lib/puppet/provider/katello_ssl_tool.rb
```diff
@@ -114,41 +59,20 @@ def common_args
...
-    end
-
-    def self.target_path(file_name = '')
-      File.join("/etc/pki/katello-certs-tools", file_name)
+    def privkey(key_name = resource[:name])
+      if resource.to_hash.key?(:hostname)
+        "#{resource[:build_dir]}/#{resource[:hostname]}/#{key_name}.key"
+      else
+        "#{resource[:build_dir]}/#{key_name}.key"
+      end
```
evgeni, line 75: shouldn't this be using `build_path` (which, ironically, doesn't use `:build_dir`)?

lib/puppet/provider/katello_ssl_tool.rb
```diff
@@ -114,41 +59,20 @@ def common_args
...
-      path = path + "-[0-9].*" + "noarch.rpm"
-
-      rpmfile = Dir[path].max_by do |file|
-        version_from_name(file)
+    def pubkey(cert_name = resource[:name])
+      if resource.to_hash.key?(:hostname)
+        "#{resource[:build_dir]}/#{resource[:hostname]}/#{cert_name}.crt"
+      else
+        "#{resource[:build_dir]}/#{cert_name}.crt"
       end
```
evgeni, line 67: shouldn't this be using `build_path` (which, ironically, doesn't use `:build_dir`)?

evgeni: because then you don't need the `if hostname` branching, as `lib/puppet/provider/cert/katello_ssl_tool.rb` already does that:
https://github.com/theforeman/puppet-certs/blob/d7e8d4e981036ef5d237f708a5a5b48d69edc470/lib/puppet/provider/cert/katello_ssl_tool.rb#L53-L55

ehelms: Cleaned this whole area up to be more succinct.

lib/puppet/provider/ca/katello_ssl_tool.rb
ehelms: Addressed all the spots I could find.

ehelms:
> Puppet providers are really not my domain, but I posted a few comments ;-)

Appreciated - they helped me clean the code up.

evgeni (approved):

PR #428, opened 2023-09-28
Support changing passwords on keystores & truststores

ekohl:
Opening the store when checking if it exists will raise an exception if the password is incorrect. This takes an approach of catching all exceptions and treating them all the same.

Right now it doesn't work yet and I'm getting integrity errors so I suspect it may not recreate the file. Need to do some further digging.

### DIFF
--- a/lib/puppet_x/certs/provider/keystore.rb
+++ b/lib/puppet_x/certs/provider/keystore.rb
@@ -11,7 +11,24 @@ module PuppetX
         end
 
         def exists?
-          File.exist?(store)
+          return false unless File.exist?(store)
+
+          begin
+            keytool(
+              '-list',
+              '-keystore', store,
+              '-storepass:file', resource[:password_file],
+            )
+          rescue Puppet::ExecutionFailure => e
+            if e.message.include?('java.security.UnrecoverableKeyException')
+              Puppet.debug("Invalid password for #{store}")
+              return false
+            else
+              Puppet.log_exception(e, "Failed to read keystore '#{store}'")
+            end
+          end
+
+          true
         end
 
         def store
@@ -24,6 +41,8 @@ module PuppetX
 
         def generate_keystore
           temp_alias = 'temporary-entry'
+
+          FileUtils.rm_f(store)
 
           begin
             keytool(
--- a/spec/acceptance/keystore_spec.rb
+++ b/spec/acceptance/keystore_spec.rb
@@ -38,10 +38,120 @@ describe 'certs' do
       it { should be_grouped_into 'root' }
     end
 
-    describe command("keytool -list -keystore /etc/pki/keystore -storepass:file /etc/pki/keystore_password-file") do
+    describe command("keytool -list -keystore /etc/pki/keystore -storepass testpassword") do
       its(:exit_status) { should eq 0 }
       its(:stdout) { should match(/^Keystore type: PKCS12$/i) }
       its(:stdout) { should match(/^Your keystore contains 0 entries$/) }
+    end
+
+    describe 'changing password' do
+      describe 'apply puppet' do
+        let(:manifest) do
+          <<-PUPPET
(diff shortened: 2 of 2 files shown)


ehelms:
This should do the trick:

```
diff --git a/lib/puppet_x/certs/provider/keystore.rb b/lib/puppet_x/certs/provider/keystore.rb
index dc0109b..f2a4ed4 100644
--- a/lib/puppet_x/certs/provider/keystore.rb
+++ b/lib/puppet_x/certs/provider/keystore.rb
@@ -22,10 +22,11 @@ module PuppetX
           rescue StandardError
             # If listing fails for any reason, pretend it doesn't exist so it
             # gets recreated
-            false
+            delete_keystore
+            return false
           end
 
-          true
+          return true
         end
 
         def store
@@ -69,7 +70,7 @@ module PuppetX
         end
 
         def delete_keystore
-          File.rm(store)
+          File.delete(store)
         end
       end
     end
```

ehelms:
Noop output with a changed password:

```
[root@centos8-64 /]# puppet apply --verbose --detailed-exitcodes /tmp/apply_manifest_152742494.pp.WnvqGY --noop
Info: Loading facts
Info: Loading facts
Info: Loading facts
Info: Loading facts
Notice: Compiled catalog for centos8-64.example.com in environment production in 0.03 seconds
Info: Using environment 'production'
Info: Applying configuration version '1696361357'
Notice: /Stage[main]/Main/File[/etc/pki/keystore_password-file]/content: current_value '{md5}455d12ba5891711c2bb67f7ea3fcc315', should be '{md5}89f1cac5f41aa8f36091994d91cdb817' (noop)
Notice: Class[Main]: Would have triggered 'refresh' from 1 event
Notice: Stage[main]: Would have triggered 'refresh' from 1 event
Notice: Applied catalog in 0.25 seconds
```

lib/puppet_x/certs/provider/keystore.rb
```diff
@@ -11,7 +11,21 @@ def destroy
...
+          begin
+            keytool(
+              '-list',
+              '-keystore', store,
+              '-storepass:file', resource[:password_file],
+            )
+          rescue StandardError
+            # If listing fails for any reason, pretend it doesn't exist so it
+            # gets recreated
+            false
```
ekohl, line 25: Thanks for pointing out this doesn't return.

lib/puppet_x/certs/provider/keystore.rb
```diff
@@ -11,7 +11,25 @@ def destroy
...
+          begin
+            keytool(
+              '-list',
+              '-keystore', store,
+              '-storepass:file', resource[:password_file],
+            )
+          rescue Puppet::ExecutionFailure => e
+            if e.message.include?('java.security.UnrecoverableKeyException')
+              Puppet.debug("Invalid password for #{store}")
+              #delete_keystore
```
ekohl, line 25: Turns out this isn't needed, but it would break noop mode where it would make changes. I've chosen to add removal to `generate_keystore`.

lib/puppet_x/certs/provider/keystore.rb
```diff
@@ -25,6 +42,8 @@ def type
         def generate_keystore
           temp_alias = 'temporary-entry'
 
+          FileUtils.rm_f(store)
```
ehelms, line 45: Does `rm_f` handle file not found?

ehelms (approved):

lib/puppet_x/certs/provider/keystore.rb
ekohl: Yes. https://ruby-doc.org/stdlib-2.4.1/libdoc/fileutils/rdoc/FileUtils.html#method-i-safe_unlink states it's equal to `FileUtils.rm(file, force: true)` which ignores those errors.

PR #429, opened 2023-10-04
Bump major version to reflect next release

ekohl:
There have been breaking changes and this reflects that. It allows creating a stable branch without immediately releasing all changes.


ehelms (approved):
You'll need to update puppet-katello and puppet-foreman_proxy_content to avoid installer failures

PR #432, opened 2023-10-05
Handle more unknown password errors

ekohl:
When using keytool on a truststore the error is different than on a keystore.

Fixes: 6fea0bbb4143ca439cff01bf9f0e54cf88140d10 ("Support changing passwords on keystores & truststores")

### DIFF
--- a/lib/puppet_x/certs/provider/keystore.rb
+++ b/lib/puppet_x/certs/provider/keystore.rb
@@ -20,7 +20,7 @@ module PuppetX
               '-storepass:file', resource[:password_file],
             )
           rescue Puppet::ExecutionFailure => e
-            if e.message.include?('java.security.UnrecoverableKeyException')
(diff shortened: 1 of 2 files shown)


lib/puppet_x/certs/provider/keystore.rb
```diff
@@ -20,7 +20,7 @@ def exists?
               '-storepass:file', resource[:password_file],
             )
           rescue Puppet::ExecutionFailure => e
-            if e.message.include?('java.security.UnrecoverableKeyException')
+            if e.message.include?('java.security.UnrecoverableKeyException') || e.message.include?('keystore password was incorrect')
```
ekohl, line 23: I debated treating any error as "it doesn't exist" and still wonder if we should take that route.

ehelms (approved):

PR #436, opened 2023-11-07
Create pki_dir structure only when deployed

ekohl:
This is an alternative to https://github.com/theforeman/puppet-certs/pull/417. The goal is to only create the `/etc/pki` directory on deployments, because on then it's needed.


manifests/keypair.pp
```diff
@@ -15,6 +15,8 @@
   Boolean $key_decrypt = false,
   Optional[Stdlib::Absolutepath] $key_password_file = undef,
 ) {
+  include certs::config
```
ehelms, line 18: This inclusion breaks all of the tests, I do not think it is needed.

PR #444, opened 2024-04-25
Fixes #37384 - properly pass fips=false when checking keystore

evgeni:
In a FIPS-enabled environment, calling `keytool -list` with a wrong password doesn't yield an error, unless we also pass `fips=false` like we do when creating the keystore:

    # keytool -list -keystore ./store -storepass wrong-password
    Keystore type: PKCS11
    Keystore provider: SunPKCS11-NSS-FIPS

    Your keystore contains 0 entries

Passing `fips=false` makes it correctly raise the expected exception:

    # keytool -list -keystore ./store -storepass wrong-password -J-Dcom.redhat.fips=false
    keytool error: java.io.IOException: keystore password was incorrect

Fixes: 6fea0bbb4143ca439cff01bf9f0e54cf88140d10

### DIFF
--- a/lib/puppet_x/certs/provider/keystore.rb
+++ b/lib/puppet_x/certs/provider/keystore.rb
@@ -18,6 +18,7 @@ module PuppetX
               '-list',
               '-keystore', store,
               '-storepass:file', resource[:password_file],
+              '-J-Dcom.redhat.fips=false',
             )
           rescue Puppet::ExecutionFailure => e
             if e.message.include?('java.security.UnrecoverableKeyException') || e.message.include?('keystore password was incorrect')


ekohl (approved):
Feels like we lack some testing with FIPS. Out of interest, on FIPS with the correct password without the `-D` option, does that pass?

evgeni:
> Feels like we lack some testing with FIPS. Out of interest, on FIPS with the correct password without the `-D` option, does that pass?

FSVO "pass"? :)

```console
# keytool -list -keystore /etc/candlepin/certs/truststore -storepass:file /etc/pki/katello/truststore_password-file
Keystore type: PKCS11
Keystore provider: SunPKCS11-NSS-FIPS

Your keystore contains 0 entries

# keytool -list -keystore /etc/candlepin/certs/truststore -storepass:file /etc/pki/katello/truststore_password-file -J-Dcom.redhat.fips=false
Keystore type: PKCS12
Keystore provider: SUN

Your keystore contains 2 entries

artemis-client, Apr 25, 2024, trustedCertEntry, 
Certificate fingerprint (SHA-256): EA:C0:F7:55:9A:F5:0C:33:0D:33:53:10:18:FF:95:FA:2B:B7:5F:AD:BE:8E:3B:66:E9:60:1A:B4:8D:80:65:80
candlepin-ca, Apr 25, 2024, trustedCertEntry, 
Certificate fingerprint (SHA-256): 32:D2:5D:17:ED:15:97:CF:CF:95:FA:63:9B:CC:85:3F:0D:FB:55:76:9F:E8:9C:89:28:AA:CC:AA:48:48:24:8F
```

PR #449, opened 2024-04-30
Add class to configure generating certificates

ehelms:
The intent of this class is to allow generating any number of the sets of certificates from this module in a simple way. This is also a way to work around the fact that parsing params by Kafo doesn't cleanly work on classes such as `certs::apache` (see https://github.com/theforeman/puppet-certs/pull/424).


ehelms:
See https://github.com/theforeman/foreman-installer/pull/935 for example of how this is intended to be used.

manifests/generate.pp
```diff
@@ -0,0 +1,41 @@
...
+# $puppet::   Generates certificates needed by Puppet
+#
+class certs::generate (
+  Boolean $apache = false,
+  Boolean $foreman = false,
+  Boolean $candlepin = false,
+  Boolean $foreman_proxy = false,
+  Boolean $puppet = false,
+) {
+  if $certs::generate::apache {
```
ekohl, line 22: You can simplify all of these since the parameters are local
```suggestion
  if $apache {
```

spec/classes/certs_generate_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+      end
+
+      describe 'with default parameters' do
+        it { should compile.with_all_deps }
+      end
+
+      describe 'with apache true' do
+        let :pre_condition do
+          "class {'certs::generate': apache => true,}"
+        end
```
ekohl, line 17: Since the subject is the class itself, you can use `params` instead of a precondition
```suggestion
        let :params do
          { apache: true }
        end
```

spec/classes/certs_generate_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+
+        it { should compile.with_all_deps }
```
ekohl, line 19: The `should` syntax is deprecated and we should (pun not intended) migrate to `is_expected.to`
```suggestion
        it { is_expected.to compile.with_all_deps }
```

ekohl (commented):
I didn't see this before, but I recall a long time ago that I already discussed something like this with @timogoebel. Then we talked about making it a `define` so you could easily declare it for multiple hosts.

In effect this is a more configurable version of [certs::foreman_proxy_content](https://github.com/theforeman/puppet-certs/blob/master/manifests/foreman_proxy_content.pp).

We could even go a step further and generalize it to make `certs_tar` an optional parameter.

ehelms:
@evgeni What do you think about merging this in to then proceed with the certs command in the installer?

evgeni:
> @evgeni What do you think about merging this in to then proceed with the certs command in the installer?

yes please

evgeni (approved):

PR #453, opened 2024-05-21
Use puppet-openssl and remove use of katello-certs-tool

ehelms:


manifests/apache.pp
```diff
@@ -59,7 +59,7 @@
   String $city = $certs::city,
   String $org = $certs::org,
   String $org_unit = $certs::org_unit,
-  String $expiration = $certs::expiration,
+  Variant[Integer, String] $expiration = $certs::expiration,
```
ekohl, line 62: I'd be open to a PR that already changes it from Strings to Integers prior to merging this.

ekohl (commented):
Interesting idea. Do we today have any checks in katello-certs-tools that we would need to port over?

ehelms:
> Interesting idea. Do we today have any checks in katello-certs-tools that we would need to port over?

Checks? None that I am aware of - do you have ideas of what you are thinking?

The riskiest part of this change I believe is to ensure that for existing installations the CA certificate / key / password are not regenerated. I've added a new migration test that uses a copy of what a standard looking katello-certs-tool based generated certificate looks like to check that it does not change.

ehelms:
I've opened a PR for the puppet-openssl support I am relying on here. (https://github.com/voxpupuli/puppet-openssl/pull/188)

PR #457, opened 2024-08-06
Move keystore and trustore password files

ehelms:
Change to storing the password files in `/etc/candlepin/certs` co-located with the trustore and keystore themselves rather than storing them arbitrarily in their current location. This helps bring the Candlepin SSL configuration together.


ekohl (approved):
Is this something you'd want in Foreman 3.12?

ehelms:
> Is this something you'd want in Foreman 3.12?

More of a nice to have then a must have - so if it makes releasing more complex we can wait.

PR #458, opened 2024-08-06
Make deploying client bundle certificate optional

ehelms:
Currently we deploy this certificate in all deployments but it should only be deployed onto a foreman-proxy where reverse proxy is present. It is useless and unused on a Foreman deployment. This keeps the default behaviour by defaulting to 'present' but provides a way to set this to false in `puppet-foreman_proxy_content` (PR to follow).

### DIFF
--- a/manifests/foreman_proxy.pp
+++ b/manifests/foreman_proxy.pp
@@ -32,6 +32,7 @@ class certs::foreman_proxy (
   String $owner = 'root',
   Stdlib::Filemode $private_key_mode = '0440',
   Stdlib::Filemode $public_key_mode = '0444',
+  Enum['present', 'absent'] $client_bundle_ensure = 'present',
 ) inherits certs {
   $proxy_cert_name = "${hostname}-foreman-proxy"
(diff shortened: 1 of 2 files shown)


ehelms:
The puppet-foreman_proxy_content module includes `certs::foreman_proxy` in multiple locations (https://github.com/search?q=repo%3Atheforeman%2Fpuppet-foreman_proxy_content+certs%3A%3Aforeman_proxy&type=code) which from my early testing makes it difficult to include this change. As I need to switch to using `class { 'certs::foreman_proxy:}` in order to set this new parameter somewhere and I encounter duplicate class declarations.

ehelms:
> Why do we need that?

I believe this is used to perform re-encryption because SSL was being terminated in Apache on the Capsule and we then need to proxy the communication while preserving the SSL client certificate header (https://github.com/theforeman/puppet-foreman_proxy_content/blob/master/manifests/reverse_proxy.pp#L62).

ekohl:
My expectation is that it would simply result in a different SSL client certificate header. The one belongs to the certificate & private key provided in the bundle.

Issue #456, opened 2024-07-07
Custom CA overwritten when certs::server_ca_cert is undefined


selfsealingstembolts:
When upgrading a smart proxy with a custom certificate bundle from Foreman 3.10, which uses puppet-certs v17.1.1, to Foreman 3.11, which uses puppet-certs v18.0.0, `foreman-installer` will fail with TLS verification errors.

The root cause is this module overwriting the custom CA in `/root/ssl-build/katello-server-ca.crt` with the  contents of `/root/ssl-build/katello-default-ca.crt` when `certs::server_ca_cert` is undefined in the foreman-installer scenario answers file.

This behavior can be seen [here](https://github.com/theforeman/puppet-certs/blob/8e24284b31a530bad855145a1e725ba55093cc83/manifests/ca.pp#L52-L68). The commit introducing this behavior is 433dadc.

A workaround is to set the following values in `/etc/foreman-installer/scenarios.d/foreman-proxy-content-answers.yaml`
```
certs:
  server_cert: "/root/ssl-build/${smart-proxy.example.com}/${smart-proxy.example.com}-foreman-proxy.crt"
  server_key: "/root/ssl-build/${smart-proxy.example.com}/${smart-proxy.example.com}-foreman-proxy.key"
  server_ca_cert: "/root/ssl-build/katello-server-ca.crt"
```
The values of certs::server_cert and certs:server_key are required along with certs:server_ca_cert otherwise the installer will fail on the katello-certs-check step.

I am not sure what an appropriate fix would look like. It does seem desirable to use the default CA if it is not explicitly defined. The obvious candidate is testing if `/root/ssl-build/katello-server-ca.crt` already exists before overwriting it with the default CA. Alternatively, applying a migration to explicitly define the values mentioned above. Migrations appear to be outside the scope of this repository however.

knoppi:
An obvious solution would be to add `replace => false` to the resource `File[$server_ca_path]` in `manifests/ca.pp`. Then it would not overwrite the existing file in `/root/ssl_build`. This works as I just tested.
And I could probably still enforce it by adding `--certs-update-server-ca` to foreman installer.

ekohl:
It's not really clear what the problem is. To me this reads as a user error. I don't see the use case where you want custom certificates but not define a server CA. Perhaps we don't enforce it today, but I think it was always assumed you pass in all 3 values.

[...]

PR #461, opened 2024-09-12
Add a test for customer server certificates in tar file

ekohl:
This asserts that the default CA and server CA are the same in one scenario and differ in the other.


ekohl:
This is growing to take a stab at fixing https://github.com/theforeman/puppet-certs/issues/456 but I'm not sure it's the right track. I suspect that all the various classes (apache, etc) will still attempt to generate a certificate.

ekohl:
@ehelms this is now green, but I'm a bit hesitant on touching such a core class.

I wonder if passing `$certs::tar_file` should imply `generate => false` and `regenerate => false`.

ekohl:
This feels like a big regression and deserves a Redmine issue. It's one of our core workflows that we broke.

ehelms:
See https://github.com/theforeman/puppet-certs/pull/463 for a more targeted approach that relies on the way this worked prior to the re-design. I have also included a Redmine issue with it.

manifests/ca.pp
```diff
@@ -21,60 +21,53 @@
...
     }
-  }
 
-  if $generate {
-    file { "${certs::ssl_build_dir}/KATELLO-TRUSTED-SSL-CERT":
-      ensure  => link,
-      target  => $server_ca_path,
-      require => File[$server_ca_path],
+    if $generate {
+      file { "${certs::ssl_build_dir}/KATELLO-TRUSTED-SSL-CERT":
```
ekohl, line 59: Random note: I couldn't find a use for this anymore. I think we can drop it in a separate backwards-incompatible PR.

manifests/ca.pp
ekohl, line 1: Should the header be?
```puppet
  Boolean $generate = $certs::generate and !$certs::tar_file,
```

Though I think it should be part of `init.pp` and reused in other places. Ideally we'd get rid of `generate` and `regenerate` altogether.

manifests/ca.pp
ehelms: Yes - to my knowledge as well this can be dropped.

ekohl: https://github.com/theforeman/puppet-certs/pull/464

ekohl:
I split off the refactors without impact in https://github.com/theforeman/puppet-certs/pull/467.

PR #463, opened 2024-09-13
Fixes #37817: Only copy server CA in build root if generate is true

ehelms:
Alternative to https://github.com/theforeman/puppet-certs/pull/461. This does include the tests from 461, but not the re-factorings. I wanted to have as crisp of a solution to the regression as possible. The re-factoring, which I like, we can layer on top of the fix afterward.

If you rewind to https://github.com/theforeman/puppet-certs/commit/433dadc5ec41c2477fc6a04e056ca061fd818980, prior to this change, the `ca` resource was used to perform the copying to the server CA in the build root. This resource had the `generate` parameter built into it. This is what prevented the current regression from happening in the old design.  When deploying a foreman-proxy-content scenario in the installer, we are supplying all the certificates in the tarball. Therefore no generation needs to occur. Which we can see as the case by looking at the answers file (https://github.com/theforeman/foreman-installer/blob/develop/config/foreman-proxy-content-answers.yaml#L13C1-L14C1):

```
certs:
  generate: false
```

I think this is the correct solution at this point in time, as it restores the prior behavior and fixes the issues (as evidenced by the reproducer tests -- thanks @ekohl). 

This issue has given me some ideas on how this could be improved in upcoming releases through some re-factoring and re-design.

### DIFF
--- a/spec/acceptance/certs_spec.rb
+++ b/spec/acceptance/certs_spec.rb
@@ -151,4 +151,115 @@ describe 'certs' do
       its(:keylength) { should be >= 2048 }
     end
   end
+
+  context 'with tar file' do
+    before(:context) do
+      ['crt', 'key'].each do |ext|
+        source_path = "fixtures/example.partial.solutions.#{ext}"
+        dest_path = "/server.#{ext}"
+        scp_to(hosts, source_path, dest_path)
+      end
+    end
+
+    context 'with default ca' do
+      before(:context) do
+        manifest = <<~PUPPET
+          class { 'certs':
+            generate => true,
+            deploy   => false,
+          }
+
+          class { 'certs::foreman_proxy_content':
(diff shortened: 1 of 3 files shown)


spec/acceptance/certs_spec.rb
```diff
@@ -151,4 +151,115 @@ class { 'certs':
...
+
+        apply_manifest(manifest, catch_failures: true)
+
+        on default, 'rm -rf /root/ssl-build'
+      end
+
+      describe 'deploy certificates' do
+        manifest = <<-PUPPET
+          class { 'certs':
+            generate => false,
```
ekohl, line 232: Note to self: this bit I missed in my PR.

ekohl (approved):

ekohl:
I fixed up my commit message (customer -> custom) and added a `Fixes` trailer to yours.

PR #470, opened 2024-11-12
Fixes #38010 - Include keyalg in keytool for OpenJDK 17

ehelms:


lib/puppet/provider/keystore_certificate/openssl.rb
```diff
@@ -53,7 +53,8 @@ def add_certificate
         '-deststorepass:file', resource[:password_file],
         '-srcalias', resource[:alias],
         '-destalias', resource[:alias],
-        '-J-Dcom.redhat.fips=false'
+        '-J-Dcom.redhat.fips=false',
+        '-keyalg', 'RSA'
```
evgeni, line 57: `-importkeystore` doesn't have `-keyalg`

evgeni (changes requested):

lib/puppet/provider/keystore_certificate/openssl.rb
ehelms: Fair, it did not error out but it is not needed.

evgeni (approved):

PR #474, opened 2025-01-14
Add class to handle creation of certs for advisor service on localhost

ehelms:


manifests/iop_advisor_engine.pp
```diff
@@ -0,0 +1,64 @@
+# Contains certs specific configurations for advisor
+class certs::iop_advisor_engine (
+  Stdlib::Fqdn $hostname = 'localhost',
+  Boolean $generate = $certs::generate,
+  Boolean $regenerate = $certs::regenerate,
+  Boolean $deploy = $certs::deploy,
+  Stdlib::Absolutepath $ca_cert = $certs::candlepin_ca_cert,
+  Stdlib::Absolutepath $ca_key = $certs::candlepin_ca_key,
+  Stdlib::Absolutepath $pki_dir = $certs::pki_dir,
```
evgeni, line 9: these seem unused?

manifests/iop_advisor_engine.pp
```diff
@@ -0,0 +1,64 @@
+# Contains certs specific configurations for advisor
+class certs::iop_advisor_engine (
+  Stdlib::Fqdn $hostname = 'localhost',
```
ekohl, line 3: I don't like certificates for localhost. I know we have it for candlepin, but is there a way we can avoid this?

manifests/iop_advisor_engine.pp
```diff
@@ -0,0 +1,64 @@
...
+  Boolean $generate = $certs::generate,
+  Boolean $regenerate = $certs::regenerate,
+  Boolean $deploy = $certs::deploy,
+  Stdlib::Absolutepath $ca_cert = $certs::candlepin_ca_cert,
+  Stdlib::Absolutepath $ca_key = $certs::candlepin_ca_key,
```
ekohl, line 8: Copy-paste from candlepin or is this intentional?

manifests/iop_advisor_engine.pp
evgeni: I think it's conceptually the same problem as with candlepin: we run the service on localhost only so it's not reachable via the network (without us having to manage a firewall), but that means that the cert if it should validate be for localhost.

evgeni: I guess the only reasonable no-SSL answer are unix sockets, but I am not sure I want to dive into that direction with containers just yet

ehelms:
Ignore AlmaLinux failures, dropping EL8 here -- https://github.com/theforeman/puppet-certs/pull/475

manifests/iop_advisor_engine.pp
ehelms: We opted for this route for consistency with our Candlepin deployment model. And to give the option to move off localhost if we ever needed to. I am wondering if we should stick with consistency here, and aim for changing how we handle localhost services in our container deployment.

manifests/iop_advisor_engine.pp
```diff
@@ -0,0 +1,69 @@
...
+  $client_cert = $certs::foreman_proxy::foreman_ssl_cert
+  $client_key = $certs::foreman_proxy::foreman_ssl_key
+  $client_ca_cert = $certs::foreman_proxy::foreman_ssl_ca_cert
+
+  file { $cert_directory:
+    ensure => directory,
+    mode   => '0755',
+    owner  => $owner,
+    group  => $group,
+  }
```
evgeni, line 37: Should this be inside `if $deploy`?

manifests/iop_advisor_engine.pp
evgeni: @ekohl would you be OK with this as-is for now, so we have it at least working, and then re-consider when coming closer to a real release?

manifests/iop_advisor_engine.pp
ehelms: Yes - fixed.

manifests/iop_advisor_engine.pp
```diff
@@ -0,0 +1,69 @@
...
+    expiration    => $expiration,
+    ca            => $certs::default_ca,
+    generate      => $generate,
+    regenerate    => $regenerate,
+    password_file => $ca_key_password_file,
+    build_dir     => $certs::ssl_build_dir,
+  }
+
+  if $deploy {
+    $cert_directory = '/etc/iop-advisor-engine'
```
evgeni, line 47: line 22/23 use that, so I guess *this* should be  outside the if?

ehelms: Does it?

idk-look-harder

spec/acceptance/iop_advisor_engine_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+      }
+
+      group { 'foreman-proxy':
+        ensure => present,
+        system => true,
+      }
+
+      file { '/etc/iop-advisor-engine':
+        ensure => directory,
+      }
```
evgeni, line 22:
```suggestion
```

We create that dir now

evgeni (approved):

PR #475, opened 2025-01-14
Drop EL8 support -- Foreman/Katello do not support EL8 anymore

ehelms:


spec/spec_helper_acceptance.rb
```diff
@@ -1,6 +1,6 @@
 require 'voxpupuli/acceptance/spec_helper_acceptance'
 
-ENV['BEAKER_setfile'] ||= 'centos8-64{hostname=centos8-64.example.com}'
+ENV['BEAKER_setfile'] ||= 'centos9-64{hostname=centos9-64.example.com}'
```
evgeni, line 3: won't this be overwritten by modulesync anyway?

ehelms: Touche.

evgeni (approved):

PR #477, opened 2025-02-14
Regenerate keystore if the keystore is empty

ehelms:
If a keystore is attempted to be generated without the `-keyalg` (see https://github.com/theforeman/puppet-certs/commit/dc0f12e0e2737953780f40b857fe5920e18feb5c) then it will create an empty keystore file. This happens in older Katello releases that do not have this change backported. This adjusts the keystore handling to regenerate the keystore if it is found to be empty.

### DIFF
--- a/lib/puppet_x/certs/provider/keystore.rb
+++ b/lib/puppet_x/certs/provider/keystore.rb
@@ -24,6 +24,9 @@ module PuppetX
             if e.message.include?('java.security.UnrecoverableKeyException') || e.message.include?('keystore password was incorrect')
               Puppet.debug("Invalid password for #{store}")
               return false
+            elsif e.message.include?('Keystore file exists, but is empty')
+              Puppet.debug(e)
+              return false
             else
               Puppet.log_exception(e, "Failed to read keystore '#{store}'")
             end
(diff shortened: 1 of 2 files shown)


lib/puppet_x/certs/provider/keystore.rb
```diff
@@ -24,6 +24,9 @@ def exists?
             if e.message.include?('java.security.UnrecoverableKeyException') || e.message.include?('keystore password was incorrect')
               Puppet.debug("Invalid password for #{store}")
               return false
+            elsif e.message.include?('Keystore file exists, but is empty')
+              Puppet.debug(e)
+              return false
             else
               Puppet.log_exception(e, "Failed to read keystore '#{store}'")
```
ekohl, line 31: Just wondering, should this also be
```suggestion
              Puppet.log_exception(e, "Failed to read keystore '#{store}'")
              return false
```

ekohl (approved):

lib/puppet_x/certs/provider/keystore.rb
ehelms: I wasn't sure. Originally we said we do not know *what* exceptions could occur, and we should just halt execution for the user to inspect further rather than assuming. I could go either way still -- what do you think?

ekohl: That's a fair point. Let's keep it this way then

PR #481, opened 2025-02-20
Only create password file when generate is true

ehelms:
This was, for example, causing a password file to get created on proxies that was not used by anything.

### DIFF
--- a/manifests/ca.pp
+++ b/manifests/ca.pp
@@ -23,14 +23,18 @@ class certs::ca (
   $default_ca_path = "${certs::ssl_build_dir}/${default_ca_name}.crt"
   $server_ca_path = "${certs::ssl_build_dir}/${server_ca_name}.crt"
 
-  file { $ca_key_password_file:
-    ensure    => file,
-    content   => $ca_key_password,
-    owner     => 'root',
-    group     => 'root',
-    mode      => '0400',
-    show_diff => false,
-  } ~>
+  if $generate {
+    file { $ca_key_password_file:
+      ensure    => file,
+      content   => $ca_key_password,
+      owner     => 'root',
+      group     => 'root',
+      mode      => '0400',
+      show_diff => false,
+      notify    => Ca[$default_ca_name],
+    }
+  }
+
   ca { $default_ca_name:
     ensure        => present,
     common_name   => $ca_common_name,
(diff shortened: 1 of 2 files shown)


ekohl (commented):
I notice we pass it in as an option:
https://github.com/theforeman/puppet-certs/blob/50b43344f63c6ac95c78afa227efd3a85ff89c4b/lib/puppet/provider/ca/katello_ssl_tool.rb#L12

Though then I also see this:

https://github.com/theforeman/puppet-certs/blob/50b43344f63c6ac95c78afa227efd3a85ff89c4b/lib/puppet/provider/katello_ssl_tool.rb#L14-L16

Combined with:

https://github.com/theforeman/puppet-certs/blob/50b43344f63c6ac95c78afa227efd3a85ff89c4b/lib/puppet/provider/katello_ssl_tool.rb#L40-L41

So that makes me wonder: is the `ca` resource itself doing anything when `generate` is `false`? In other words, should that resource itself also be hidden behind `if $generate`?

Taking it a step further, removing `generate` from the types/providers would make it way easier to follow, but feel free to consider that out of scope.

ehelms:
> So that makes me wonder: is the `ca` resource itself doing anything when `generate` is `false`? In other words, should that resource itself also be hidden behind `if $generate`?
> 
> Taking it a step further, removing `generate` from the types/providers would make it way easier to follow, but feel free to consider that out of scope.

Yea... it's kind of afunky parameter, that I've been debating whether or not to mess with. We shouldn't need to wrap any of the CA/Cert resources in an `if` block from all of the testing and playing around I have been doing. I'll keep it in mind as I keep peeling back the layers. 

I am effectively trying to find and peel back as many layers as I can to help make the transition to containers / new installer smoother. This is related to https://github.com/theforeman/foreman-quadlet/issues/52 -- should I somehow attach these to that in some way?

ehelms:
Can this get a re-review?

ehelms:
> Should we ensure it's absent as well? To clean up on content proxies.

This is a weird grey area where sometimes generate get's set to false but deploy true, but you don't want to "remove" any files from the filesystem. I think the right spot to make it absent if we wanted is within this block -- https://github.com/theforeman/puppet-foreman_proxy_content/blob/master/manifests/init.pp#L163

ehelms:
@ekohl Thoughts given my previous answer?

evgeni:
LGTM, I guess, but would prefer to hear from Ewoud too :)

ekohl (approved):
Still have this nagging feeling this doesn't go far enough because it is passed into the `ca` as well, but I suppose it's a start.

PR #482, opened 2025-02-24
only create pki_dir when deploy=true

evgeni:


ekohl (commented):
I had a déjà vu moment: https://github.com/theforeman/puppet-certs/pull/436.

Any thoughts on my approach vs yours?

evgeni:
Well, clearly I should have looked first :see_no_evil:

ekohl:
To be fair, it was closed.

ehelms:
> I had a déjà vu moment: #436.

I also had a solution! https://github.com/theforeman/puppet-certs/pull/417  which inspired @ekohl :)

ekohl:
I guess @evgeni also went through the ritual to write a patch to remove `pki_dir`. Welcome to the club.

evgeni:
Right, and now I run into the problem y'all mentioned: I want the foreman_proxy certs deployed, but the rest not, so deploy is "true" globally… fuck

ehelms:
> Right, and now I run into the problem y'all mentioned: I want the foreman_proxy certs deployed, but the rest not, so deploy is "true" globally… fuck

I can work on that problem, as I think it's the one I figured you'd run into and we'd need to unravel a bit. Just to ensure we are on the same page, can you describe what you are needing to happen in more detail?

ehelms:
This has effectively been solved via https://github.com/theforeman/puppet-certs/pull/491

PR #484, opened 2025-02-27
Cleanup trustore and keystore passwordfiles in pki_dir

ehelms:


ehelms:
These were moved with https://github.com/theforeman/puppet-certs/commit/127357a299ac932d061074cd5fe7a1fa20fe3a0c 8 months ago.

ekohl (commented):
127357a299ac932d061074cd5fe7a1fa20fe3a0c was released in puppet-certs 19.0.0, which made its way into Foreman 3.12. I suppose we can now count on users to have applied this.

In researching https://github.com/theforeman/foreman-quadlet/issues/68 I wondered if we shouldn't actually move to configuring Candlepin/Tomcat to read files with passwords instead of embedding them into config files. At least where possible. That would actually move back in the opposite direction, but it makes it easier if we can statically configure the service and only treat the files as secrets.

ehelms:
> [127357a](https://github.com/theforeman/puppet-certs/commit/127357a299ac932d061074cd5fe7a1fa20fe3a0c) was released in puppet-certs 19.0.0, which made its way into Foreman 3.12. I suppose we can now count on users to have applied this.

Is that an approve for this particular change? :)



> In researching [theforeman/foreman-quadlet#68](https://github.com/theforeman/foreman-quadlet/issues/68) I wondered if we shouldn't actually move to configuring Candlepin/Tomcat to read files with passwords instead of embedding them into config files. At least where possible. That would actually move back in the opposite direction, but it makes it easier if we can statically configure the service and only treat the files as secrets.

I'm not sure it's possible for things like Tomcat, but we can investigate that further over in that effort.

ekohl:
> Is that an approve for this particular change? :)

Depends on where we go. If we reintroduce it, that would obsolete this change ;)

ehelms:
> > Is that an approve for this particular change? :)
> 
> Depends on where we go. If we reintroduce it, that would obsolete this change ;)

Nope! We didn't remove the password files, we simply moved them (https://github.com/theforeman/puppet-certs/commit/127357a299ac932d061074cd5fe7a1fa20fe3a0c).

ekohl (approved):
Ah, fair enough.

PR #485, opened 2025-02-27
Drop references to the deployed CA key that is not deployed

ehelms:
This only exists in the build directory and these references are unused.


ekohl (approved):
Looks like it's been unused since 95a9a34e75492b7ad1833d9723a6295b3a78735f

PR #487, opened 2025-02-28
Update and consolidate tarfile test cases

ehelms:


spec/acceptance/certs_tar_extract_spec.rb
```diff
@@ -3,87 +3,154 @@
...
+        }
+
+        include certs::apache
+      PUPPET
+
+      # generation of a certs tar archive and extraction of it are not idempotent by design
+      apply_manifest(install_certs, catch_failures: true)
+    end
+
+    describe 'default and server ca certs match' do
```
evgeni, line 132: this should read `default and server CA certs DON'T match`, right? (yes, it's copied code, but still confusing)

spec/acceptance/certs_tar_extract_spec.rb
```diff
@@ -3,87 +3,154 @@
...
+      its(:subject) { should match_without_whitespace(/C = US, ST = North Carolina, L = Raleigh, O = Katello, OU = SomeOrgUnit, CN = #{fact('fqdn')}/) }
+      its(:keylength) { should be >= 4096 }
+    end
+
+    describe x509_certificate('/etc/pki/katello/certs/katello-server-ca.crt') do
+      it { should be_certificate }
+      it { should be_valid }
+      it { should have_purpose 'SSL server CA' }
+      # These don't match since we only configure it with the intermediate
+      # and not the actual root
```
evgeni, line 150: Don't match what? Clearly they `match_without_whitespace` ;)
Should this rather read:
```suggestion
      # The issuer and subject are not identical as we only configure it with the intermediate
      # and not the actual root
```

spec/acceptance/certs_tar_extract_spec.rb
```diff
@@ -3,87 +3,154 @@
...
+      PUPPET
+
+      # generation of a certs tar archive and extraction of it are not idempotent by design
+      apply_manifest(install_certs, catch_failures: true)
+    end
+
+    describe x509_certificate('/etc/pki/katello/certs/katello-apache.crt') do
+      it { should be_certificate }
+      it { should be_valid }
+      it { should have_purpose 'server' }
```
evgeni, line 43: I know, this is again copied from the past, but reading it made me stumble for a moment.

Here you look for `server`, while in some place later you look for `SSL server CA`, the CA vs non-CA part is obvious, but the SSL made me read https://github.com/mizzy/serverspec/blob/master/lib/serverspec/type/x509_certificate.rb -- I shouldn't have done that.

What it does under the hood is `openssl x509 … -purpose | grep "${passed string} : Yes"`. The added ` : Yes` makes the CA/non-CA case "unique", but what about the SSL prefix?

```
# openssl x509 -in /etc/pki/katello/certs/katello-apache.crt -noout -purpose |grep server
SSL server : Yes
SSL server CA : No
Netscape SSL server : Yes
Netscape SSL server CA : No
```

I'd argue that we should always look for `SSL server` and `SSL server CA` *explicitly*. (Not that I think that anything would still produce Netscape-but-not-SSL tagged stuff in 2025, but still…)

evgeni:
I like the consolidation and left a few comments about the quality of the moved code, which shouldn't block merging, but are probably still a good update? ;-)

ehelms:
Updated and added a commit for the `have_purpose SSL server` change.

evgeni (approved):

PR #489, opened 2025-02-28
Deploy the default CA certificate into /etc/foreman

ehelms:
This is to help loosen up dependencies between different services certificate deployments.


ehelms:
See https://github.com/theforeman/puppet-katello/pull/511 for usage

PR #490, opened 2025-02-28
Bundle default and server CA certificate and use for Foreman client CA

ehelms:


evgeni:
Just to check whether I correctly understand the intent here: You want to create a bundle that contains all CAs Foreman needs to trust to be able to talk to the infra (we deploy).

What about services we _sometimes_ deploy? Like advisor. In https://github.com/theforeman/foreman_rh_cloud/commit/d3980f578f39d685b2faa488ee845d58e336aef3 we took a rather arcane approach of "use the special CA if cloud, otherwise rely on the system to trust the CA already"

Would/could/should this become
```ruby
if ForemanRhCloud.with_local_advisor_engine?
  Setting[:ssl_ca_file]
else
  "#{ForemanRhCloud::Engine.root}/config/rh_cert-api_chain.pem"
end
```
(Other calls in RH Cloud also rely on "system trusts the CA already", so it probably needs some Foreman-central API that we can rely on?)

ehelms:
> Just to check whether I correctly understand the intent here: You want to create a bundle that contains all CAs Foreman needs to trust to be able to talk to the infra (we deploy).

Correct.

> Would/could/should this become

Yes, we should unify this. https://github.com/Katello/katello/pull/11327 will do this for Candlepin communication. For system trust... I guess that depends on if we are relying on something that is intended o come from a distribution like RHEL and already be there (though if this isn't true for containers then scrap this thought and see the next one). If *we* are putting things in the trust and relying on that, we should stop.

evgeni:
> For system trust... I guess that depends on if we are relying on something that is intended o come from a distribution like RHEL and already be there (though if this isn't true for containers then scrap this thought and see the next one).

"like RHEL" and "be there" are interesting aspects ;-)

I'd be OK with relying on "normal" (browser forum approved) CAs being accepted "by the system" via the means of something like `ca-certificates`

but on a RHEL box the "RH CA" will "be there" (even if not trusted by default by the system), and I'd not rely on that (but we don't do that today, I think? [nope](https://github.com/Katello/katello/blob/master/ca/redhat-uep.pem))

> If _we_ are putting things in the trust and relying on that, we should stop.

The one place I know of is https://github.com/theforeman/puppet-foreman_proxy_content/blob/afb1f220a0e35dea52e1a305393daf36ab44f1b4/manifests/bootstrap_rpm.pp#L20-L24 which originates all the way back in https://github.com/theforeman/puppet-certs/commit/2c1992fc92a4807e0e39bb335d88144550b28f5e

I fixed one occurrence on "relying on this" in https://github.com/theforeman/smart_proxy_pulp/commit/12244ddef4d9c8f0377af16099a85faa94dcb9d3 but I bet there are more?

evgeni:
you broke the rebase

ehelms:
> you broke the rebase

Broke it hard. But this aint no humpty dumpty, and I put the pieces back together again.

evgeni:
Thanks. Now if GH would stop pretending "Job is completed before starting. " and cancel jobs before they start…

ehelms:
> I fixed one occurrence on "relying on this" in [theforeman/smart_proxy_pulp@12244dd](https://github.com/theforeman/smart_proxy_pulp/commit/12244ddef4d9c8f0377af16099a85faa94dcb9d3) but I bet there are more?

Probably... you are more pointing out other areas this might help fix or we should do a better job with? And not that those issues are blockers to this?

evgeni:
> Probably... you are more pointing out other areas this might help fix or we should do a better job with? And not that those issues are blockers to this?

No blockers, no.

More a "we should rip out those three lines and see what else starts smelling funny"

ekohl:
> but on a RHEL box the "RH CA" will "be there" (even if not trusted by default by the system), and I'd not rely on that (but we don't do that today, I think? [nope](https://github.com/Katello/katello/blob/master/ca/redhat-uep.pem))

I've been thinking about using an RPM dependency and forcing a symlink to replace that file. Probably in `rubygem-katello`.

evgeni:
> > but on a RHEL box the "RH CA" will "be there" (even if not trusted by default by the system), and I'd not rely on that (but we don't do that today, I think? [nope](https://github.com/Katello/katello/blob/master/ca/redhat-uep.pem))
> 
> I've been thinking about using an RPM dependency and forcing a symlink to replace that file. Probably in `rubygem-katello`.

This will break installation on at least Oracle, as they don't ship `subscription-manager-rhsm-certificates`

ehelms:
> More a "we should rip out those three lines and see what else starts smelling funny"

Agreed - let's track this elsewhere.

ehelms:
@evgeni Could you take another look?

manifests/ca.pp
```diff
@@ -53,6 +54,28 @@
...
+
+    if $certs::server_ca_cert {
+      concat::fragment { 'server-ca':
+        target => $ca_bundle_path,
+        source => $server_ca_path,
+        order  => '02',
+      }
+    }
+
+    file { "${certs::ssl_build_dir}/KATELLO-TRUSTED-SSL-CERT":
```
evgeni, line 76: Bad rebase?

evgeni (approved):

PR #492, opened 2025-03-01
Rely on the generated CA certificate location when deploying the CA cert

ehelms:
The `/etc/pki/katello` location for the CA certificate is needed by the Apache service and not as a general purpose location for the CA certificate. Other services that have certificates can rely on the CA from the `ssl-build` directory when deploying rather than an intermediate step that has it deployed to the Apache needed location. This should give a bit better separation of concerns.

Requires - https://github.com/theforeman/puppet-certs/pull/487


manifests/foreman.pp
```diff
@@ -55,12 +55,11 @@
     }
 
     file { $ssl_ca_cert:
-      ensure  => file,
-      source  => $server_ca,
-      owner   => 'root',
-      group   => $group,
-      mode    => '0440',
-      require => File[$server_ca],
```
ekohl, line 63: Isn't this still required? Or does Puppet automatically create that dependency since it's specified as a source?

manifests/ca.pp
```diff
@@ -58,24 +58,4 @@
...
-
-    file { $certs::katello_default_ca_cert:
-      ensure => file,
-      source => $default_ca_path,
-      owner  => 'root',
-      group  => 'root',
-      mode   => '0644',
-    }
-
-    file { $katello_server_ca_cert:
```
ekohl, line 73: That's a parameter on this class, but I think it's now unused and can be dropped.

While checking, I noticed 2 others were already unused: https://github.com/theforeman/puppet-certs/pull/495

manifests/ca.pp
```diff
@@ -58,24 +58,4 @@
       ensure  => absent,
     }
   }
-
-  if $deploy {
```
ekohl, line 62: This is now unused and can be dropped as a class parameter

manifests/apache.pp
```diff
@@ -134,5 +134,21 @@
...
+      owner  => 'root',
+      group  => 'root',
+      mode   => '0644',
+    }
+
+    file { $certs::katello_server_ca_cert:
+      ensure => file,
+      source => $certs::ca::server_ca_path,
+      owner  => 'root',
+      group  => $certs::group,
```
ekohl, line 150: It's world readable so why not hardcode it to `root`?
```suggestion
      group  => 'root',
```

ekohl (commented):
I like the direction, but CI doesn't like you.

manifests/foreman.pp
ehelms: From my experience and testing, yes, Puppet handles this since we set the source.

ehelms:
See https://github.com/theforeman/puppet-certs/pull/488#issuecomment-2698372078 about the failure

manifests/foreman.pp
binford2k: This `require` is not strictly required, but not because it's autorequired via the `source` parameter. It's because that's a resource contained in the `certs::ca` class that you require in line #44.

manifests/foreman.pp
```diff
@@ -55,12 +57,11 @@
...
+      ensure => file,
+      source => pick($server_ca, $certs::ca::server_ca_path),
```
ehelms, line 61: @ekohl Thoughts on this approach? With the `undef` optional parameter in the class?

I could just drop the class parameter all together - I don't think it really gets any value or will be used at this point. The `server_ca_path` is tied to the `certs::ca` class as that is what generates / defines it.

ehelms:
Still relevant for a few othe rclasses, @ekohl what do you think about an approach that uses:

```
      source => pick($server_ca, $certs::ca::server_ca_path),
```

And then makes `$server_ca` optional at the class parameter level?

ekohl:
> And then makes `$server_ca` optional at the class parameter level?

Depends. Today we have code calling `$certs::foreman::server_ca` and if it's `undef` at the top level, you'll need some other way to reuse the variable.

One thing we should consider is for the server side certificates is to use bundles where both the public key and the chain is included. Then we don't pass in the CA at all anymore. We'll need to verify this when a users passes them in, but it could simplify things.

ehelms:
The more I think about it, I don't think that `server_ca_cert` parameter is necessary and it doesn't make sense with the way the certificates are intended to work and depend upon the build directory.

ekohl:
Partly you're right: you really need to serve the chain. The client should already have the root CA, but you do need to serve the intermediates.

For example:
```
Certificate chain
 0 s:CN=example.com
   i:C=US, O=Let's Encrypt, CN=R10
   a:PKEY: rsaEncryption, 4096 (bit); sigalg: RSA-SHA256
   v:NotBefore: Jan 19 10:08:30 2025 GMT; NotAfter: Apr 19 10:08:29 2025 GMT
 1 s:C=US, O=Let's Encrypt, CN=R10
   i:C=US, O=Internet Security Research Group, CN=ISRG Root X1
   a:PKEY: rsaEncryption, 2048 (bit); sigalg: RSA-SHA256
   v:NotBefore: Mar 13 00:00:00 2024 GMT; NotAfter: Mar 12 23:59:59 2027 GMT
```

Here you can see certificate 0 is the server certificate, which you obviously need to serve. Then we also see the intermediate CA (`C=US, O=Let's Encrypt, CN=R10`) that issued the server certificate, but you don't see the root CA that issues the intermediate (`C=US, O=Internet Security Research Group, CN=ISRG Root X1`).

What you can do is mandate the intermediate CA is included in the `cert.pem` file and omit the `server_ca_file`.

If you ever looked at what certbot does then you see it writes out 4 files:

* `cert.pem` contains only the server certificate
* `chain.pem` contains only the chain between the server certificate and the root
* `fullchain.pem` is both `cert.pem` and `chain.pem` concatenated
* `privkey.pem` is the private key

Now you can configure Apache with:
```
SSLCertificateFile /path/to/fullchain.pem
SSLCertificateKeyFile /path/to/privkey.pem
```

The deprecated way (that we today still use in our installer):
```
SSLCertificateFile /path/to/cert.pem
SSLCertificateChainFile /path/to/chain.pem
SSLCertificateKeyFile /path/to/privkey.pem
```

And `SSLCACertificateFile` is only for client auth.

This is pretty easy to build for our default CA: we're in control. However, we're not for what users input and I'd struggle to come up with a good way to prevent users to shoot themselves in the foot. This is why I'd love to adopt [ACME](https://en.wikipedia.org/wiki/Automatic_Certificate_Management_Environment), which is what Let's Encrypt came up with. These days it's also implemented by others, including dogtag (the CA component of FreeIPA).

ehelms:
I am writing up a "future" certificate design with the guidelines we want to abide by as part of the foreman-quadlet research repository. This is great input to that, but I don't think I should include any of it here, specifically in this PR. I am aiming to mostly cleanup this puppet module to make it easier to consume in our future states by more clearly splitting generate and deploy, and deploying certificates to the locations they are needed for the service that needs them there.

Nearly all of these service classes are not used stand-alone anywhere today, and nor were they really designed to be even though they exposed parameters that would lead one to think that. The removed parameter `server_ca` was really not intended to ever be used due to how generate and deploy, are tied together and how this module in general was never meant to be used with anything other than the katello-certs-tools generated certificate method. We will fix that but I want to do that somewhere cleaner than here and just make this module have cleaner generate / deploy separation for flexibility.

ekohl:
I get that and it was me thinking zoomed out and comparing to a desired state before zooming back in again.

ehelms:
Cool - thoughts then on this specific change being good to merge?

PR #493, opened 2025-03-01
Remove cleanup of KATELLO-TRUSTED-SSL-CERT

ehelms:
Refs 33130e2aa712657c746210f5419026c00ffb1cc2

Additional cleanup of files that were marked absent in a previous release and should be deployed by now.

### DIFF
--- a/manifests/ca.pp
+++ b/manifests/ca.pp
@@ -53,10 +53,6 @@ class certs::ca (
       group  => 'root',
       mode   => '0644',
     }
-
-    file { "${certs::ssl_build_dir}/KATELLO-TRUSTED-SSL-CERT":
-      ensure  => absent,
-    }
   }
 
   if $deploy {


evgeni:
https://github.com/theforeman/puppet-certs/commit/33130e2aa712657c746210f5419026c00ffb1cc2 is "only" in 20.0, which means Foreman 3.14.
If somebody would skip that release, they'd be left with that uncleaned symlink.
I think that's OK, as we don't officially allow skipping releases, but wonder what @ekohl thinks?

ehelms:
Oh, you are right, I had looked at the PR date and not the merge date. It can wait, but I feel like 3.14 as the only release containing it should be sufficient.

ekohl (approved):
Normally I'm a bit more conservative but since we're already doing breaking changes in this release ...

